PR #9, opened 2019-08-19
docs(delimited-list): add wrapper with width to storybook demo

### LINKED ISSUE #3: DelimitedList - truncation doesn't work
## Bug - The `<DelimitedList />` component does not truncate

**Expected behavior -** The v2 `<DelimitedList />` should behave the same as the [master component](https://pages.github.ibm.com/security/carbon-addons-security/branch/master/?path=/story/application-delimitedlist--default) i.e. it should show an ellipsis when you reach the max width of the component, along with the number of total items in the list.

Should truncate as shown below:
ba760280-bde4-11e9-8a5e-440e898ccb27

**Actual behavior -** The component doesn't truncate at all, and just displays the overflow.

jendowns:
## Affected issues

- Resolves #3 

## Proposed changes

- remove unnecessary type class from storybook demo
- add wrapper with a `width` rule to storybook demo, to demonstrate truncation


src/components/DelimitedList/DelimitedList.stories.js
```diff
@@ -29,7 +29,11 @@ storiesOf(components('DelimitedList'), module)
   .addDecorator(centered)
   .add(
     'Default',
-    () => <DelimitedList className="bx--type-body-long-01" {...storyProps()} />,
+    () => (
+      <div style={{ width: '360px' }}>
+        <DelimitedList {...storyProps()} />
```
SimonFinney: - We can add this `style` property directly to the component itself to avoid an unnecessary `div`
- Also, are there `width` values we can leverage from the Carbon layout package here? We need to standardize on a width value across Storybook

SimonFinney: Nice! That's an elegant solution and reinforces grid usage

PR #16, opened 2019-08-21
feat(panel-v2): add experimental component with mocks, tests, and story

jendowns:
### PanelV2
- add *experimental* `PanelV2` component that was originally created by @areddon in our private repo

### ~PortalV2~
- ~add a single file `PortalV2` that includes slight updates to the existing Portal, but doesn't break any existing usage~

### Storybook config
- add "experimental" category and apply to `PanelV2`

## Testing instructions

- Storybook - https://deploy-preview-16--ibm-security.netlify.com/


src/components/Panel/_mixins.scss
```diff
@@ -10,13 +10,13 @@
...
 
-@mixin panel {
-  /// Width.
-  /// @type Length
-  $panel__sizing__width: carbon--mini-units(
-    $count: 40,
-  );
+/// Width.
+/// @type Length
+$panel__sizing__width: carbon--mini-units(
```
jendowns: Needed to move this out so it can be used by `PanelV2`. 
(NOTE: we should probably move all of these `Panel` variables up in scope like other components? But for this PR I wanted to keep changes at a minimum so I only moved the variable that was necessary 👍 )

src/components/PanelV2/PanelV2.js
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  componentDidMount() {
+    if (isClient() && this.state.isOpen) {
+      this.setBodyMargin();
+    }
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.isOpen && !prevState.isOpen && isClient()) {
```
danielcooke1996: Looks good, this is the only line i'm worried about. 

Checking to make sure `prevState.isOpen` is false - will this not prevent consumers from from swapping out the panel content? 

E.g
- Panel children is swapped out to fill a larger space
- Because the panel was not explicitly closed, `isOpen` will be true
- `setBodyMargin()` will not be called in this instance.

Would it be possible to remove this condition from the check?

In the `qradar-ui` rewrite we use a single Panel component across our app, which simply swaps out the content when the user opens a new panel.

This is to prevent overlapping panel instances.

One defect we are seeing with the v1 panel is that if a user opens a v4 IP. `setBodyMargin` is called initially which leaves just enough margin for the v4 IP address:

But if they then open a v6 IP Address - `setBodyMargin` is not called and the following is seen:
image

jendowns: Thank you @danielcooke1996! I agree with the premise here, but unfortunately removing that condition will cause a loop for `componentDidUpdate` & throw a "maximum depth" error 😭 

Generally though, I think we could still address your issue if we were to refactor the use of these lifecycle methods a bit. We currently have the older, deprecated methods anyway so there's some cleanup to be done. 🤔 

I think for now though... I'd like to keep things as-is just so we can get this experimental component in, for a couple teams to use. And we could address your concern (along with a lifecycle refactor) in a separate PR.

jendowns:
@SimonFinney Design review is good to go! Could you take a look when you get a chance? Since we last talked, I did add a `PortalV2.js` file because I had accidentally left out a couple Portal updates from Alex. Namely, he had removed the usage of the `PanelContainer`, which is not a part of the `PanelV2`. So I created a `PortalV2` just to be extra safe, so that existing `Portal` usage is unaffected by this change.

Also I'm curious what you think of @danielcooke1996's comment above & my response here: https://github.com/carbon-design-system/ibm-security/pull/16#discussion_r316863523

Thanks!

jendowns:
Hey @SimonFinney could you please take a look at this when you get a chance? We've got support on the design side of things for this change, and there are two applications hoping to utilize this `PanelV2` as soon as possible.

SimonFinney:
@jendowns Apologies for the delay! I was hoping to look at this today, but it's more than likely going to be tomorrow or Monday before I can dedicate time to review all of the changes! 😭

As per our conversation in Slack, am I right in saying that the custom footer can still be utilised by these application teams in the current `Panel` if the work needs to happen right now?

joshkimmell:
Considering that the custom footer isn't the ONLY requirement, both apps are required to use the V2 panel. We really need that ASAP. The current panel is essentially broken and v2 provides many fixes.

SimonFinney:
@jendowns Thanks for your ongoing patience - I have some suggestions based off what we've discussed over the last few days that I feel positions us better to manage any potential debt as a result of shipping the updated `Panel`.

I'm confident that absorbing the `DataDecorator` and `Portal` changes will have little impact on consumers bar those that are expecting to leverage this work immediately. The example PR (github.com/carbon-design-system/ibm-security/pull/49) illustrates some of this, but should only be used as a guideline. 

As we're against the clock and all of this has been through an extensive review process at different points already, I feel the changes are smaller than they seem, structural, and relate mostly to the mitigation of long-term debt. Key points include (in order of importance): 

- The `DataDecorator` and `Portal` updates should be absorbed into those components without affecting the inclusion of `renderFooter`
- The `primaryButton` and `secondaryButton` API should remain to ease migration for those not needing a custom footer
- It's not necessary for us to export the `PanelContent` as `PanelV2Content`, as this component requires no breaking change and will only be extra work to remove later
- As we expect the `PanelV2` implementation to be final, we no longer need to designate it as 'experimental'
- We should ensure the migration documentation is included to empower teams to migrate immediately
- The `PanelV2` stories should closely match the `Panel` stories

jendowns:
Hey @SimonFinney I'm not clear on these 2 points:

> The primaryButton and secondaryButton API should remain to ease migration for those not needing a custom footer

Why not take the opportunity to deprecate these two props? We are dealing with some prop bloat & this API is more restrictive than allowing people to just add their own buttons (would be straightforward with `renderFooter` right?). This would be the perfect opportunity to clean up the API, yes? 🤔 


----

About the `Portal` --

So we are good to remove usage of the panel container class in that component in the next release?

jendowns:
Hey @SimonFinney per our conversation earlier I just did the following --

* removed "experimental" category and put `PanelV2` in the "patterns" category
* restored `primaryButton` and `secondaryButton` props (with story examples and deprecation warnings)

And then separately, in another PR, would could update the `DataDecorator`.


Could you take another look when you get a chance? Thanks!

jendowns:
@SimonFinney I just pushed an update to the `DataDecorator` to include that in this PR.

The changes are copy & pasted from your changes here: https://github.com/carbon-design-system/ibm-security/pull/49

The primary difference is that I went ahead and added deprecation warnings for `primaryButton` and `secondaryButton` props...

And then I updated the default prop value for `primaryButton` in both `DataDecorator` and `PanelV2` from `null` to `undefined` so that it doesn't trigger a deprecation warning no matter what 👍 

Let me know what you think - thanks!

SimonFinney:
Thanks @jendowns - this all looks great so far! I was unable to locate the `Portal` revisions - were these carried over?

jendowns:
@SimonFinney Portal updates are in now 👍

SimonFinney:
@jendowns I think we're good to go! Thank you!

PR #28, opened 2019-08-23
fix(type-layout): set xs and sm font sizes to caption-01

### LINKED ISSUE #27: TypeLayout Component (with small & x-small) are wrong font-size
## Bug - TypeLayout Component (with small & x-small) are wrong font-size.

**Expected behavior -**

The font-size should follow the label style from Carbon, which is 12px.

**Actual behavior -**

The current font-size follows the body style, which is 14px.

### Steps for reproducing

1.  View component in storybook: https://ibm-security.carbondesignsystem.com/?path=/story/components-typelayout--with-x-small

jendowns:
## Affected issues

- Resolves #27 

## Proposed changes

- set xs and sm font sizes to `caption-01` (12px)

### DIFF
--- a/src/components/TypeLayout/_mixins.scss
+++ b/src/components/TypeLayout/_mixins.scss
@@ -61,11 +61,7 @@
       padding: $carbon--spacing-05;
     }
 
-    &--sm {
-      @include carbon--type-style($name: caption-01);
-      padding: $carbon--spacing-02;
-    }
-
+    &--sm,
     &--xs,
     &--xs:first-of-type {
       @include carbon--type-style($name: caption-01);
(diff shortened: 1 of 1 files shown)


src/components/TypeLayout/_mixins.scss
```diff
@@ -62,11 +62,13 @@
     }
 
     &--sm {
+      @include carbon--type-style($name: caption-01);
```
SimonFinney: It looks like there's some duplication here - We can reduce that by grouping the selectors

jendowns: 👍 Done

PR #33, opened 2019-08-26
[WIP] feat(nav): add renderLink prop for custom links in nav and toolbar

### LINKED ISSUE #31: Support for other React routers in Nav
### Proposal

Currently can not use Shell/Header/Nav with custom routers like [react-router](https://reacttraining.com/react-router/) or [@reach/router](https://reach.tech/router) due to config in Shell not allowing for custom Link component

### Goals / use cases

- [ ] Use IBM Security Shell in a React SPA

jendowns:
## Affected issues

- Resolves #31

## Proposed changes

- adds `renderLink` prop to `NavItem` that accepts a custom link component
- adds logic to `Toolbar` to process a potential `renderLink` prop
- update mocks and stories to include example custom link via `renderLink`


jendowns:
@SimonFinney @j1mie @diego-codes when any of you have a chance, could you please take a look at this PR? 🙏 Thank you

SimonFinney:
@jendowns Sorry for the delay on this. Currently, the consumer has to add the correct props themselves.

What if the API looked something more like https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/UIShell/Link.js#L17 ? 

With the limitations engineering teams are currently facing with the `Shell`, I suspect we'll need to rearchitect it in the not too distant future to allow consumers to pass in more complex data to this component, more akin to Carbon's `UIShell` - I believe this would be the first step in that direction.

jendowns:
Thanks @SimonFinney! I like that idea. I'll move this to a "WIP" state & update the API.

jendowns:
Closing this PR since the direction I've gone now is so different. Will reopen with a different approach shortly 🎉

PR #41, opened 2019-08-28
 fix(accordion): make component styles modular

j1mie:
## Proposed changes

- Updates Accordion to use correct partials
- Updates SkeletonText (referenced by Accordion) to use correct partials

## Testing instructions

- Pull down PR locally, and import `_index.scss` in `src/Accordion/Accordion.stories.js`


src/components/Accordion/_index.scss
```diff
@@ -4,8 +4,22 @@
...
+@import '../../globals/variables/index';
+
+// Carbon components
 @import 'carbon-components/scss/components/accordion/accordion';
 
+// Security components
+@import '../../components/SkeletonText/index';
+
+// Carbon mixins
+@import '@carbon/type/scss/index';
```
SimonFinney: This is more of a general question, but should we be referencing the package entry points or the specific modules? In this case `@carbon/type/scss/styles`

j1mie: Good point - I guess we should reference the relevant partials only. Will update now!

PR #42, opened 2019-08-28
fix(shell): restructure shell styles

### LINKED ISSUE #32: Modularise `Shell` styles
#### Context

Modularise `Shell` component and sub-components styles  - 

- [ ] `Accordion`
- [ ] `Breadcrumb`
- [ ] `Button`
- [ ] `Header`
- [ ] `IconButton`
- [ ] `Nav`
- [ ] `ProfileImage`
- [ ] `Toolbar`
- [ ] `Tooltip`

SimonFinney:
## Affected issues

- Resolves #32 

## Proposed changes

- Update `Shell` and sub-components to use correct partials
- Move Carbon body overrides into `platform/body` to prevent override unless the platform styles are required
- Move Carbon type overrides into `globals/type` to prevent override unless the global styles are required
- Add comments for lines to revert

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -71,6 +71,8 @@ addParameters({
     ),
     storySort: (a, b) =>
       ORDER.indexOf(getCategory(a)) - ORDER.indexOf(getCategory(b)),
+
+    // TODO: `modular-styles` - Revert.
     // theme,
   },
 });
--- a/src/components/Header/_mixins.scss
+++ b/src/components/Header/_mixins.scss
@@ -6,7 +6,7 @@
 
 @import '@carbon/layout/scss/mini-unit';
 @import '@carbon/layout/scss/spacing';
-@import '@carbon/themes/scss/mixins';
+@import '@carbon/themes/scss/tokens';
 @import '@carbon/type/scss/styles';
 @import 'carbon-components/scss/globals/scss/helper-mixins';
 @import 'carbon-components/scss/globals/scss/layout';
--- a/src/components/Header/_mixins.scss
+++ b/src/components/Header/_mixins.scss
@@ -4,9 +4,21 @@
 /// @copyright IBM Security 2019
 ////
 
-@import '../../globals/index';
+@import '@carbon/layout/scss/mini-unit';
+@import '@carbon/layout/scss/spacing';
+@import '@carbon/themes/scss/mixins';
+@import '@carbon/type/scss/styles';
+@import 'carbon-components/scss/globals/scss/helper-mixins';
+@import 'carbon-components/scss/globals/scss/layout';
+
+@import '../../globals/motion/index';
+@import '../../globals/namespace/index';
 
 @import '../../platform/body/mixins';
+
+@import '../IconButton/mixins';
+
+@import 'variables';
 
 /// Name.
 /// @type String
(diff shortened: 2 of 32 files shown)


.storybook/config.js
```diff
@@ -71,6 +71,8 @@ addParameters({
     ),
     storySort: (a, b) =>
       ORDER.indexOf(getCategory(a)) - ORDER.indexOf(getCategory(b)),
+
+    // TODO: `modular-styles` - Revert.
```
j1mie: Good thinking 🎖

src/components/Header/_mixins.scss
```diff
@@ -4,10 +4,22 @@
...
+@import '@carbon/type/scss/styles';
+@import 'carbon-components/scss/globals/scss/helper-mixins';
+@import 'carbon-components/scss/globals/scss/layout';
+
+@import '../../globals/motion/index';
+@import '../../globals/namespace/index';
 
 @import '../../platform/body/mixins';
 
+@import '../IconButton/mixins';
```
j1mie: This is one issue I couldn't really figure out - we'd talked about keeping imports from other components in the components `index.scss` but this doesn't always seem to work. Thought?

src/components/Header/_variables.scss
```diff
@@ -0,0 +1,18 @@
+////
+/// Header variables.
+/// @group header
+/// @copyright IBM Security 2019
+////
+
+@import '@carbon/layout/scss/spacing';
+@import 'carbon-components/scss/globals/scss/layout';
```
j1mie: Hadn't come across a `_variables.scss` partial before. Interesting. Is this something we're using elsewhere?

src/components/Toolbar/_mixins.scss
```diff
@@ -4,10 +4,17 @@
...
+@import '@carbon/themes/scss/mixins';
+@import 'carbon-components/scss/globals/scss/layout';
+
+@import '../../globals/motion/index';
+@import '../../globals/namespace/index';
+@import '../../globals/rtl/index';
 
 @import '../../platform/body/mixins';
-@import '../Header/mixins';
+@import '../Header/variables';
```
j1mie: Makes sense now.

src/components/_index.scss
```diff
@@ -5,6 +5,7 @@
 ////
 
 // Application.
+@import 'Breadcrumb/index';
```
j1mie: These were just missing from the component stylesheet? Guessing they were being brought in by other components?

src/globals/layout/_index.scss
```diff
@@ -4,8 +4,6 @@
 /// @copyright IBM Security 2019
 ////
 
-@import '../variables/index';
-
 @import '@carbon/layout/scss/layout';
```
j1mie: What's the point of this file

j1mie: 😄

src/components/Header/_mixins.scss
SimonFinney: In this file, we explicitly require the `IconButton` mixin, but not the styles themselves

src/components/_index.scss
SimonFinney: We had it further below, but I rearranged it to the top

src/globals/layout/_index.scss
SimonFinney: This is a deprecation helper for consumers of older layout variables like `$spacing-*`, `get-fixed-size`, etc.

PR #43, opened 2019-08-29
DO NOT MERGE - fix(yarn): remove offline mirror

SimonFinney:
## DO NOT MERGE - fix(yarn): remove offline mirror

This is to investigate the long-term impact of the offline mirror on repo size. Every time a dependency is updated, that version along with any previous version will be stored in the history, bloating the repo over time.

If we were to just update to a new version of Carbon once a month, we can expect the Git repo to grow in size around 44 MB per year. This is only a small sample and the actual number would be much larger. 

### Branches

- `dev` - 68 MB
- `remove-offline-mirror` - 2.9 MB

### Dependency examples

- `carbon-components` - 2.1 MB
- `carbon-components-react` - 723 KB
- `carbon-icons-react` - 870 KB

### Proposed changes

- Remove offline mirror


jendowns:
@SimonFinney thinking about this... 🤔 

Is there a risk associated with repo size?

When someone starts by cloning the repo, couldn't we just direct them to make a shallow clone?

SimonFinney:
@jendowns Thanks for chiming in! This is exploratory at this stage, and I'm casually investigating pros and cons, so feel free to use this PR as a jumping-off point for discussion because there's no plan or pressure on merging this in any time soon.

I agree we could direct folks to make a shallow clone, but that seems like a workaround solution for something that should just work, right?

jendowns:
I'm not sure I would call a shallow clone a workaround solution. They definitely have a purpose & are an effective built-in option if you want to quickly clone & only get a certain history depth.

SimonFinney:
Absolutely, and I agree that they have a place, but I worry may be counterintuitive to advise changing the default and expected workflow of contributors to accommodate potential repo bloat... 

I'm trying to weigh up the value versus the long-term impact here - I remember discussing benefits, but I can't remember where! 😆

jendowns:
Yeah I'm trying to wrap my head around where an extended or large git history would have a negative effect on developers. So far the only answer I've found is it would make a fresh `git clone ...` (without depth options) take longer. But day to day development wouldn't be affected?

Some benefits of yarn offline mode that have come up in the past are referenced here: https://yarnpkg.com/blog/2016/11/24/offline-mirror/
Honestly the big benefit for me is the reliability of repeated CI builds without having to touch the registry.

SimonFinney:
Closing, as I don't think is any real concern

PR #47, opened 2019-08-30
feat(string-formatter): add string formatter

### LINKED ISSUE #24: `StringFormatter`
### Context

Absorb `StringFormatter` pattern.

SimonFinney:
## Affected issues

- Resolves #24 

## Proposed changes

- Add `StringFormatter` component

## Testing instructions

### DIFF
--- a/src/components/StringFormatter/StringFormatter.js
+++ b/src/components/StringFormatter/StringFormatter.js
@@ -1,0 +1,46 @@
+/**
+ * @file String formatter.
+ * @copyright IBM Security 2019
+ */
+
+import classnames from 'classnames';
+import { bool, string } from 'prop-types';
+import React from 'react';
+
+import { getComponentNamespace } from '../../globals/namespace';
+
+const namespace = getComponentNamespace('string-formatter');
+
+const StringFormatter = ({ className, truncate, value, width }) => (
+  <span
+    className={classnames(namespace, className, {
+      [`${namespace}--truncate`]: truncate,
+    })}
+    style={{ maxWidth: width }}
+    title={truncate && value}
+  >
+    {value}
+  </span>
+);
+
+StringFormatter.propTypes = {
+  /** @type {string} Optional class name. */
+  className: string,
+
+  /** @type {boolean} Whether or not the value should be truncated. */
+  truncate: bool,
+
+  /** @type {string} Value to format. */
+  value: string.isRequired,
+
+  /** @type {string} width to max the string at. */
+  width: string,
+};
+
+StringFormatter.defaultProps = {
+  className: null,
+  truncate: false,
+  width: null,
+};
+
+export default StringFormatter;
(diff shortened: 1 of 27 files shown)


src/components/StringFormatter/StringFormatter.js
```diff
@@ -0,0 +1,46 @@
...
+  /** @type {string} Optional class name. */
+  className: string,
+
+  /** @type {boolean} Whether or not the value should be truncated. */
+  truncate: bool,
+
+  /** @type {string} Value to format. */
+  value: string.isRequired,
+
+  /** @type {string} width to max the string at. */
```
jendowns: Small thing, but can we reword this to: "Max width of string."

src/components/StringFormatter/StringFormatter.js
```diff
@@ -0,0 +1,46 @@
...
+  width: string,
+};
+
+StringFormatter.defaultProps = {
+  className: null,
```
jendowns: I think for almost all other component we use an empty string in this case. 🤔 Could this be `className: ''` for consistency?

src/components/StringFormatter/__tests__/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+  beforeEach(() => {
+    stringFormatter = shallow(<StringFormatter {...props} />);
+  });
+
+  it('renders', () => {
+    expect(stringFormatter).toMatchSnapshot();
+  });
+
+  it('renders the truncation variation', () => {
```
jendowns: "renders the **truncated** variation" ?

src/components/StringFormatter/_index.scss
```diff
@@ -0,0 +1,18 @@
...
+/// @copyright IBM Security 2019
+////
+
+@import '../Component/mixins';
+
+@include component($name: string-formatter) {
+  display: inline-block;
+
+  &--truncate {
+    max-width: 80ch;
```
jendowns: I realize we are dealing with a character width here, but that `80` is still divisible by the mini unit 🤔 Could you use mini-units here for the value?

src/components/StringFormatter/_mocks_/index.js
```diff
@@ -0,0 +1,8 @@
+/**
```
jendowns: I'm wondering if it is worth having this file, if it's just `value: 'Value'` being saved? Could we simplify and remove the mock file?

src/components/StringFormatter/StringFormatter.js
SimonFinney: Good catch!

src/components/StringFormatter/__tests__/index.js
SimonFinney: Thanks!

src/components/StringFormatter/_index.scss
```diff
@@ -0,0 +1,20 @@
...
+
+@import '@carbon/layout/scss/mini-unit';
+
+@import '../Component/mixins';
+
+@include component($name: string-formatter) {
+  display: inline-block;
+
+  &--truncate {
+    max-width: carbon--mini-units($count: 84);
```
jendowns: I know I suggested changing this but now I'm questioning it. 😅 Since it comes out as `42rem` which is a bit different from what the original intent was, using the `ch` unit.

PR #49, opened 2019-08-30
DO NOT MERGE - chore(panel): minimise breaking changes

SimonFinney:
## DO NOT MERGE - chore(panel): minimise breaking changes

Spike to explore delivering `PanelV2` functionality while still maintaining the existing API for consumers.

### Proposed changes

- Explore minimising breaking changes for the panel updates


SimonFinney:
@jendowns I had 30 minutes to roughly explore how we could introduce the panel changes, while still maintaining the original APIs for everything except the `PanelV2` - Let me know if I'm completely missing something here!

jendowns:
Hey @SimonFinney I like the direction of this, and I feel like it would be a very appropriate effort down the line, when we are ready to do a major release that bumps up `PanelV2` to `Panel` and `DataDecoratorV2` to `DataDecorator`

To me this seems risky, given timelines. 

The other PRs for experimental components are based on previous PRs in our private repo that have been talked about & reviewed over a couple month period. They've been review and approved in other forms by me, you, and a couple other folks.

Moreover, introducing those `V2` components as experimental, rather than overriding any existing component, will *definitely* not have an effect on existing users. 

This effort seems to increase our risk for these changes dramatically. It is rushing in major changes to existing components. **I worry that we don't have the time to properly absorb the impact of these changes. Whereas, again, the experimental components give us more leeway to explore impact while also delivering required functionality in an abbreviated timeline.**

SimonFinney:
Thanks for your feedback - I should’ve prefaced this PR with the intention that it never should be merged, it was purely to illustrate how we could introduce the required breaking changes for the panel without duplicating the ‘DataDecorator’ and ‘Portal’ - please do not take this as verbatim as it’s a rough draft.

PR #68, opened 2019-09-04
feat(combo-button): add component with tests and story

### LINKED ISSUE #66: Request - add ComboButton component
### Details

- **Application / team -** [redacted]
- **Work item -** [redacted]
- **Needed by -** ASAP

### Proposal

Promote `ComboButton` component from an application team to this library.

jendowns:
## Affected issues

- Resolves #66

## Proposed changes

- add `ComboButton` and `ComboButtonItem` with tests and story


src/components/ComboButton/ComboButton.stories.js
```diff
@@ -0,0 +1,46 @@
...
+  overflowMenuDirection: radios(
+    'Direction',
+    {
+      top: TooltipDirection.TOP,
+      bottom: TooltipDirection.BOTTOM,
+    },
+    TooltipDirection.TOP
+  ),
+});
+storiesOf(patterns('ComboButton'), module)
```
diego-codes: What's the difference between a pattern and a component?

src/components/ComboButton/ComboButton.stories.js
```diff
@@ -0,0 +1,46 @@
...
+  .addDecorator(withA11y)
+  .addDecorator(centered)
+  .add('default', () => <ComboButton {...props()} actions={actions} />)
+  .add('single item', () => <ComboButton {...props()} actions={[actions[0]]} />)
+  .add('overflow items', () => {
+    const newActions = [...actions];
```
diego-codes: I think the purpose of the overflow items story is to have lots of actions to demonstrate the max-height styles of the overflow menu.
```suggestion
    const newActions = [...actions, {
      label: 'Filter list',
      renderIcon: Filter20,
      onClick: action('Filter list 1'),
    },
    {
      label: 'Filter list',
      renderIcon: Filter20,
      onClick: action('Filter list 2'),
    },
    {
      label: 'Filter list',
      renderIcon: Filter20,
      onClick: action('Filter list 3'),
    }];
```

src/components/ComboButton/__tests__/ComboButton.spec.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import ComboButton from '../';
+
+import actions from '../_mocks_';
+
+describe('ComboButton', () => {
```
diego-codes: It would be good to add tests to ensure that one action does not render the overflow menu

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+    <div
+      className={classnames(`${namespace}__wrapper`, className)}
+      data-testid="combo-button"
+      ref={wrapper}
+      data-floating-menu-container
+    >
+      <Button
+        {...firstAction}
+        renderIcon={restActions.length > 1 ? null : firstAction.renderIcon}
+        data-testid="combo-button-primary-button"
```
diego-codes: It doesn't seem like you're using testids in your tests so I would remove them.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+
+const { prefix } = settings;
+
+const OVERFLOW_MENU_ICON = {
+  OPEN: ChevronUp16,
+  CLOSE: ChevronDown16,
+};
+
+const ComboButton = props => {
+  const [overflowMenuIcon, setOverflowMenuIcon] = useState(
```
diego-codes: It might be better to toggle a boolean state value instead of a component object because I would think that objects would have to be serialized to be compared. You can then use the boolean value to determine which icon to render.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+  );
+  const wrapper = useRef(null);
+  const { className, overflowMenuDirection, actions } = props;
+  const [firstAction, ...restActions] = actions;
+  const isTop = overflowMenuDirection === TooltipDirection.TOP;
+
+  const getMenuOffset = () => {
+    const { top } = wrapper.current.getBoundingClientRect();
+    return {
+      top: isTop ? top : top * -1,
```
diego-codes: It would be better to divide by `-1` instead of multiply because if `top` is `0` the result would be `-1` instead of `0`

```suggestion
      top: isTop ? top : top / -1,
```

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+      left: 'auto',
```
diego-codes: Should the left value also be determined by the bounding client rect as well? I think `auto` might be relying on browser implementations of `auto` which may be different.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+        primaryFocus={index === (isTop ? items.length - 1 : 0)}
+        onClick={item.onClick}
+      />
+    ));
+  };
+
+  return (
+    <div
+      className={classnames(`${namespace}__wrapper`, className)}
+      data-testid="combo-button"
```
diego-codes: It doesn't seem like you're using testids in your tests so I would remove them.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+      {restActions.length > 0 && (
+        <OverflowMenu
+          menuOffset={getMenuOffset}
+          className={classnames(
+            buttonNamespace,
+            `${prefix}--btn`,
+            `${prefix}--btn--primary`
+          )}
+          menuOptionsClass={`${prefix}--list-box__menu`}
+          data-testid="combo-button-overflow-button"
```
diego-codes: It doesn't seem like you're using testids in your tests so I would remove them.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+        primaryFocus={index === (isTop ? items.length - 1 : 0)}
+        onClick={item.onClick}
+      />
+    ));
+  };
+
+  return (
+    <div
+      className={classnames(`${namespace}__wrapper`, className)}
```
diego-codes: I would think that we could get away with just using the namespace as the root element classname instead of `__wrapper`
```suggestion
      className={classnames(namespace, className)}
```

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+  className: PropTypes.string,
+
+  /** @type {string} Overflow menu direction. */
+  overflowMenuDirection: PropTypes.oneOf(
+    TooltipDirection.TOP,
+    TooltipDirection.BOTTOM
+  ),
+
+  /** @type {array<string>} Array of actions. */
+  actions: PropTypes.arrayOf(
```
diego-codes: I'm rethinking the idea that accepting an array of options makes sense. Similar to the `OverflowMenu` that accepts `OverflowMenuItem`s as children, this component should probably do the same and accept  `ComboButtonItem`s or `ComboButtonAction`s as children.

I'm not sure how feasible this would be but it would be the ideal API for this component for debugging and consistency reasons. It would make debugging much easier than debugging object shapes in an array as a prop.


Example:
```jsx
<ComboButton>
  <ComboButtonItem renderIcon={Icon} onClick={() => {}}>
    Action 1
  </ComboButtonItem>
  <ComboButtonItem renderIcon={Icon} onClick={() => {}}>
    Action 2
  </ComboButtonItem>
</ComboButton>

File: src/components/ComboButton/_mixins.scss [hunk 12]
```diff
@@ -0,0 +1,95 @@
...
+/// @copyright IBM Security 2019
+////
+
+@mixin combo-button {
+  &__wrapper {
+    position: relative;
+    display: flex;
+    justify-content: flex-end;
+
+    & .#{$prefix}--btn--primary:first-child {
```
diego-codes: For styling purposes of this component, I would suggest that combo-button classes are better than repurposing component classnames that we don't have control over.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+      >
+        {firstAction.label}
+      </Button>
+      {restActions.length > 0 && (
+        <OverflowMenu
+          menuOffset={getMenuOffset}
+          className={classnames(
+            buttonNamespace,
+            `${prefix}--btn`,
+            `${prefix}--btn--primary`
```
diego-codes: You could add a combo button specific classname for styling purposes specifically for this component.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+          )}
+          menuOptionsClass={`${prefix}--list-box__menu`}
```
diego-codes: I'm pretty sure this the overflow menu component already provides this classname 🤔

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+      top: isTop ? top : top * -1,
+      left: 'auto',
+    };
+  };
+
+  const renderActions = items => {
+    const itemClassName = `${prefix}--list-box__menu-item`;
+    return items.map((item, index) => (
+      <OverflowMenuItem
+        className={itemClassName}
```
diego-codes: I'm pretty the overflow menu component already provides this classname also 🤔

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,137 @@
...
+        key={item.label}
+        itemText={
+          <Fragment>
+            <span className={`${itemClassName}-label`} title={item.label}>
+              {item.label}
+            </span>
+            <item.renderIcon className={`${itemClassName}-icon`} />
```
diego-codes: Should the icon have an empty description?

src/components/ComboButton/ComboButton.stories.js
jendowns: A pattern is like a "complex component" that is made up of several pieces. And then a component is more like something core (and usually proxied from Carbon).

@petervachon did you have anything extra to add to this explanation? ☝️

src/components/ComboButton/ComboButton.js
jendowns: @diego-codes I think it should be good since `iconDescription` no longer has a default here. So nothing extra gets added to the svg and it keeps `aria-hidden="true"`.

But I did go ahead and remove that `title` from the `span` because it doesn't seem necessary. 🤔 What do you think about it now?

src/components/ComboButton/ComboButton.js
jendowns: I think this is the *one* class that isn't getting added. 🤔 Or at least when I tried to remove it, the class was not added (and that prevents a `maxHeight` from getting added so the menu scrolls)

src/components/ComboButton/ComboButton.js
jendowns: Just clarifying: understanding is that these specific classes are needed to apply some of the Carbon `Button` styling to the `OverflowMenu`

Definitely agree though that I should add a new specific security-namespaced class for custom styling 👍

diego-codes: Yeah, I believe you're right and I'm with you on using our own namespace for custom styling ⭐️

src/components/ComboButton/ComboButton.js
diego-codes: I think the `title` is necessary for when the text overflows out of the allowed space so it is truncated with an ellipsis.

src/components/ComboButton/ComboButton.js
diego-codes: hmm that's bizarre

src/components/ComboButton/ComboButton.js
jendowns: oh oops, you are right -- will add that back 👍

jendowns: Hey one thing about this: now that we're using `ComboButtonItem`, which is basically just a renamed `OverflowMenuItem` that allows users to pass in any old node via the `itemText` prop, they would likely need to control this truncation themselves 🤔 I added an example & notes to the storybook on how to do that though.

src/components/ComboButton/ComboButton.js
jendowns: Good point but honestly I'm unsure what other value this should be 🤔

jendowns:
Hey @diego-codes this is ready to look at again! I refactored to follow the structure you suggested above: https://github.com/carbon-design-system/ibm-security/pull/68#discussion_r321819098

I also added a note about truncation and how it's going to be up to the user now: https://github.com/carbon-design-system/ibm-security/pull/68#discussion_r322415873

Let me know what you think! Thanks 👍

src/components/ComboButton/__tests__/ComboButton.spec.js
```diff
@@ -0,0 +1,52 @@
...
+            <Fragment>
+              <span>Filter list</span>
+            </Fragment>
+          }
+        />
+      </ComboButton>
+    );
+
+    expect(comboButton).toMatchSnapshot();
+    expect(comboButton.render()).toMatchSnapshot();
```
diego-codes: If you're already doing a snapshot assertion with the render method, do you really need the snapshot test above?

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,113 @@
...
+          {children}
+        </OverflowMenu>
+      )}
+    </div>
+  );
+};
+
+ComboButton.propTypes = {
+  /** @type {Object.<string, string>} An object list of props for the primary button. */
+  primaryAction: PropTypes.shape({
```
diego-codes: I was thinking that the primary action would be gathered from the children group. For example usage like this:

```jsx
<ComboButton>
  <ComboButtonItem renderIcon={Icon} onClick={() => {}}>
    Action 1
  </ComboButtonItem>
  <ComboButtonItem renderIcon={Icon} onClick={() => {}}>
    Action 2
  </ComboButtonItem>
</ComboButton>
```
Would take the first action and make it the primary action, basically abstracting away that logic within the component. Alternatively, a consumer can tag one of the `ComboButtonItem`s as the primary action:

```jsx
<ComboButtonItem renderIcon={Icon} onClick={() => {}} primary>Action 1</ComboButtonItem>
```

What do you think? Is this something that is doable and makes sense? My main motivation is to focus on flat props (instead of objects) which are easier to work with and debug IMO.

jendowns: Ah yeah that makes sense! 👍 Sounds good to me; I can make that change

src/components/ComboButton/__tests__/ComboButton.spec.js
jendowns: Good catch - I'm going to keep line 49 & remove line 48. 👍

src/components/ComboButton/ComboButton.js
jendowns: Well @diego-codes I was thinking about this & it may be difficult to achieve 🤔 

Right now `ComboButtonItem` is just a renamed `OverflowMenuItem` (side note: doesn't *have* to be this way, but it was definitely the path of least resistance, otherwise I was getting strange focus / tabbing results that I messaged you in Slack about).

The props for `ComboButtonItem` are pretty different from the props for `Button` (which is the underlying "primary action"). I notice in your mocked example you have button props like `renderIcon` etc but those aren't available to `ComboButtonItem` currently.

To that end, I think the biggest challenge is the `ComboButtonItem` / `OverflowMenuItem` components' `itemText` prop. That prop accepts nodes / anything really, so the user has a lot of freedom there. In the story examples I tried explaining how best to use `itemText` to get a well-formatted `ComboButton` in general, but it would be strange to take what's passed into `itemText`, especially when we are talking about maybe `span` elements and icon components, and then fit that into a `Button`. 🤔 

Anyway what do you think about all of this? 
I'm thinking that the 2 routes to solve this, if we want to stick with this general structure, are:
1. Keep `primaryActions` object
2. Have something like `ComboButtonAction` that renames a `Button` for this purpose, so it can be used as a sibling of `ComboButtonItem` components? Or just have a `Button` be used directly? My one worry here is that the API may be too open / lenient, it may imply that it's okay to put any number of other children here. One way to get around may be to add some warning if something other than `ComboButtonItem` or `ComboButtonAction`/`Button` are added as children of `ComboButton`?

What do you think? 🤔 😅

diego-codes: I definitely see what you mean. I like the idea of having a separate component for the primary action. It reminds me of the Carbon UI Shell and DataTable components where it allows developers compose components with the provided building blocks.

I'm also thinking that another approach could be that the `ComboButtonItem` component becomes a shell component that renders either a `Button` or `OverflowMenu`. Pardon my over directing, here is what I could see working:

```jsx
const ComboButtonItem = props => {
  const { primary, renderIcon: Icon, children, className, onClick, iconDescription } = props;
  const classNames = classnames(namespace, { [`${namespace}--primary`]: primary }, className);
  if (primary) {
    return (
      <Button
        className={classNames}
        renderIcon={Icon}
        onClick={onClick}
        iconDescription={iconDescription}
      >
        {children}
      </Button>
    );
  }

  return (
    <OverflowMenuItem
      className={classNames}
      itemText={
        <Fragment>
          <span title={children}>{children}</span>
          <Icon iconDescription={iconDescription} />
        </Fragment>
      }
    />
  );
};

ComboButtonItem.propTypes = {
  primary: PropTypes.bool,
  className: PropTypes.string,
  renderIcon: PropTypes.node,
  children: PropTypes.node.isRequired,
  onClick: PropTypes.func.isRequired,
  iconDescription: PropTypes.string,
};
```

This is just a hunch that it would work, but I haven't tested it. Either approach I think would work.

jendowns: Thanks @diego-codes! This definitely makes sense & I appreciate you sharing your approach. 

One challenge that I see is that the primary `Button` and the `OverflowMenuItem` components are not siblings within `children` of the `ComboButton` component. 🤔 The `Button` needs to be outside of the underlying `OverflowMenu`,  whereas the `OverflowMenuItem` components need to be inside... I'm just wondering how best to accomplish that without adding too much complexity.

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,133 @@
...
+    return {
+      top: isTop ? top : top * -1,
+      left: 'auto',
+    };
+  };
+
+  const childrenArray = React.Children.toArray(children);
+
+  // Save first child (e.g., primary action) to use as a `Button`:
+  const primaryAction = React.cloneElement(childrenArray[0]);
```
diego-codes: I wonder if you need to actually clone the element first or if you can just use `childrenArray[0]`?
```suggestion  
  const primaryActionWithProps = [childrenArray[0]].map(
```

src/components/ComboButton/ComboButton.js
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+  // Save remaining children to be displayed in the `OverflowMenu`:
+  let overflowItems;
+  let overflowMenuItemWithProps;
+  if (childrenArray.length > 1) {
+    overflowItems = childrenArray.slice(1);
+
+    // Create `OverflowMenuItem` components:
+    overflowMenuItemWithProps = React.Children.toArray(overflowItems).map(
```
diego-codes: I don't think you need to run this through `toArray` again since `overflowItems` is already an array

src/components/ComboButton/ComboButton.stories.js
```diff
@@ -0,0 +1,84 @@
...
+                1} - text may be long and will be truncated`;
+              return (
+                <ComboButtonItem
+                  className="some-class"
+                  key={`combo-button-item-'${item.id}`}
+                  index={index}
+                  onClick={action(`onClick ("${text}")`)}
+                >
+                  <span title={text}>{text}</span>
+                  <Filter20 />
```
diego-codes: I would make the API for the `ComboButtonItem` consistent. I like the use of the `renderIcon` that is used for the primary action instead of having the user provide their own children markup for the overflow actions.

src/components/ComboButton/ComboButtonItem/ComboButtonItem.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @file Combo button item.
+ * @copyright IBM Security 2019
+ */
+
+import React from 'react';
+import { namespace } from '../ComboButton';
+
+const ComboButtonItem = props => {
```
diego-codes: It's probably best to document the necessary prop types for this component otherwise consumers won't know what they need.

src/components/ComboButton/ComboButtonItem/ComboButtonItem.js
```diff
@@ -0,0 +1,26 @@
...
+import OverflowMenuItem from '../../OverflowMenuItem';
+import { namespace } from '../ComboButton';
+
+const ComboButtonItem = props => {
+  const { ...rest } = props;
+  return <span {...rest} className={`${namespace}-item`} />;
+};
+
+ComboButtonItem.defaultProps = {
+  ...Button.defaultProps,
```
diego-codes: These props might be overwhelmingly long and conflicting possibly. It opens up way too much for a consumer to make sense of. For example `isDelete` does not apply to Buttons, but the prop types would say that it is a valid prop. It would force consumers to worry too much about the implementation of the `ComboButton`. I would just list out the specific ones that you need to render and nothing else.

They could just be: 
```js
{
  children: PropTypes.string.isRequired,
  renderIcon: PropTypes.node,
  onClick: PropTypes.onClick,
  disabled: PropTypes.bool,
  className: PropTypes.string,
}
```

Any other prop would just be confusing and unnecessary.

PR #71, opened 2019-09-09
docs(ica): add grid example and add box-sizing reset to storybook styles

### LINKED ISSUE #23: ICA - add story variation of multiple ICAs, inside grid columns
### Proposal

Need to add a storybook variation for the ICA component that demos how to include multiple ICAs side by side, aligned to the grid. Template referred to as an "ICA Wall"

jendowns:
## Affected issues

- Resolves #23 

## Proposed changes

- add "with ICA wall" example that shows how to use grid classes to display multiple ICA components
- reformat a couple of the storybook names to match convention used for other stories ("with" etc)
- add `box-sizing: border-box` reset to Storybook (⚠️ NOTE: this is necessary to use the grid, as we don't apply a reset globally)

### DIFF
--- a/src/components/ICA/ICA.stories.js
+++ b/src/components/ICA/ICA.stories.js
@@ -43,6 +43,11 @@ storiesOf(components('ICA'), module)
     () => (
       <div className="bx--grid bx--grid--full-width">
         <div className="bx--row">
+          <div className="bx--col">
+            <h4>4 spaced</h4>
+          </div>
+        </div>
+        <div className="bx--row">
           {Array(4)
             .fill(0)
             .map(item => (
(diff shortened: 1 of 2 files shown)


src/components/ICA/ICA.stories.js
```diff
@@ -30,11 +29,54 @@ const storyProps = ({ value = icaValue, total } = {}) => ({
...
+  .add('with 1000000 value', () => (
     <ICA {...storyProps({ value: icaValue * 100000 })} />
   ))
-  .add('Total', () => <ICA {...storyProps({ total })} />);
+  .add('with total', () => <ICA {...storyProps({ total })} />)
+  .add(
+    'in an ICA wall',
+    () => (
+      <div className="bx--grid bx--grid--full-width">
+        <div className="bx--row">
```
diego-codes: This looks good to me, although it's a bit jarring to see this layout: 
image

I first thought that it was the grid wrapping in weird ways and that all of the ICAs were meant to go in one row. A way around this could be to just add a label on each row to differentiate between the two different layouts. Something like this, but perhaps better worded and styled:

image

diego-codes: Or the wall preview could just be a single row?

jendowns:
Thanks @diego-codes! I ended up adding 2 more rows to contain headings for the ica wall examples. I like that route personally because it's more of an opportunity to demonstrate grid classes (since these two new rows use the `bx--col` (no breakpoints or span) class in good context). What do you think of this update?

PR #86, opened 2019-09-18
fix(panel-v2): add combobutton to panelv2 story & update flex style rule

jendowns:
## Proposed changes

- Use `ComboButton` component in custom footer example in `PanelV2` story
- Add parameter to change position of "show info" bottom so it doesn't block the `ComboButton` example
- Update `flex-grow` style rule for children of `PanelV2` footer


src/components/PanelV2/PanelV2.stories.js
```diff
@@ -125,13 +175,31 @@ storiesOf(patterns('PanelV2'), module)
...
-                    onClick={this.closeSecond}
-                  >
-                    Add
-                  </Button>
+                  <ComboButton {...comboButtonProps()}>
+                    <ComboButtonItem
+                      onClick={action('onClick (Item 1 - primary button)')}
+                      renderIcon={ArrowRight20}
+                    >
+                      Item 1 (becomes primary button and text will be truncated)
```
j1mie: Should we do this as a template string { `Item...` }

src/components/PanelV2/PanelV2.stories.js
```diff
@@ -142,7 +210,11 @@ storiesOf(patterns('PanelV2'), module)
                       <CodeSnippet type="inline" light>
                         renderFooter
                       </CodeSnippet>{' '}
-                      prop to render a custom footer.
+                      prop to render a{' '}
+                      <CodeSnippet type="inline" light>
+                        ComboButton
+                      </CodeSnippet>{' '}
+                      in a custom footer.
```
j1mie: Same as above

src/components/PanelV2/_mixins.scss
```diff
@@ -140,7 +140,7 @@
     bottom: 0;
     width: $panel__sizing__width;
 
-    .#{$namespace}button {
+    .#{$namespace}button:not(.#{$namespace}combo-button__overflow-menu) {
```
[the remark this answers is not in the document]
jendowns: @j1mie that would likely only catch situations where someone uses the deprecated `primaryAction` prop or renders a `ComboButton` inside the new `renderFooter` render prop. 🤔 

Since any button could be rendered via `renderFooter`, it seems best to just filter out the one instance where we don't want that flex rule to apply.

src/components/PanelV2/PanelV2.stories.js
jendowns: Sure thing -- this was autoformatting from prettier so I don't mind either way 👍

jendowns:
@j1mie Thanks for the feedback! I made a couple changes to the story in my latest commit, and I also responded to your question about the style rule. Let me know what you think! 👍

PR #93, opened 2019-09-23
fix(panel-v2): if title prop is a string, use h2 element

jendowns:
Currently working in an application where an `h2` would be appropriate here, given that a `string` is being passed in. So I think it would be nice to check for a `string`, that way we could add proper heading hierarchy to the content of the panel. This is would be ideal to help assistive tools/tech navigate the panel more effectively.

**NOTE:** since the `title` prop allows `node` types

## Proposed changes

- add `typeof` check for `PanelV2`'s `title` prop (if `string`, use `h2` element instead of a plain `div`)
- update snapshots


src/components/PanelV2/PanelV2.js
```diff
@@ -102,7 +102,13 @@ export default class PanelV2 extends Component {
               <header ref={this.header} className={`${namespace}__header`}>
                 {title && (
                   <div className={`${namespace}__header__container--title`}>
-                    <div className={`${namespace}__header--title`}>{title}</div>
+                    {typeof title === 'string' ? (
+                      <h2 className={`${namespace}__header--title`}>{title}</h2>
```
jendowns: Decided to go with an `h2` instead of an `h1` here to mimic the structure of the `Modal`: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L343

SimonFinney:
The intent makes sense, but I have a question around the implementation ( @diego-codes ,
 @j1mie , @jendowns ) - if we're now declaring the title as a `node`, shouldn't the expectation be that the consumer supplies whatever element they want and the rendering of that is consistent regardless of the type (see https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Accordion/AccordionItem.js#L73)?

If the purpose of this is to help with assistive tools, perhaps there are ARIA roles that can be leveraged on the element to prevent this kind of extra type checking?

jendowns:
@SimonFinney 

>  if we're now declaring the title as a node, shouldn't the expectation be that the consumer supplies whatever element they want and the rendering of that is consistent regardless of the type (see https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Accordion/AccordionItem.js#L73)?

I think a more apt comparison would be a `Modal`, which is a similar type of component.  `Modal`s have headings, and `Accordion` are effectively just lists of content. I can see why Carbon wouldn't want to impose a heading on an `Accordion` because those can be ANYWHERE in the page hierarchy, whereas with a `Modal` (and a panel, in this case) we are talking about portals that are effectively new page sections.

The `Modal` uses an `h2` when heading text is providing: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L347

So in this case, we would be accommodating both scenarios with the `node` -- the consumer provides their own heading element, or whatever the case may be.... or the consumer provides text and we make it more understandable for assistive tech with an `h2` element.

_____

> If the purpose of this is to help with assistive tools, perhaps there are ARIA roles that can be leveraged on the element to prevent this kind of extra type checking?

For a string: I'm struggling to think of a situation where a native heading element is less ideal than an ARIA label. 🤔 

From MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#Best_practices

> The best way to use this role is to not use it at all, and instead use the native heading tags <h1> through <h6> as shown in the example above. The heading role and aria-level attribute should only be used to retrofit accessibility on legacy code that you cannot make major changes to.

For a node: This is introducing a different scenario were we add an ARIA label to any `node` passed in. That can be done with `aria-level` *but* while IMO this change is quick and easy, that change I think needs to be thought out more. Again, because it would be a `node`, it could be something other than a `span` or `p` or `div` -- we would need to make sure we aren't doing more harm than good by adding the label.

----

IMO the most straight-forward thing to do is to use built-in capabilities to check for a string & add a native heading element (that matches `Modal` structure as well) when appropriate 👍

And if someone is inserting a `node` on their own, we should leave it open for them to apply the `node`s they  thing are appropriate... which could include a different heading element, for example. But I think think carefully about trying to modify that node, as it's meant for flexibility.

SimonFinney:
Those are fair points, and I agree that the modal is a far better example - In this situation, it looks like Carbon wrap their `modalLabel` in a `h2` regardless of the type that's passed in (https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L56), and I would prefer an approach like this versus additional type checking.

jendowns:
@SimonFinney True but that seems riskier to me --

Since it's a `node` type prop, what if someone passes in a *different* heading element? What if someone uses a conflicting `aria-level`?  The `node` should allow for this type of flexibility, but right now that's an undocumented conflict that can happen & have adverse affects on assistive tech reading & understanding page hierarchy.

This is why I think it is safest to only make the `title` an `h2` if a string is passed in. 😅

PR #95, opened 2019-09-24
docs(wizard): add detail for `validate` prop

### LINKED ISSUE #64: Wizard component doesn't easily enable primary button after form gets valid
## Bug - 

Captured from Slack channel:

> or the Wizard component there is a block of code as follows:
```
get isValid() {
    if (this.state.valid == null) {
      const valid = this.currentStep.validate(this.state.componentState);
      this.setState({ valid });

      return valid;
    }

    return this.state.valid;
  }
```

> This is used to disable/enable the primary action in the component. I’m trying to pass a validate method that returns false in the initial render of the component and then turns true after my form is valid, however since this method continually returns false (since it’s not null anymore and this.currentStep.validate can’t trigger again)  my primary button never enables. The only way to trigger this validation again is to use setState in the renderMain and trigger it that way.
The documentation may need to be clearer, or API cleaned to allow the return of true to enable the button automatically.
 
```
/** @type {function(state: Object): boolean} Validator. When it returns true, the "Next" or "Save" button becomes active * depending on whether you are using a sequential or not. */
  validate: PropTypes.func,
```

**Expected behavior -**

**Actual behavior -**

### Steps for reproducing

1.  
2.  
3.

SimonFinney:
## Affected issues

- Resolves #64 

## Proposed changes

- Add detail for `Wizard` `validate` prop

### DIFF
--- a/src/components/Wizard/WizardStep.js
+++ b/src/components/Wizard/WizardStep.js
@@ -34,8 +34,8 @@ WizardStep.propTypes = {
   /** @type {function(state: Object): Promise} this is called after each step. You can do async work here or save your changes
    * depending on whether you are using a sequential wizard or not. */
   next: PropTypes.func,
-  /** @type {function(state: Object): boolean} Validator. When it returns true, the "Next" or "Save" button becomes active
-   * depending on whether you are using a sequential or not. */
+
+  /** @type {function(state: Object): boolean} Pure function for the contents of the wizard step that determines the overall wizard's state. When it returns `true`, parts of the wizard (buttons, navigation, etc.) become active. */
   validate: PropTypes.func,
 };
 
--- a/src/components/Wizard/WizardStep.js
+++ b/src/components/Wizard/WizardStep.js
@@ -35,7 +35,7 @@ WizardStep.propTypes = {
    * depending on whether you are using a sequential wizard or not. */
   next: PropTypes.func,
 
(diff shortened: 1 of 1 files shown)


src/components/Wizard/WizardStep.js
```diff
@@ -34,8 +34,8 @@ WizardStep.propTypes = {
   /** @type {function(state: Object): Promise} this is called after each step. You can do async work here or save your changes
    * depending on whether you are using a sequential wizard or not. */
   next: PropTypes.func,
-  /** @type {function(state: Object): boolean} Validator. When it returns true, the "Next" or "Save" button becomes active
-   * depending on whether you are using a sequential or not. */
+
+  /** @type {function(state: Object): boolean} Pure function for the contents of the wizard step that determines the overall wizard's state. When it returns `true`, parts of the wizard (buttons, navigation, etc.) become active. */
```
mister-what: It is important to mention, that this function runs with every state change of the wizard (ie. after every keystroke).
```suggestion
  /** @type {function(state: Object): boolean} Predicate to validate the wizards state of the current step. This function must be pure and synchronous. Validation happens after every state change (i.e. the passed function is being called after every keystroke of a user). Don't use this as a hook to trigger any side-effects, nor allow the return value to depend on anything else than the argument. When it returns `true`, parts of the wizard (buttons, navigation, etc.) become active. */
```

PR #97, opened 2019-09-24
feat: update to latest versions for all Carbon packages

### LINKED ISSUE #77: Upgrade to `carbon-components@10.6.1` & `carbon-components-react@7.6.1`
Release info here: https://github.com/carbon-design-system/carbon/releases/tag/v10.6.1

Need to upgrade to the following: 
`carbon-components@10.6.1`
`carbon-components-react@7.6.1`

jendowns:
## Affected issues

- Resolves #77

## Proposed changes

- update all Carbon package to the latest versions available
- add a `Checkbox` style override (plus add note about fix required in Carbon)

🚨 **NOTE:** @SimonFinney what do you think about this being a `feat`? I think it should be, since the Carbon versions themselves are non-trivial updates 🤔 But what do you think?


SimonFinney:
Thanks @jendowns ! I definitely agree a `feat` makes sense here, considering we're jumping 2 minor versions - I'll look at manually verifying against changes outlined in:

- https://github.com/carbon-design-system/carbon/releases/tag/v10.5.0
- https://github.com/carbon-design-system/carbon/releases/tag/v10.6.0
- https://github.com/carbon-design-system/carbon/releases/tag/v10.6.1

PR #100, opened 2019-09-26
feat(component): namespace and privatise `component` mixin

### LINKED ISSUE #99: `component` mixin causes namespace conflicts in consumers Sass
## Bug - `component` mixin causes namespace conflicts in consumers Sass

**Expected behavior -**

A consumer should be able to create a mixin without it causing unwanted side effects by importing `@carbon/ibm-security`.

**Actual behavior -**

As per @j1mie , defining a mixin named `component` and importing `@carbon/ibm-security` Sass causes that mixin to be overridden.

### Steps for reproducing

1.  Define a mixin named `component`
2.  Import `@carbon/ibm-security` afterwards
3.  Include the `component` mixin
4. See the mixin being overwritten

SimonFinney:
## Affected issues

- Resolves #99 

## Proposed changes

- Rename `component` mixin and mark as private in SassDoc
- Prevent `$namespace` variable from being overwritten and mark as private in SassDoc

## Testing instructions

-

### DIFF
--- a/src/components/Button/_index.scss
+++ b/src/components/Button/_index.scss
@@ -10,6 +10,6 @@
 
 @import 'mixins';
 
-@include component($name: button) {
+@include security--component($name: button) {
   @include button;
 }
--- a/src/components/Card/CardSkeleton/_index.scss
+++ b/src/components/Card/CardSkeleton/_index.scss
@@ -10,6 +10,6 @@
 
 @import 'mixins';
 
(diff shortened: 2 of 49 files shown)


jendowns:
@SimonFinney @j1mie I wonder if this PR should be a `feat` since it is taking two things "away" (the component mixin and the `security` prefix`) and marking them as private, whereas they weren't previously 🤔

Issue #104, opened 2019-09-29
Warnings appear for InlineLoading.js


stvcisco:
## Bug - 
While running an app that uses @carbon/ibm-security I am seeing 2 warnings  in the browser console related to `node_modules/@carbon/ibm-security/node_modules/carbon-components-react/.../components/InlineLoading/InlineLoading.js`
1) Warning: Invalid arguments supplied to oneOf, expected an array, got 4 arguments. A common mistake is to write oneOf(x, y, z) instead of oneOf([x, y, z]).
2) Warning: The prop `success` for InlineLoading has been deprecated in favor of `status`. Please use `status="finished"` instead.

I should not be seeing warnings.

The warning seem simple and correct.  
1) see line 91 of the `InlineLoading.js` source file. Prototypes.oneOf should be given a single array as its parameter.  Just wrap the 4 strings in square brackets to fix.
2) Line 86 injects this warning. I believe it's caused by having the default value of `success=false` being set in `InlineLoading.defaultProps`.  Just delete that default value since it's no longer needed. 

**Actual behavior -**
Warnings appear in the console indicating a problem in the component that should be corrected.

### Steps for reproducing
Any app that uses InlineLoading should show this issue.

### Affected browsers
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.90 Safari/537.36
But is likely not browser-specific

### Optional information
Node v10.16
Next.js v8.1.0

SimonFinney:
@stvcisco Thanks for raising! As this is a larger Carbon issue, I'll propose the change there.

stvcisco:
when is the next release expected? I moved up to @carbon/ibm-security 1.3.0 and still see the warnings

SimonFinney:
@stvcisco Thanks for your patience - we need to upgrade (https://github.com/carbon-design-system/carbon/releases/tag/v10.6.4) as either part of this work item or in a separate PR to resolve the upstream Carbon warnings.

Can you confirm the severity on your end and we can prioritise accordingly?

stvcisco:
@SimonFinney Low severity - it's just throwing stuff into the console that would be nice to be free of. Not blocking us.

SimonFinney:
Thanks! We have a work item for this sprint to upgrade - https://github.com/carbon-design-system/ibm-security/issues/117

PR #106, opened 2019-09-30
fix(status-indicator): update deprecated `InlineLoading` prop

SimonFinney:
## Affected issues

- Resolves partially #104

## Proposed changes

- Update deprecated `success` `InlineLoading` prop in `StatusIndicator` to `status`

### DIFF
--- a/src/components/InlineLoading/InlineLoading.stories.js
+++ b/src/components/InlineLoading/InlineLoading.stories.js
@@ -8,14 +8,18 @@ import { storiesOf } from '@storybook/react';
 import { withA11y } from '@storybook/addon-a11y';
 import { action } from '@storybook/addon-actions';
 import centered from '@storybook/addon-centered/react';
-import { boolean, number, text } from '@storybook/addon-knobs';
+import { number, select, text } from '@storybook/addon-knobs';
 
 import { components } from '../../../.storybook';
 
 import { Button, InlineLoading } from '../../';
 
 const props = () => ({
-  success: boolean('Loading successful state (success)', false),
+  status: select(
+    'Loading status (status)',
+    ['inactive', 'active', 'finished', 'error'],
+    'active'
+  ),
   description: text(
     'Loading progress description (description)',
     'Loading data...'
@@ -92,7 +96,7 @@ storiesOf(components('InlineLoading'), module)
                 <InlineLoading
                   style={{ marginLeft: '1rem' }}
                   description="Submitting..."
-                  success={success}
+                  status={success ? 'finished' : 'active'}
                 />
               ) : (
                 <Button onClick={handleSubmit}>Submit</Button>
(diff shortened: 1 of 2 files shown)


SimonFinney:
@jendowns Good point! I'll look at this in the morning

SimonFinney:
@jendowns Absolutely! Since this table is generated from the available component prop types, it will continue to be displayed until Carbon has removed it from its definition - https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/InlineLoading/InlineLoading.js#L84

jendowns:
@simonfinney Not necessarily — one solution is to just redefine the props inside the story & that would give the prop table accurate info

SimonFinney:
@jendowns I'm not sure I follow, but that might help avoid confusion in cases like this! Could you provide an example and I can try it out?

jendowns:
@SimonFinney I hear you, but I don't think it's necessarily something wrong with the component's props itself... I think there is something strange going on with something *WE* have done. I think it is related to the way we import the component (it is proxied) and the fact that we don't appear to have the right docgen info (again, because it is proxied?). 

But I think it would be a reasonable quickfix to redefine props in the docgen info in the story, like this. That way people have accurate information. It's a little copying & pasting for us, but honestly I think giving people accurate info is important in this case:

```
InlineLoading.__docgenInfo = {
  props: {
    className: {
      type: 'string',
      description: 'Specify a custom className to be applied to the container node',
    },
    success: {
      type: 'string',
      description: 'This prop has been deprecated in favor of `status`. Please use `status="finished" instead',
    },
    status: {
      type: 'oneOf \'default\' | \'field\' | \'large\' | \'small\'',
      description: 'Specify the loading status.',
    },
    description: {
      type: 'string',
      description: 'Specify the description of the inline loading text.'
    },
    iconDescription: {
      type: 'string',
      description: 'Specify the description for the inline loading text.'
    },
    onSuccess: {
      type: 'function',
      description: 'Provide an optional handler to be inovked when <InlineLoading> is successful',
    },
    successDelay: {
      type: 'number',
      description: 'Provide a delay for the `setTimeout` for success.',
      default: { value: 1500 }
    }
  },
};
```

(And what's nice about this route is you aren't actually messing with any proptypes they've set on the component... so the warning still shows up if you use `success`, for example)

SimonFinney:
@jendowns Nice investigation - I hadn't realized this is an issue across all components we proxy from Carbon! I don't believe this was always the case?

Either way, we definitely need to create a separate issue for this, and find a larger solution instead of trying to stay in sync with ALL of Carbon's props and story information - it's not a good use of our time. 😱

Anyway, the change you initially requested is here - https://github.com/carbon-design-system/ibm-security/pull/106/files#diff-c41b47d28315f6b9c09af61d6e042d68R99

PR #108, opened 2019-09-30
feat(button): add ghost danger button

### LINKED ISSUE #19: Add "ghost danger" button variation
### Proposal

We have a "ghost danger" button in the Tearsheet, but application teams are going to need the "ghost danger" button in other contexts as well. Therefore, it would be best if we broke it out into a reusable component.

b3nk3:
## Affected issues

- Resolves #19 

## Proposed changes

- Adds 'ghost-danger' option to Button component
- Removes current ghost danger implementation from Tearsheet and uses the one implemented on Button

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -18,16 +18,23 @@ const { defaultProps, propTypes } = CarbonButton;
 
 // TODO: V3 - Remove deprecated props `largeText`.
 
-const Button = ({ className, largeText, size, ...other }) => {
+const Button = ({ className, largeText, size, kind, ...other }) => {
   const isSize = value => size === value;
   const isLarge = isSize('large') || largeText || isSize('lg') || isSize('xlg');
+
+  const isKind = value => kind === value;
+  const isGhostDanger = isKind('ghost-danger');
+
+  console.log(!isGhostDanger ? kind : null);
 
   return (
     <CarbonButton
       className={classnames(namespace, className, {
         [`${namespace}--large`]: isLarge,
+        [`${namespace}--ghost-danger`]: isGhostDanger,
       })}
       size={!isLarge ? size : null}
+      kind={!isGhostDanger ? kind : 'ghost'}
       {...other}
     />
   );
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -22,7 +22,6 @@ const Button = ({ className, largeText, size, kind, ...other }) => {
   const isSize = value => size === value;
   const isLarge = isSize('large') || largeText || isSize('lg') || isSize('xlg');
 
-  const isKind = value => kind === value;
 
 const isGhostDanger = kind === 'ghost-danger';
 
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -23,7 +23,8 @@ const Button = ({ className, largeText, size, kind, ...other }) => {
   const isLarge = isSize('large') || largeText || isSize('lg') || isSize('xlg');
 
   const isKind = value => kind === value;
-  const isGhostDanger = isKind('ghost-danger');
+
+const isGhostDanger = kind === 'ghost-danger';
 
 
   return (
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.js
```diff
@@ -18,16 +18,23 @@ const { defaultProps, propTypes } = CarbonButton;
...
 
-const Button = ({ className, largeText, size, ...other }) => {
+const Button = ({ className, largeText, size, kind, ...other }) => {
   const isSize = value => size === value;
   const isLarge = isSize('large') || largeText || isSize('lg') || isSize('xlg');
 
+  const isKind = value => kind === value;
+  const isGhostDanger = isKind('ghost-danger');
+
+  console.log(!isGhostDanger ? kind : null);
```
jendowns: Could you remove this `console.log()`?

src/components/Button/Button.js
```diff
@@ -18,16 +18,23 @@ const { defaultProps, propTypes } = CarbonButton;
...
+
   return (
     <CarbonButton
       className={classnames(namespace, className, {
         [`${namespace}--large`]: isLarge,
+        [`${namespace}--ghost-danger`]: isGhostDanger,
       })}
       size={!isLarge ? size : null}
+      kind={!isGhostDanger ? kind : 'ghost'}
```
jendowns: Just checking -- are you keeping the `kind='ghost'` in this case to get the Carbon ghost button classes?

b3nk3: Yes, as that is how the implementation worked in the Tearsheet component.

I am not too familiar with the carbon system, so there might be a more efficient and future proof way of doing this, if so, do let me know.

src/components/Button/Button.js
b3nk3: Ooops, forgot about that. Soz.

src/components/Button/Button.js
```diff
@@ -18,16 +18,22 @@ const { defaultProps, propTypes } = CarbonButton;
 
 // TODO: V3 - Remove deprecated props `largeText`.
 
-const Button = ({ className, largeText, size, ...other }) => {
+const Button = ({ className, largeText, size, kind, ...other }) => {
   const isSize = value => size === value;
   const isLarge = isSize('large') || largeText || isSize('lg') || isSize('xlg');
 
+  const isKind = value => kind === value;
+  const isGhostDanger = isKind('ghost-danger');
```
SimonFinney: - [ ] The `isKind` method is redundant here and can be removed if it's not being used elsewhere:

```suggestion

const isGhostDanger = kind === 'ghost-danger';
```

src/components/Button/Button.stories.js
```diff
@@ -28,6 +28,7 @@ const kinds = {
   'Secondary button (secondary)': 'secondary',
   'Danger button (danger)': 'danger',
   'Ghost button (ghost)': 'ghost',
+  'Ghost Danger button (ghost-danger)': 'ghost-danger',
```
SimonFinney: - [ ] I recommend using standard case type here for consistency:

```suggestion
'Ghost danger button (ghost-danger)': 'ghost-danger',
```

src/components/Button/Button.js
b3nk3: Thanks for this, you are absolutely right.

src/components/Button/Button.js
```diff
@@ -18,16 +18,23 @@ const { defaultProps, propTypes } = CarbonButton;
 
 // TODO: V3 - Remove deprecated props `largeText`.
 
-const Button = ({ className, largeText, size, ...other }) => {
+const Button = ({ className, largeText, size, kind, ...other }) => {
   const isSize = value => size === value;
   const isLarge = isSize('large') || largeText || isSize('lg') || isSize('xlg');
 
+  const isKind = value => kind === value;
```
SimonFinney: - [ ] Thanks! The final thing here is to remove this line now it's no longer needed:

```suggestion
```

PR #111, opened 2019-09-30
fix(button): amend PropTypes

### LINKED ISSUE #76: Button - all props are not displayed in Storybook proptable
## Bug - 

**Expected behavior -**

All `Button` props are shown in the storybook prop table.

**Actual behavior -**

Only two of the many `Button` props are shown in the storybook prop table: https://ibm-security.carbondesignsystem.com/?path=/story/components-button--default

### Optional information

**Version -** `1.1.0`

b3nk3:
## Affected issues

- Resolves #76 

## Proposed changes

* refactored PropTypes import
* removed propTypes/defaultProp import from CarbonButton
* removed spread from propTypes and defaultProps
* added missing props to PropTypes
* added default props as per Story defaults

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -71,6 +71,9 @@ Button.propTypes = {
    */
   size: PropTypes.oneOf(['default', 'field', 'large', 'small']),
 
+  /**
+   * Deprecated prop. Use 'size' instead.
+   */
   largeText: deprecatedProp('size="large"', PropTypes.bool),
 };
 
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -4,7 +4,7 @@
  */
 
 import classnames from 'classnames';
-import { bool, oneOf } from 'prop-types';
+import PropTypes from 'prop-types';
 import React from 'react';
 
 import CarbonButton from 'carbon-components-react/lib/components/Button';
@@ -13,8 +13,6 @@ import { getComponentNamespace } from '../../globals/namespace';
 import deprecatedProp from '../../globals/prop-types';
 
 export const namespace = getComponentNamespace('button');
-
-const { defaultProps, propTypes } = CarbonButton;
 
 // TODO: V3 - Remove deprecated props `largeText`.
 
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.js
```diff
@@ -33,17 +31,48 @@ const Button = ({ className, largeText, size, ...other }) => {
...
 
   // It's not possible to add to Carbon's values here, so the `PropType` is recreated to include the large variant.
-  size: oneOf(['default', 'field', 'large', 'small']),
-  largeText: deprecatedProp('size="large"', bool),
+  /**
+   * Choose the size of the button
+   */
+  size: PropTypes.oneOf(['default', 'field', 'large', 'small']),
+
+  largeText: deprecatedProp('size="large"', PropTypes.bool),
```
jendowns: Could you add a description for `largeText` to indicate that it is deprecated, so users can see that message in the storybook prop table? Something short like this would work:

```suggestion
  /**
   * Deprecated prop. Use 'size' instead.
   */
  largeText: deprecatedProp('size="large"', PropTypes.bool),
```

src/components/Button/Button.js
```diff
@@ -33,17 +31,50 @@ const Button = ({ className, largeText, size, ...other }) => {
   );
 };
 
+/* eslint-disable react/require-default-props */
 Button.defaultProps = {
```
SimonFinney: - [ ] As we build on top of Carbon's `Button`, these `defaultProps` should reflect Carbon's, with the exception of `largeText`, which should remain as-is - https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Button/Button.js#L204

src/components/Button/Button.js
```diff
@@ -33,17 +31,50 @@ const Button = ({ className, largeText, size, ...other }) => {
...
   largeText: null,
+  icon: 'None',
+  kind: 'primary',
+  disabled: false,
+  size: 'default',
 };
 
 Button.propTypes = {
-  ...propTypes,
+  /**
```
SimonFinney: - [ ] As we build on top of Carbon's `Button`, these `propTypes` should reflect Carbon's, with the exception of `largeText` and `size` (as we introduce a `large` variant), which should remain as-is - https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Button/Button.js#L105

PR #112, opened 2019-10-01
fix(combo-button): add href prop and filter out falsey combo button item children

jendowns:
**NOTE:** This PR is required for some sprint work right now, so please let me know what you think as soon as you can review it 🙏 

Namely the big thing is the `href` prop & filtering for truthy `ComboButtonItem` children.

For truthiness in particular, there are definitely situations where an `undefined`, `null`, or empty string child is passed into `ComboButton`. For example, this came up in sprint work yesterday:

```jsx
// Note: in this case we are using lodash,
// but could also be a ternary where if/else comes out to undefined
const doesThingExist = lodashGet(object, 'thing', undefined)

<ComboButton>
  {doesThingExist && (
    <ComboButtonItem>
      Blah
    </ComboButtonItem>
  )}
  <ComboButtonItem>
    Bloo
  </ComboButtonItem>
</ComboButton>
```

So `doesThingExist` would appropriately be `undefined` in some cases (or other examples it could be another falsey value like `null` or an empty string). And because currently the `ComboButton` assumes everything inside the `children` array is truthy, it will fail in the case above. (Note: the workaround here would be to set `doesThingExist` to an empty string & have the check be `doesThingExist.length > 0` -- *however* that is not ideal. `undefined` or `null` would be more appropriate given the context, so the `ComboButton` should allow these types of checks instead of error'ing out hard.)

Filtering for truthy values dumps anything `undefined` etc. Quick console example, with the `hello` string filtered appropriately:

Screen Shot 2019-09-30 at 9 18 40 PM

## Proposed changes

- allow `href` prop which is used for the `ComboButtonItem`'s `Button` and `OverflowMenuItem`s
- filter for truthy `ComboButtonItem`s


src/components/ComboButton/ComboButton.js
```diff
@@ -55,9 +58,10 @@ const ComboButton = ({ children, className, direction }) => {
       <Button
         className={classnames(className, `${namespace}--primary`)}
         disabled={disabled}
+        href={href}
         iconDescription={iconDescription}
         kind="primary"
-        key={button.id}
+        key={children}
```
SimonFinney: Would you mind explaining the reasoning behind this and https://github.com/carbon-design-system/ibm-security/pull/112/files#diff-3773b98c8952dbe82cae025bb1042247R113 ?

As React recommends keys to be strings ( https://reactjs.org/docs/lists-and-keys.html#keys ), we can't be entirely certain that the `children` for these will always be as such.

jendowns: `children` at this point is often a string, which is why I changed it from `button.id` (and similarly for other instances I changed, because errors where still being shown for no `key` n the console). But it's possible that `children` could be something other than a string so I'm just going to remove these changes from the PR so we can think about it more in a separate set of changes 👍

jendowns:
Hey @SimonFinney I made the changes you suggested above 👍 Could you take another look when you get a chance?

PR #120, opened 2019-10-03
feat(accordion): move icon to right side of accordion heading

jendowns:
**NOTE:** I'm marking this as a `feat` because the icon position changes with this PR & that would be a very noticeable to users.

Jimmy & Cameron have asked that we revert our style overrides for the accordion so that we are more in line with Carbon here: http://react.carbondesignsystem.com/?path=/story/accordion--default

## Proposed changes

- remove style overrides that change icon position
- remove font style overrides (no longer needed)

### DIFF
--- a/src/components/Accordion/_index.scss
+++ b/src/components/Accordion/_index.scss
@@ -1,53 +1,26 @@
 ////
 /// Accordion component.
 /// @group accordion
-/// @copyright IBM Security 2018
+/// @copyright IBM Security 2019
 ////
 
 @import 'carbon-components/scss/components/accordion/accordion';
 
 @include export-namespace($name: 'accordion') {
   .#{$prefix}--accordion {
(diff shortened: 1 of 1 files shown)


src/components/Accordion/_index.scss
```diff
@@ -1,53 +1,26 @@
...
     }
 
-    &__content {
-      margin-left: $carbon--spacing-06;
-
-      // TODO - needs to be fixed on Carbon's side.
-      p {
-        @include carbon--type-style($name: body-long-01);
+    // Prevents gaps between closed accordion items.
+    &__content > * {
```
jendowns: NOTE: We still need this 😅

jendowns:
@SimonFinney I'm closing this right now since we aren't ready to make changes to other affected components yet

PR #129, opened 2019-10-07
feat(build): reduce build time to less than 50 seconds

SimonFinney:
Once merged, the Netlify build command needs to be replaced with `yarn build:post:storybook`

## Proposed changes

- Remove unused icon macro
- Reorder build scripts to improve parallelism
- Isolate Storybook build command and allow it to run independently
- Ignore mocks in build output


SimonFinney:
@jendowns I believe we just need to update it via the Netlify dashboard once this PR goes in

PR #130, opened 2019-10-07
feat(data-table): add `TableToolbarDownload` component

sbutler2901:
## Proposed changes

- Adding a DataTable Toolbar button that provides support for exporting the table's data to a CSV file.

#### Importing:
```
import TableToolbarDownload from '@carbon/ibm-security/lib/components/DataTable/TableToolbarDownload';
```

#### Usage:
```
<DataTable>
  ...
  <TableToolbar>
    <TableToolbarContent>
      <TableToolbarDownload headers={this.props.headers} rows={this.props.rows}/>
    </TableToolbarContent>
  </TableToolbar>
  ...
</DataTable>
```


SimonFinney:
@sbutler2901 Thanks for putting this together! It looks like the tests are failing due to the `export-to-csv` package not being found in offline mode

sbutler2901:
@SimonFinney I've pushed a new commit containing the offline-mirror for the package. Can you verify that the exporting technique is correct? I tried to replicate the TableOverflowCell component.

src/components/DataTable/TableToolbarDownload.js
```diff
@@ -0,0 +1,91 @@
...
+    return newRow;
+  });
+
+  if (title) {
+    options.showTitle = true;
+    options.title = title;
+  }
+  options.filename = filename;
+
+  options.headers = headers.map(head => head.header);
```
SimonFinney: - [ ] You can destructure `header` here to reduce duplicating `head` - 

```suggestion
  options.headers = headers.map(({ header }) => header);
```

src/components/DataTable/DataTable.js
```diff
@@ -211,6 +212,7 @@ export const dataTableExports = {
   TableToolbarContent,
   TableToolbarSearch,
   TableToolbarMenu,
+  TableToolbarDownload,
```
SimonFinney: - [ ] Would you also mind exporting this from `src/index`?

src/components/DataTable/TableToolbarDownload.js
```diff
@@ -0,0 +1,91 @@
...
+
+TableToolbarDownload.propTypes = {
+  /**
+   * @type {Object.<Object, *>}
+   * The `rows` prop is where you provide us with a list of all the rows that
+   * you want to render in the table. The only hard requirement is that this
+   * is an array of objects, and that each object has a unique `id` field
+   * available on it.
+   */
+  rows: PropTypes.arrayOf(
```
SimonFinney: - [ ] We could use `DataTable.propTypes.rows` and `DataTable.propTypes.headers` here to avoid updating these should those change

sbutler2901: @SimonFinney I have tried many different iterations none of which work when exported and used in a consuming app. I have tried things such as 
```
import { propTypes } from './DataTable';
```
with the usage as:
 ```
rows: propTypes.rows.isRequired,
headers: propTypes.headers.isRequired,
``` 
and receive the error:
```
TypeError: Cannot read property 'rows' of undefined
```
or,
```
import { DataTable } from './DataTable';
```
with the usage as:
```
  rows: DataTable.propTypes.rows.isRequired,
  headers: DataTable.propTypes.headers.isRequired
```
and receive the error:
```
TypeError: Cannot read property 'propTypes' of undefined
```

To note, the package builds fine without any issues.

Comment (SimonFinney) [reply src/components/DataTable/TableToolbarDownload.js]: That's odd! This usage sould typically work -

```js
  rows: DataTable.propTypes.rows.isRequired,
  headers: DataTable.propTypes.headers.isRequired
```

I think this is a nice-to-have but probably not a dealbreaker for now - let me know when this ready for re-review!

sbutler2901:
@SimonFinney okay great and whats the best course of action?

PR #136, opened 2019-10-09
feat(summary-card): add component and building blocks

### LINKED ISSUE #128: Add SummaryCard component
### Details

Suggested API:

```jsx
<SummaryCard>

  <SummaryCardHeader 
    title="my title"
    status={<Tooltip renderIcon={MyIcon}>Tooltip content</Tooltip>}
   />

  <SummaryCardBody>
    Card content
  </SummaryCardBody>

  <SummaryCardFooter>
    <SummaryCardAction renderIcon={MyIcon} expandsDetails>Primary action</SummaryCardAction>
    <SummaryCardAction renderIcon={MyIcon}>Secondary action</SummaryCardAction>
    <SummaryCardAction renderIcon={MyIcon} iconOnly>Icon action</SummaryCardAction>
    <SummaryCardAction renderIcon={MyIcon} iconOnly disabled>Icon action</SummaryCardAction>

    <SummaryCardDetails>
      Expanded area content
    </SummaryCardDetails>
  </SummaryCardFooter>

</SummaryCard>
```

jendowns:
## Affected issues

- Resolves #128 

### 🚨 **NOTE:** a couple features have been left out of this first pass, in order to keep the PR small --
* making the main body of the card clickable (and able to apply a custom link component -- this is likely already doable since the body subcomponent can accept nodes as children, but it's worth looking into this anyway & including a demo in storybook)
* an overflow menu action for the footer -- this may need to be a new subcomponent 🤔 

## Proposed changes

- add `SummaryCard` component with subcomponents


src/components/SummaryCard/SummaryCardFooter/SummaryCardFooter.js
```diff
@@ -0,0 +1,30 @@
+/**
```
SimonFinney: 👏

src/components/SummaryCard/SummaryCardHeader/SummaryCardHeader.js
```diff
@@ -0,0 +1,37 @@
...
+ * @copyright IBM Security 2019
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import { getComponentNamespace } from '../../../globals/namespace/index';
+
+const namespace = getComponentNamespace('summary-card-header');
```
SimonFinney: - [ ] The appropriate BEM syntax for this would be `summary-card__header`

src/components/SummaryCard/SummaryCardSkeleton/SummaryCardSkeleton.js
```diff
@@ -0,0 +1,38 @@
+/**
```
SimonFinney: 👏

src/components/SummaryCard/_mixins.scss
```diff
@@ -0,0 +1,160 @@
...
+/// @copyright IBM Security 2019
+////
+
+@import '../../globals/index';
+
+@mixin summary-card {
+  /// Summary card header height.
+  /// @type Length
+  /// @access private
+  $header--height: carbon--mini-units(4);
```
SimonFinney: For consistency, can we either use keyword arguments (https://sass-lang.com/documentation/at-rules/mixin#keyword-arguments) here or remove parts where we are using them? For example - https://github.com/carbon-design-system/ibm-security/pull/136/files#diff-c20efeb2ceca6394e0d14940502bcfd5R32

src/components/SummaryCard/_mixins.scss
```diff
@@ -0,0 +1,160 @@
...
+    background-color: $ui-01;
+  }
+
+  &-action {
+    // Required for truncation:
+    display: inline-block;
+
+    flex: 1;
+
+    &.#{$prefix}--btn--icon-only {
```
SimonFinney: We may be able to better control this selector by checking for `hasIconOnly` in the component https://github.com/carbon-design-system/ibm-security/pull/136/files#diff-da5aee03e9dbd92fefc9e73c67712adcR48 and applying our own class name here

src/components/SummaryCard/_mixins.scss
```diff
@@ -0,0 +1,160 @@
...
+      justify-content: center;
+      width: $footer--height;
+      flex: 0 0 auto;
+
+      &,
```
SimonFinney: Would you mind explaining what these rules are doing?

src/components/SummaryCard/SummaryCard.stories.js
```diff
@@ -0,0 +1,90 @@
...
+  SummaryCardAction,
+  SummaryCardBody,
+  SummaryCardFooter,
+  SummaryCardHeader,
+  SummaryCardSkeleton,
+  Tooltip,
+} from '../..';
+
+storiesOf(patterns('SummaryCard'), module)
+  .addDecorator(withA11y)
```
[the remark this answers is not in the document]
jendowns: It's definitely possible, but I don't think it's as valuable as showing how this card fits into the grid (which would be how it's used in product as well.) I can add an offset though, or add more cards in a single row? 🤔

diego-codes:
Hi there! API-wise, I think the component looks good. I'm curious what happens if more than one `SummaryCardAction` has a `expandedContent` prop.

diego-codes:
The design seems to be off from the design mocks. When the expanded area is open:
 - the expanded button text becomes header of the expanded area content.
 - the footer buttons are hidden away.
 - there is an overlay that covers the initial content of the summary card.

jendowns:
Thanks @diego-codes! 
* "the expanded button text becomes header of the expanded area content."
  - Oh right, I'll add that 👍 But worth noting I can only do that is the `children` to `SummaryCardAction` is `typeof === 'string'` otherwise if it is an unexpected node then we might have some problems 🤔 Any objections to that?

* "the footer buttons are hidden away."
  - Yeah I reached out to Jimmy about that -- I don't think it's a good idea to hide these buttons, since the overlay itself isn't like a traditional modal that focus traps (it's more like expanded accordion content). If we hide the buttons, then it's very likely someone can tab past this overlay and get "stuck" behind it, without realizing "where" they are. So he said it was all right with him to keep the buttons visible (I do still need to get his design review once I finish addressing feedback)

* "there is an overlay that covers the initial content of the summary card."
- What do you mean by this? There is an overlay that can open right now, if you click on that first button in the footer... or are you referring to something else?

diego-codes:
Wow you're fast!
- "the expanded button text becomes header of the expanded area content.": Great!
- "the footer buttons are hidden away.": Makes sense
- "there is an overlay that covers the initial content of the summary card.": Sorry I mean, that the main content of the summary card is expected to be darkened. Think of a modal or tearsheet that darkens the content behind this.

src/components/SummaryCard/SummaryCardAction/SummaryCardAction.js
```diff
@@ -0,0 +1,116 @@
...
+import Button from '../../Button';
+import Transition from '../../Transition';
+
+import { getComponentNamespace } from '../../../globals/namespace/index';
+
+const { prefix } = settings;
+
+const namespace = getComponentNamespace('summary-card-action');
+
+const transitionSegment = 150; // duration--moderate-01
```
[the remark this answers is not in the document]
jendowns: I thought so at first... but only cubic-beziers are exported? https://github.com/carbon-design-system/carbon/tree/master/packages/motion#javascript

src/components/SummaryCard/SummaryCardBody/SummaryCardBody.js
```diff
@@ -0,0 +1,39 @@
...
+
+import ScrollGradient from '../../ScrollGradient';
+
+import { getComponentNamespace } from '../../../globals/namespace/index';
+
+const namespace = getComponentNamespace('summary-card-body');
+
+const SummaryCardBody = ({ children, className, scrollGradientColor }) => (
+  <div className={classnames(namespace, className)}>
+    <ScrollGradient color={scrollGradientColor}>{children}</ScrollGradient>
```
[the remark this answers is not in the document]
jendowns: Yeah that was the goal but now that you ask I'm not sure if it's necessary? 🤔 (Think I should remove it?)

src/components/SummaryCard/_mixins.scss
jendowns: According to the design spec these icon-only buttons need to have white icons, but default button styles set them to blue40. The overrides have to be aggressive in this case because the defaults set blue40 for hover via `currentColor` AND then for hover on the icon path. 😭

src/components/SummaryCard/_mixins.scss
jendowns: What do you mean about about consistency etc? 🤔

src/components/SummaryCard/SummaryCard.stories.js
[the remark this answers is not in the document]
SimonFinney: That's a good point! Perhaps multiple cards are the best route if that's the intended usage

src/components/SummaryCard/_mixins.scss
SimonFinney: Consistency in how we call methods, for example `carbon--mini-units($count: 4)` versus `carbon--mini-units(4)`

jendowns:
@SimonFinney This should be ready to review again. I believe I've addressed all of the feedback above?

The one question I have at the moment is -- do you prefer the `SummaryCard` examples in the same story, in a grid? Or alone as separate story variations?

src/components/SummaryCard/_mixins.scss
```diff
@@ -0,0 +1,228 @@
...
+      @include carbon--type-style($name: body-long-01);
+      margin: 0;
+    }
+
+    &__title {
+      padding: $carbon--spacing-05;
+    }
+
+    &__content {
+      height: $body--height - $header--height - $footer--height - 1rem;
```
SimonFinney:
```suggestion
	      height: $body--height - $header--height - $footer--height - carbon--mini-units($count: 2);
```

src/components/SummaryCard/_mixins.scss
```diff
@@ -0,0 +1,228 @@
...
+        color: $text-01;
+
+        .#{$prefix}--btn__icon path {
+          fill: $icon-01;
+        }
+      }
+
+      .#{$prefix}--btn__icon {
+        margin-left: 0;
+        height: carbon--mini-units(2.5);
```
SimonFinney:
```suggestion
        height: carbon--mini-units($count: 2.5);
```

src/components/SummaryCard/_mixins.scss
```diff
@@ -0,0 +1,228 @@
...
+        width: carbon--mini-units(2.5);
```
SimonFinney:
```suggestion
        height: carbon--mini-units($count: 2.5);
```

PR #138, opened 2019-10-09
feat(stringformatter): enable multiline clamping

j1mie:
## Proposed changes

- Enables mutli line clamping for string formatter

[Note on `line-clamp` and its browser support](https://caniuse.com/#search=line-clamp) - how does this align to this library?


j1mie:
@SimonFinney Please review

PR #151, opened 2019-10-11
fix(status-indicator): remove additional paddings for incomplete icon

### LINKED ISSUE #148: StatusIndicator - "incomplete" icon is very small
**Expected behavior -**

The "incomplete" icon in the `StatusIndicator` should be the same size as the other icons. (20px)

**Actual behavior -**

The icon is VERY tiny right now and does not match the other status icons: https://ibm-security.carbondesignsystem.com/?path=/story/patterns-statusindicator--default

Screen Shot 2019-10-10 at 4 57 23 PM

I saw this in Chrome and Firefox

### Steps for reproducing

See the storybook demo here: https://ibm-security.carbondesignsystem.com/?path=/story/patterns-statusindicator--default

### Affected browsers

Latest Chrome & Firefox

kleyu:
## Affected issues

- Resolves #148 

## Proposed changes

- removed `padding-top` and `padding-bottom` from `incomplete` icon as it was interfering with the icon size. 

## Testing instructions

- check if the `incomplete` icon on `StatusIndicator` component is now properly sized.

## Additional notes

If someone has an idea why these paddings were decreasing the icon size, I'd greatly appreciate any input as I'm pretty clueless about this specific thing 😄 .

Also, I've noticed that the `complete` icon is 16x16 px size and it doesn't look the best IMO for example in `StatusIndicator` - `Dynamic` story, when the icon sizes change from 20x20 to 16x16. (not sure if it's supposed to be like this, or it is a bug)

### DIFF
--- a/src/components/StatusIndicator/_mixins.scss
+++ b/src/components/StatusIndicator/_mixins.scss
@@ -63,8 +63,6 @@
 
     &--incomplete #{$step}__icon {
       fill: $disabled-03;
-      padding-top: $carbon--spacing-03;
-      padding-bottom: $carbon--spacing-03;
     }
 
     &__icon-wrapper {


kleyu:
> Thank you @kleyu!
> 
> About your question -- I believe the reason removing padding stopped shrinking the icon is because we now have `box-sizing: border-box` applied in this environment. Previously without that rule, the box-modal for that icon was not accounting for the padding 🤔 (MDN has as good write up about this here: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing)
> 
> About the complete icon -- WHOOPS I didn't even notice that 😆 Yeah that icon should be the same size as the others, not 16px.

Thanks @jendowns !
I guess I just assumed that everything uses `box-sizing: border-box` and hadn't encountered a problem before with box-sizing being a major factor 😁 . I'll take a look at MDN for sure, thanks again.

PR #155, opened 2019-10-14
feat(nav-item-link): add component that accepts custom tag, include in nav

### LINKED ISSUE #31: Support for other React routers in Nav

jendowns:
## Affected issues

- Partially addresses #31 for the `Nav` only (followup to feedback from https://github.com/carbon-design-system/ibm-security/pull/33) 

## Proposed changes

- create `NavItemLink` that accepts custom tag
- pass through extra props in `NavItem` to `NavItemLink` (so that a custom routing element with its given props can be used)
- update story and snapshots

### DIFF
--- a/src/components/Nav/NavItem/NavItem.js
+++ b/src/components/Nav/NavItem/NavItem.js
@@ -6,8 +6,9 @@
 import Launch16 from '@carbon/icons-react/lib/launch/16';
 
 import classnames from 'classnames';
-import { bool, func, node, number, string } from 'prop-types';
+import { bool, elementType, func, node, number, string } from 'prop-types';
 import React, { Component } from 'react';
+import NavItemLink from '../NavItemLink';
 
 import Icon from '../../Icon';
 
@@ -25,7 +26,9 @@ export default class NavItem extends Component {
     className: '',
     current: null,
     disabled: false,
+    element: 'a',
     handleItemSelect: null,
+    href: undefined,
     id: namespace,
     label: '',
     link: true,
@@ -50,8 +53,14 @@ export default class NavItem extends Component {
     /** @type {bool} Whether the item is disabled. */
     disabled: bool,
 
+    /** @type {elementType} The base element to use to build the link. Defaults to `a`, can also accept alternative tag names or custom components like `Link` from `react-router`. */
+    element: elementType,
+
     /** @type {Function} Click handler of an item. */
     handleItemSelect: func,
+
+    /** @type {string} The href of the nav item. */
+    href: string,
 
     /** @type {string} Identifier. */
     id: string,
@@ -83,6 +92,7 @@ export default class NavItem extends Component {
   render() {
     const {
       className,
+      element,
       tabIndex,
       children,
       disabled,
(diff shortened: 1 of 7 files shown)


src/components/Nav/NavItem/NavItem.js
```diff
@@ -120,20 +131,21 @@ export default class NavItem extends Component {
...
+        label={label}
         onClick={event => handleDisabled(onClick(event, href))}
         onKeyPress={event => handleDisabled(onClick(event, href))}
         role="menuitem"
         tabIndex={handleDisabled(children ? -1 : tabIndex)}
-        {...other}
       >
         {link ? (
-          <a
+          <NavItemLink
```
SimonFinney: 👏

src/components/Nav/NavItem/NavItem.js
```diff
@@ -120,20 +131,21 @@ export default class NavItem extends Component {
...
             id={id}
             className={classnames(linkClassName, {
```
SimonFinney: - [ ] It would make sense for these classes to be part of the `NavItemLink` component

src/components/Nav/NavItemLink/NavItemLink.js
```diff
@@ -0,0 +1,25 @@
...
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const NavItemLink = React.forwardRef(function NavItemLink(props, ref) {
+  const { element, ...rest } = props;
+  return React.createElement(element, { ...rest, ref });
+});
+
+NavItemLink.displayName = 'Link';
```
SimonFinney: Is this line renaming the display name from `NavItemLink` to `Link`? If so, I think it's good to leave it as-is to avoid any confusion in test output - https://github.com/carbon-design-system/ibm-security/pull/155/files#diff-061f99714b39a9b07ec0dd9e79fd3ae7R12

src/components/Nav/NavItem/NavItem.js
jendowns: Unfortunately that would also involve moving all of the external link + icon logic into the `NavItemLink` component as well 🤔 Could we keep these classes here, since `NavItemLink` is utilized directly by `NavItem` (and not necessarily defined by the user?)

SimonFinney: Fair point - if it's specific to the `NavItem`, my gut would be to bring all of that logic with it.

However, it's more likely that this logic could be used in links elsewhere throughout the `Shell`, I would recommend moving and renaming it to something like `ShellLink`, and that way we could start to use it in other places like the `Header` in future. What are your thoughts?

jendowns: That's an interesting approach but but I'm not sure where else it would be used in the shell... 😅 where else are "items" defined (except for the inside the `Nav` in the `Toolbar`)? Everything else (like notifications) uses an object & the "list" isn't necessarily exposed. The `Nav` is the only place this type of thing happens.

jendowns: I don't see an inherit "risk" in creating this component because it's just something used by `NavItem`. It wouldn't necessarily make a `Shell` refactor difficult or impossible. It's just taking some logic out of `NavItem.js` for the forwardRef + custom element.

PR #156, opened 2019-10-14
feat(tooltip-definition): proxy component, replace usage in HeaderNotification & StringFormatter

### LINKED ISSUE #140: `HeaderNotification` content tooltip not displaying
## Bug - `HeaderNotification` content tooltip not displaying

**Expected behavior -**

The `HeaderNotification` should display a tooltip when the content is truncated to provide the full title.

**Actual behavior -**

The tooltip for the `HeaderNotification` appears to be displaying incorrectly.

### LINKED ISSUE #152: TooltipDefinition - export component, replace usage in HeaderNotification
### Details

Need to export the `TooltipDefinition` component and replace its usage in the `HeaderNotification` component.

NOTE: possibly will affect #140?

jendowns:
## Affected issues

- Resolves #152 
- Resolve #140 

@SimonFinney @j1mie  As I went to update `TooltipDefinition` usage after exporting it for #151, I saw the issues with #140. There are a lot of workarounds with the way the `TooltipDefinition` is used inside the `HeaderNotification` -- so to simplify things, I propose removing it and using a native `title`-based tooltip. And for truncation, we could make use of `StringFormatter` and remove `react-truncate`.

I realize there are technically 3 things happening in this PR so I'm happy to split it up if you want. Or if you disagree with using a native tooltip instead, just let me know. 😅 I just kept them together since they all seemed related -- one thing led to another, etc.

## Proposed changes

- proxy `TooltipDefinition` component
- in `StringFormatter` ...
  - allow for extra props (like `aria-label` or `title`, see below)
- in `HeaderNotification`...
  - replace usage of `TooltipDefinition` with native tooltip via `title` attribute
  - add `aria-label` attribute
  - replace `Truncate` from `react-truncate` with `StringFormatter`
  - remove `role='button'` on anchor element (🤔 reasoning: it's using an `href`, not an `onClick` so shouldn't need a `button` role)
  - remove unnecessary tooltip styles


package.json
```diff
@@ -60,7 +60,6 @@
     "prop-types": "^15.7.2",
     "ramda": "^0.26.1",
     "react-addons-css-transition-group": "^15.6.2",
-    "react-truncate": "^2.4.0",
```
j1mie: Good stuff! This project hasn't been updated in a while [and has some issues with breaking text](https://github.com/One-com/react-truncate/issues/16).

jendowns:
Thank you @SimonFinney -- that's a really good question. I'll look into this really quick...  if it ends up not being visible via keyboard activation, I may need to rethink this. 🤔

jendowns:
@SimonFinney @j1mie Could you both take another look at my recent updates when you get a chance? https://deploy-preview-156--ibm-security.netlify.com/?path=/story/patterns-header--with-notifications

Changes since you last viewed:
* changed native `title` (tooltip) usage in `StringFormatter` to Carbon's more accessible `TooltipDefintion` component. If `truncate={true}`, then the `TooltipDefintion` will wrap the `StringFormatter`'s inner `span
* applied the `StringFormatter`/`TooltipDefintion` changes to the `Header`'s notifications. The first notification will have a downward tooltip, whereas all others will have an upward tooltip. 

One downside of this approach -- 
The `Header` notifications are always going to have a tooltip in this case, even if they don't need one, because `truncate={true}` is set on the `StringFormatter`. 🤔 Personally I don't think this is too bad, and in the future we could look into "smartly" checking what to set `truncate` to... for now though, to get a workable and accessible tooltip on this component, it seems acceptable... but what do you two thing?

src/components/Header/HeaderNotification/_mixins.scss
```diff
@@ -22,6 +22,34 @@
   box-sizing: border-box;
   justify-content: space-between;
 
+  // Inverse styling for string formatter's tooltip defintion:
```
j1mie: Hey @jendowns can you explain this part to me please. Thanks in advance!

jendowns: Yeah so there are a LOT of very specific styles that come from the `TooltipDefintion`, and this is me overriding a lot of that... (mostly this is due to theming, because we have themed the `Header` notification popover like a "light" themed component but the `TooltipDefinition` styles are anticipating a dark themed component --

----------

```
  .#{$namespace}string-formatter__tooltip.#{$prefix}--tooltip--definition
    .#{$prefix}--tooltip__trigger {
    color: currentColor;
    text-align: left;
    line-height: initial;
...
```

☝️ Without that, the color of the text in the `StringFormatter`, inside the `TooltipDefintion`, is *white* (unreadable), centered, and has weird line-height.

----------

```
    // Tooltip top variation arrow:
    &.#{$prefix}--tooltip__trigger--definition.#{$prefix}--tooltip--top.#{$prefix}--tooltip--align-start::before {
      border-color: $inverse-01 transparent transparent transparent;
    }

    // Tooltip bottom variation arrow:
    &.#{$prefix}--tooltip__trigger--definition.#{$prefix}--tooltip--bottom.#{$prefix}--tooltip--align-start::before {
      border-color: transparent transparent $inverse-01 transparent;
    }
```

☝️  Color of the tooltip's arrow is "light" themed so I need to make it inverse color & appear dark.

----------

```
    // Tooltip content:
    & + .#{$prefix}--assistive-text,
    &:focus
      + .#{$prefix}--assistive-text
      &:hover
      + .#{$prefix}--assistive-text {
      background-color: $inverse-01;
      color: $inverse-02;
    }
  }
```

☝️ For some reason the tooltip has different colors for `:hover` and for `:focus` states, hence this selector 😆  This rule inverts the colors as mentioned previously.

src/components/Header/Header.js
```diff
@@ -337,21 +337,27 @@ export default class Header extends Component {
...
+                        clearButtonLabel={`${labels.notifications.clear} '${description}'`}
+                        dateTime={datetime}
+                        description={description}
+                        href={href}
+                        onClearButtonClick={() => this.clearNotification(id)}
+                        product={product}
+                        timeLabel={label}
+                        viaLabel={labels.notifications.via}
+                        tabIndex="0"
+                        tooltipDirection={index === 0 ? 'bottom' : 'top'}
```
SimonFinney: 👏

src/components/Header/HeaderNotification/_index.scss
```diff
@@ -4,6 +4,7 @@
 /// @copyright IBM Security 2019
 ////
 
+@import '../../StringFormatter/index';
```
SimonFinney: 👏

src/components/Header/HeaderNotification/_mixins.scss
j1mie: @jendowns really appreciate the detailed explanation, thanks for clarifying ⭐️

jendowns: Yeah no worries, I hope it wasn't too much 😆 The selectors are so long it's almost hard to read, and the prettier formatting doesn't help... 🤔

jendowns:
Thanks @SimonFinney! I just pushed a commit that adds a focus outline to those notification texts (which are tooltip triggers). I opted to add it manually, rather than use the `focus-outline` mixin, because of the inverse theming we are doing here... (currently that outline would be white, as we've set inverse color tokens for the background etc.)

src/components/Header/HeaderNotification/_mixins.scss
```diff
@@ -22,6 +22,39 @@
...
+  .#{$namespace}string-formatter__tooltip.#{$prefix}--tooltip--definition
+    .#{$prefix}--tooltip__trigger {
+    color: currentColor;
+    text-align: left;
+    line-height: initial;
+
+    &:focus {
+      outline: 2px solid $interactive-01;
+      outline-offset: 2px;
```
SimonFinney:
```suggestion
@include focus-outline($type: outline);
outline-color: $inverse-focus-ui;
```

SimonFinney: Ah, I didn't realise! In that case, that makes sense to me

PR #187, opened 2019-10-30
feat(yarn): upgrade packages

### LINKED ISSUE #137: Upgrade to Carbon `v10.7.3`
## Context

Upgrade to the latest Carbon packages and manually verify that no issues were introduced:

- https://github.com/carbon-design-system/carbon/releases/tag/v10.7.0
- https://github.com/carbon-design-system/carbon/releases/tag/v10.7.1
- https://github.com/carbon-design-system/carbon/releases/tag/v10.7.2
- https://github.com/carbon-design-system/carbon/releases/tag/v10.7.3

SimonFinney:
## Affected issues

- Resolves #137 

## Proposed changes

- Upgrade Carbon packages and manually verify against the release notes
- Apply `FilterPanel` fix for the height that was regressing from changes to the `Accordion`
- Remove CSS custom properties from Storybook
- Snapshot tests color updates
- Yarn upgrade spring-cleaning

## Testing instructions

Manual verification against the release notes and Storybook was completed but I suggest an additional pass to ensure nothing is overlooked.


SimonFinney:
@jendowns Good catch - It looks like this styling has been applied to the underlying `StringFormatter`!

I think we need to decide if we want to absorb the `TooltipDefinition` styles (dotted underline) for the `StringFormatter` if it's going to be something we're leveraging. I personally am good with that, but I'd like to hear yours and perhaps @petervachon 's thoughts too.

petervachon:
Screen Shot 2019-11-01 at 6 02 53 AM

This isa questionable one. The blue dotted line being full width looks more like an underline you would see for a link than an ellipsis typically seen to represent truncation.

SimonFinney:
Admittedly it's confusing and deceptive, especially in situations in which we have a piece of text that isn't truncated but still has a tooltip - I'll do some minor investigation, but for the purposes of this PR, we can just override the underline similar to how it previously behaved.

jendowns:
@SimonFinney Do you think this is ready to merge?

SimonFinney:
@jendowns Almost - I'm just fixing the type size and then I think we should be good!

SimonFinney:
@jendowns Changes are in!

PR #195, opened 2019-11-01
feat(storybook): add proxy proof-of-concept

### LINKED ISSUE #134: Prop tables in Storybook for proxied components
#### Context

Currently many proxied components (e.g., imported from `carbon-components-react`, sometimes modified slightly, and then exported from this library) do not have the correct prop information in Storybook.

For example, the inline `CodeSnippet` component's prop table shows all of the props, but they are missing their descriptions and correct type info: https://ibm-security.carbondesignsystem.com/?path=/story/components-codesnippet--single-line

We need a consistent way to address these incorrect prop tables that minimizes duplication and future maintenance. Ideally, this 'fix' would not require us to redefine props, which could quickly get out of date with Carbon's codebase.

###### Carbon

Carbon

###### Carbon for IBM Security - development and production environment

Code snippet

#### Acceptance criteria

- [ ] all proxied components show correct prop information in their storybook prop tables, OR instead of prop tables, there is a link to the Carbon's story with the correct prop information

##### Documentation

- [ ] `defaultProps` and `propTypes` are correct documented, OR correctly linked to

SimonFinney:
## Affected issues

- Resolves #134 - Guidance will be used for incremental refactors

## Proposed changes

- Add ability to supply Carbon story ID to redirect to in 'Show Info' panel


src/components/Accordion/Accordion.stories.js
```diff
@@ -25,6 +25,10 @@ const props = {
   style: { width: '100vw' },
 };
 
+const story = 'accordion';
+const description =
+  'Accordions allow users to expand and collapse sections of content.';
+
```
jendowns: Could you add a display name for the `Select` used in this component demo, so that it doesn't show as `[object Object] in the story source?

```suggestion

Select.displayName = 'Select';
```

SimonFinney: 🤔 I guess this is still a symptom of the maintenance issue we're having! I'll spend a bit more time debugging this, but otherwise, if it's the same build issue, we're going to need to do this for every component we proxy

SimonFinney: I guess this is still a symptom of the maintenance issue we're having! I'll spend a bit more time debugging this, but otherwise, if it's the same build issue, we're going to need to do this for every component we proxy

jendowns: Resolving this convo since the fix was proposed upstream in Carbon 👍

PR #196, opened 2019-11-01
fix(accordion): add prop for changing accordion heading alignment

### LINKED ISSUE #123: Right-aligned prop of the Accordion
## Request - I need a prop that right-aligns the chevron in the accordion that matches Carbon's current accordion. 

This direction is approved by Cameron Calder. Until a consistent component has been evaluated and approved via the CLX team, we need a way to use Carbon's current right-aligned accordion component. 

See the example image below:

Artboard

jendowns:
## Affected issues

- Resolves #123 

## Proposed changes

- wrap Carbon's `Accordion` to accept a new prop, `alignment`, and modify class names passed to component
- add new prop `alignment` that accepts `'left'` (default) or `'right'`
- update story with knob for changing alignment
- remove some unnecessary style overrides from `Accordion`


src/components/Accordion/Accordion.js
```diff
@@ -0,0 +1,51 @@
...
+
+import CarbonAccordion from 'carbon-components-react/lib/components/Accordion';
+
+import { getComponentNamespace } from '../../globals/namespace';
+
+const namespace = getComponentNamespace('accordion');
+
+const { defaultProps, propTypes } = CarbonAccordion;
+
+const Accordion = ({ className, alignment, ...rest }) => {
```
SimonFinney: - Can we call this prop `align` and use the values `start` and `end` to maintain consistency with Carbon? https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/TooltipDefinition/TooltipDefinition.js#L78
- For consistency with the majority of our spread props, can we use `...other` here?

src/components/Accordion/_index.scss
```diff
@@ -8,46 +8,36 @@
...
-
-    &__arrow {
-      margin: 0 0 0 $carbon--spacing-03;
-      transform: rotate(0deg);
-    }
-
-    &__item--active {
-      .#{$prefix}--accordion__title {
-        font-weight: 600;
+    &.#{$namespace}accordion--left {
```
SimonFinney: - [ ] Can we import `globals/namespace/index` here if we're leveraging this variable? It will be required once we finish the style modularising work - thanks!

src/components/Accordion/Accordion.js
```diff
@@ -0,0 +1,51 @@
...
+  const isRightAligned = alignment === 'right';
+
+  const accordionClasses = classnames(namespace, className, {
+    [`${namespace}--right`]: isRightAligned,
```
SimonFinney: Similar to Carbon, we can avoid redundancy in these declarations by writing:

```js
[`${namespace}--${align}`]: align
```
https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/TooltipDefinition/TooltipDefinition.js#L39

jendowns:
Thanks @SimonFinney -- I just made the changes you suggest above. Could you take another look when you get a chance?

PR #197, opened 2019-11-01
chore(styles): update sass doc group & file comments for consistency

jendowns:
## Proposed changes

- update sass doc groups & comments for consistency


src/components/DataDecorator/_index.scss
```diff
@@ -1,6 +1,6 @@
 ////
-/// DataDecorator component.
-/// @group data--decorator
+/// Decorator component.
+/// @group decorator
```
jendowns: Reasoning: the mixin is called `decorator` (not `data--decorator` etc)

PR #201, opened 2019-11-05
fix(tearsheet-small): add content overflow to prevent scrolling

SimonFinney:
## Proposed changes

- Allow the content to horizontally scroll within the tearsheet if it's not styled to break


SimonFinney:
@petervachon Agreed! This isn't ideal, but I spoke with Chris and it's regarding a time-sensitive fix

PR #205, opened 2019-11-06
fix(accordion): vertically center items in heading

jendowns:
## Proposed changes

- vertically center items in the `Accordion` heading

### DIFF
--- a/src/components/Accordion/_index.scss
+++ b/src/components/Accordion/_index.scss
@@ -29,6 +29,10 @@
       }
     }
 
+    &__heading {
+      align-items: center;
+    }
+
     &__item--active &__title {
       font-weight: 600;
     }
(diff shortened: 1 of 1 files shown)


src/components/Accordion/_index.scss
```diff
@@ -29,6 +29,10 @@
       }
     }
 
+    &__heading {
```
SimonFinney: - [ ] The change looks good to me, but I feel like this only needs to be applied to our variation, while Carbon's needs to apply to theirs

jendowns: I have a feeling that it could look confuse people somehow, but then again it's just an assumption so... better to make a smaller change at first 👍

jendowns:
Thanks @petervachon I just opened an issue for that here: https://github.com/carbon-design-system/ibm-security/issues/209

jendowns:
@SimonFinney I just pushed an update to move that style rule inside the `start` variation class only 👍

SimonFinney:
LGTM! I've just discovered this, but it exists in `dev` so we can create a defect to resolve 😆 -

Accordion

PR #207, opened 2019-11-07
feat(status-indicator):  refactor component and update class names, fix icon size and text misalignment

### LINKED ISSUE #193: `StatusIndicator` - Icons should be same size
## Bug - `StatusIndicator` - Icons should be same size

SimonFinney:
## Affected issues

- Resolves #193

## Proposed changes

- Refactor component markup and styles
- Update ghost button to include correct icon and styling
- Update snapshot tests


src/components/StatusIndicator/__tests__/__snapshots__/StatusIndicator.spec.js.snap
```diff
@@ -1,14 +1,16 @@
...
+<div
+  className="security--status-indicator"
+>
   <h1
     className="security--status-indicator__title"
   >
     Status indicator title
   </h1>
   <ul
-    className="security--status-indicator"
```
jendowns: Is there a way to avoid these class name changes? 🤔 More of a situation of "just in case" -- this component has been around for a while so I worry that a consumer may be relying on these class names for one reason or another, so this change may break a test, style override, etc.

src/components/StatusIndicator/__tests__/__snapshots__/StatusIndicator.spec.js.snap
```diff
@@ -43,40 +45,22 @@ exports[`StatusIndicator renders 1`] = `
...
+        }
+      }
       tabIndex={0}
       type="button"
     >
-      <span
-        className="security--status-indicator__step__icon-wrapper"
-      >
-        <Icon
-          className="security--status-indicator__retry-button__icon"
```
jendowns: Can we still apply this class to the rendered Icon? As in, pass in the icon as a component with this class applied? Just in case someone is relying on the class name for a test or something like that. Same goes for the wrapper.

src/components/StatusIndicator/__tests__/__snapshots__/StatusIndicator.spec.js.snap
```diff
@@ -43,40 +45,22 @@ exports[`StatusIndicator renders 1`] = `
...
-            }
-          }
-          size={20}
-          title="Retry"
-          viewBox="0 0 32 32"
-          width={null}
-        />
-      </span>
-      <span
-        className="security--status-indicator__retry-button__message"
```
jendowns: Similar comment as above -- can we keep this span with the class name?

src/components/StatusIndicator/_mixins.scss
```diff
@@ -4,102 +4,114 @@
...
-
-    &__icon {
-      fill: $interactive-04;
-    }
-
-    &__message {
-      @include carbon--type-scale($step: 2);
-      margin-left: $carbon--spacing-03;
-    }
+  &__button {
```
jendowns: Would like to still call this `retry-button` instead, if possible? (see prev comments about class names)

src/components/StatusIndicator/__tests__/__snapshots__/StatusIndicator.spec.js.snap
```diff
@@ -43,40 +45,22 @@ exports[`StatusIndicator renders 1`] = `
       status="incomplete"
     />
     <Button
-      className="security--status-indicator__retry-button"
+      className="security--status-indicator__button--retry"
```
jendowns: Hey I noticed this class isn't the same as before 😅 

Previously it was `security--status-indicator__retry-button`
How do you feel about changing it back, just in case someone's tests or styles are relying on it?

src/components/StatusIndicator/__tests__/__snapshots__/StatusIndicator.spec.js.snap
SimonFinney: Good point - I'll do a cursory search for impact and make changes where necessary

PR #219, opened 2019-11-11
fix(icon-button-bar): update menu options classes and box shadow application

### LINKED ISSUE #213: `IconButtonBar` - Drop shadow missing behind `OverflowMenu`
## Bug - `IconButtonBar` - Drop shadow missing behind `OverflowMenu`

**Expected behavior -**

There should be a drop shadow behind the `OverflowMenu`

IconButtonBar

jendowns:
## Affected issues

- Resolves #213 

## Proposed changes

- add a new class for direction of the overflow menu options
- if direction === top, then the top class will be applied & Carbon's box-shadow rule will be inverted (the inversion = `-2px` in the rule)

## Testing instructions

- The box shadow is usually very hard to see in the gray100 theme so I suggest changing the background color of the storybook demo to something else lighter.


src/components/IconButtonBar/IconButtonBar.js
```diff
@@ -88,12 +92,7 @@ const IconButtonBar = ({
             direction={overflowMenuDirection}
             flipped
             menuOptionsClass={iconButtonBarMenuOptionsClasses}
-            renderIcon={() =>
-              renderIconButton({
-                renderIcon: getOverflowMenuIcon(),
-                tooltip: false,
-              })
-            }
+            renderIcon={getOverflowMenuIcon()}
```
jendowns: I noticed that this component was rendering our `IconButton`, versus rendering the button that Carbon actually uses in the `OverflowMenu` component, which was causing some issues with our own `IconButton` classes & styles overriding their `OverflowMenu`-related classes. So that's why I made this change, which also means we have to do fewer style overrides 👍

src/components/IconButtonBar/_mixins.scss
```diff
@@ -36,13 +36,12 @@
...
-  }
-
-  &__overflow-menu-options {
-    box-shadow: unset;
 
-    &.#{$prefix}--overflow-menu-options::after {
-      height: 0;
+    &__overflow-menu-options--#{$icon-button-bar__size}.bx--overflow-menu--flip {
+      &.#{$prefix}--overflow-menu-options[data-floating-menu-direction='top']::after,
+      &.#{$prefix}--overflow-menu-options[data-floating-menu-direction='bottom']::after {
```
jendowns: The style rules being overridden are applied to: `.#{$prefix}--overflow-menu-options[data-floating-menu-direction='top']::after` & `.#{$prefix}--overflow-menu-options[data-floating-menu-direction='bottom']::after`, so that's why this rule is so specific

src/components/IconButtonBar/_mixins.scss
```diff
@@ -36,13 +36,12 @@
         width: $icon-button-bar__sizing__dimensions;
       }
     }
-  }
-
-  &__overflow-menu-options {
-    box-shadow: unset;
 
-    &.#{$prefix}--overflow-menu-options::after {
-      height: 0;
```
jendowns: This shouldn't be hidden or zeroed out, as Carbon appears to use it as a way to prevent the menu's `box-shadow` from covering part of the overflow menu button.

jendowns:
Also @SimonFinney FYI -- it's hard to see for dark themes, but for light themes you can see a slight box shadow overlap on the button itself (below that little `::after` content used to cover up the box-shadow. I opened an issue & PR about that here: https://github.com/carbon-design-system/carbon/issues/4667

cameroncalder:
This looks good 👍

PR #226, opened 2019-11-13
fix(breadcrumb): force `visited` state override

SimonFinney:
## Proposed changes

- Force `visited` state for `Breadcrumbs` to remain as default state


jendowns:
@SimonFinney Do you know where this feedback came from?

I seem to recall us going back and forth on this in the GHE repo. 🤔

SimonFinney:
This issue came up this evening in discussion with @j1mie as part of their offering - it was something we had in place, but got lost with migration to `2.x`.

Full disclosure, I think it was me! 🤣I've posted a link on Slack to the original line.

PR #228, opened 2019-11-14
feat(summary-card): add multiselect, extend `DataTable` batch actions

### LINKED ISSUE #153: Multiselect `SummaryCard`
### Details

- **Application / team -** Design team
- **Needed by -** ASAP

### Proposal

For the `SummaryCard`, we need a way for this component to be multi-selectable and combined with the batch actions that comes with the `DataTable`. Screenshots below illustrate both states:

image
image

#### A user

- [ ] can select a `SummaryCard` and begin a multiselect process
- [ ] can perform batch actions on the selected `SummaryCard`s (leveraging the `DataTable` batch actions where possible)

SimonFinney:
## Affected issues

- Resolves #153 

## Proposed changes

- Add `SummaryCardBatchAction`, `SummaryCardBatchActions`, `SummaryCardContainer`, and `SummaryCardSelect` components
- Leverage `DataTable` batch actions as much as possible
- Update `SummaryCardBody` scroll gradient color to reflect recent theme updates
- Reduce redundant `SummaryCard` import path


SimonFinney:
@jendowns Tagging you while this is still a draft to keep an eye on progress and add comments as you see fit

SimonFinney:
@jendowns This is feature complete - I'm holding off from writing unit tests until the first review! Let me know if you have any thoughts

src/components/SummaryCard/SummaryCardContainer/SummaryCardContainer.js
```diff
@@ -0,0 +1,66 @@
...
+
+import { namespace as summaryCardNamespace } from '../SummaryCard';
+import { namespace as summaryCardSelectNamespace } from '../SummaryCardSelect/SummaryCardSelect';
+
+const getSummaryCard = (summaryCards, selectedId) =>
+  summaryCards.filter(({ id }) => id === selectedId);
+
+const resetSelectedSummaryCards = state => state([]);
+
+const SummaryCardContainer = ({ render, summaryCards }) => {
```
jendowns: I'm wondering why we need `SummaryCardContainer`? Do you think this could be added to `SummaryCard` instead?

SimonFinney: Good question! This closely mimics the `DataTable` API - https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/DataTable/DataTable.js

We need the `SummaryCardContainer` to control the batch action state of all the summary cards and batch actions components, as well as return the calculated props and output for each of the sub-components, for example, `getBatchActionProps`, `getSelectionProps`, and `selectedSummaryCards`.

Admittedly the naming seems off but I wanted to follow the `DataTable` example as closely as possible - I'm entirely open to suggestions here!

jendowns: Could we start to use a `function` declaration for components constructed in this way, that use hooks? 

It matches the format of the React docs (https://reactjs.org/docs/hooks-overview.html) & separately it was brought up to me in a recent PR for Carbon core. It sounds like a direction they want to go, in terms of how their components are written/constructed, so perhaps we can start going in the same direction as well?

package.json
```diff
@@ -96,6 +96,7 @@
     "@storybook/addons": "^5.2.4",
     "@storybook/react": "^5.2.4",
     "@storybook/theming": "^5.2.4",
+    "@testing-library/react": "^9.3.2",
```
jendowns: Could we add this in a separate PR instead? 😅

src/components/SummaryCard/SummaryCard.stories.js
```diff
@@ -3,120 +3,183 @@
...
+    <SummaryCardContainer
+      render={({
+        getBatchActionProps,
+        getSelectionProps,
+        selectedSummaryCards,
+        summaryCards,
+      }) => (
+        <Fragment>
+          <SummaryCardBatchActions {...getBatchActionProps()}>
+            <SummaryCardBatchAction
```
jendowns: I realize this component is based on `TableBatchAction` -- in that context, I was wondering how someone would be able to customize the icon used here? It looks like the "trash"/delete icon is used in the mockups, but this defaults to a filled plus icon?

jendowns:
Also @SimonFinney I was wondering if we could change the title of this PR to something like:

```
feat(summary-card): add batch actions subcomponents
```

What do you think? (To me, `skeleton components` is not clear)

src/components/SummaryCard/SummaryCardContainer/SummaryCardContainer.js
SimonFinney: That makese sense!

package.json
SimonFinney: I could... I'll be writing unit tests in this branch with this package, so I'm not sure if that additional step is needed.

SimonFinney:
@jendowns Absolutely! The plan was to update to something more suitable once the PR was completed.

src/components/SummaryCard/SummaryCard.stories.js
SimonFinney: Good catch! It resorts to the default icon that Carbon sets for the `BatchAction`, but for the purposes of this story, I'll use the trash icon.

package.json
jendowns: It just seems like a big deal... to introduce this new testing library & approach. This work isn't tied to an issue, and it is slipped into this PR about something different. So tracking the introduction of it in release notes will be difficult when this is squashed + merged, plus again there's no issue/work tied to it.

SimonFinney: Thanks - I added this in https://github.com/carbon-design-system/ibm-security/pull/245 as I ended up leveraging it a number of other PRs.

I don't think this will be tracked in release notes as it's build-related, but I agree it's good practice to have the commit in the history if for some reason we need to return to the change.

src/components/SummaryCard/SummaryCard.stories.js
```diff
@@ -3,120 +3,182 @@
...
+                action('SummaryCardBatchAction onClick')(selectedSummaryCards)
+              }
+              renderIcon={Delete16}
+            >
+              Action
+            </SummaryCardBatchAction>
+          </SummaryCardBatchActions>
+
+          <div className="bx--row">
+            {summaryCards.map(({ id }) => (
```
jendowns: Unfortunately this story structure makes the source look strange & does not reflect the structure of the component at all:

Screen Shot 2019-12-05 at 1 33 29 PM

I can see someone getting confused when they see that, and they wouldn't be able to copy + paste (which I always think of as a very ideal scenario, personally)

src/components/SummaryCard/SummaryCardContainer/SummaryCardContainer.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+SummaryCardContainer.propTypes = {
+  /** Provide a list of all the summary cards to render in the container */
+  summaryCards: arrayOf(
+    shape({
+      id: string.isRequired,
+    })
+  ).isRequired,
+};
```
jendowns: Is there any way avoid this situation? This feels like a divergence from the prev structure of the component, where we were avoiding props like this?

This feels like a mixture of establishing the API and also enforcing how the user adds `SummaryCard`s to the container. Could they just add `children` in their own way?

src/components/SummaryCard/SummaryCardContainer/_mocks_/index.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @file Summary card container mocks.
+ * @copyright IBM Security 2019
+ */
+
+export default {
+  summaryCards: new Array(5)
+    .fill()
+    .map((title = 'summary-card', id) => ({ id: `${title}__${id}` })),
+};
```
jendowns: Related to my previous comment about the structure of this component - I hope we can avoid arrays like this. What do you think?

src/components/SummaryCard/_mixins.scss
```diff
@@ -4,17 +4,22 @@
...
+@import 'carbon-components/scss/globals/scss/helper-mixins';
+@import 'carbon-components/scss/globals/scss/vars';
 
-/// Use until Carbon fixes `text-overflow` helper mixin.
-/// @access private
-@mixin summary-card--ellipsis {
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
-}
```
jendowns: Hey do you mind keeping this & restoring it's usage? I don't believe the `text-overflow` helper mixin has been changed. Specifically this line: https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/globals/scss/_helper-mixins.scss#L34

SimonFinney:
@jendowns This is very odd - After addressing the truncation from your previous review, I'm unable to produce in Chrome or Firefox - Can you confirm?

Multiselect

src/components/SummaryCard/_mixins.scss
SimonFinney: I wonder if this is related to https://github.com/carbon-design-system/ibm-security/pull/228#issuecomment-562574234 and the disparity between testing environments - I'm happy to restore once we can confirm the issue is still occurring.

src/components/SummaryCard/SummaryCardContainer/SummaryCardContainer.js
SimonFinney: As far as I understand it, we'll always need an array of objects containing at least an ID (any other custom properties defined by the consumer are passed straight through), so we can track what cards have been selected and manage the logic of the batch actions, similarly to how Carbon handles this for the `DataTable`.

I was searching for examples of how consumers might currently display lists of cards, but because I couldn't find any currently in-product, the closest example we have is the `QueryCard`, which this is based off - https://github.ibm.com/investigate/investigate/blob/develop/src/components/queryCardList/QueryCardList.js#L45

My assumption is that the data for these are always going to be an array that's fetched from somewhere, and then subsequently looped through to be rendered. The renderProp should still enable teams to render the cards as they wish, the only requirement being an ID on our end for state management.

I feel like this is acceptable, but I'm open to an alternative solution if it satisfies those requirements. Would you have an example of how this might work?

jendowns: Thank you for the explanation @SimonFinney! I see what you mean now. 😅

src/components/SummaryCard/_mixins.scss
jendowns: Yeah sounds good -- I wish I more accurately remembered what he issue was that I was trying to fix. 😅 I'll look into it & post if I re-discover what the original problem was.

jendowns: Actually disregard -- I don't believe the issue is occurring anymore, and actually based on my local testing just now I think `overflow-x: hidden` is very vital.

SimonFinney:
@jendowns Thank you!

PR #229, opened 2019-11-14
feat(ui-shell): proxy carbon uishell and add simple uishell header stories

### LINKED ISSUE #29: End-Users Require Horizontal Navigation
## Request - 
The ask is for a horizontal navigation for end-user's frequently used nav items.

### Details

- **Application / team -** User Self Care
- **Work item -** Cloud Identity
- **Needed by -** September 15, 2019 -- if possible

### Proposal

The target audience for our product are end-users who require immediate access to a handful of navigational items. The end-user navigates back and forth between these select nav items several times each session. To provide the best possible user-experience, we need to present these navigational components up front rather than hidden behind the hamburger menu.

image

There's a need to differentiate between both Carbon and Security's shells in Storybook to avoid consumer confusion.

### Goals / use cases

- [ ] 
- [ ] 
- [ ]

jendowns:
## Affected issues

- Resolves #29 

## Proposed changes

- proxies all of Carbon's `UIShell` sub components
- 🚨 **IMPORTANT: renamed their `Header` as `CarbonHeader`** 🚨 
- add simple stories for `CarbonHeader` (for #29) only, to keep PR small


src/components/UIShell/index.js
```diff
@@ -0,0 +1,37 @@
...
+export SideNavFooter from 'carbon-components-react/lib/components/UIShell/SideNavFooter';
+export SideNavHeader from 'carbon-components-react/lib/components/UIShell/SideNavHeader';
+export SideNavIcon from 'carbon-components-react/lib/components/UIShell/SideNavIcon';
+export SideNavItem from 'carbon-components-react/lib/components/UIShell/SideNavItem';
+export SideNavItems from 'carbon-components-react/lib/components/UIShell/SideNavItems';
+export SideNavLink from 'carbon-components-react/lib/components/UIShell/SideNavLink';
+export SideNavLinkText from 'carbon-components-react/lib/components/UIShell/SideNavLinkText';
+export SideNavMenu from 'carbon-components-react/lib/components/UIShell/SideNavMenu';
+export SideNavMenuItem from 'carbon-components-react/lib/components/UIShell/SideNavMenuItem';
+export SideNavSwitcher from 'carbon-components-react/lib/components/UIShell/SideNavSwitcher';
```
jendowns: Compare to: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/UIShell/index.js

The only difference is here we are aliasing `Header` as `CarbonHeader` so it doesn't class with our `Header` component.

src/components/Shell/Shell.stories.js
```diff
@@ -29,7 +29,7 @@ const shellProps = {
   toolbar,
 };
 
-disableCentered(storiesOf(patterns('Shell'), module))
+disableCentered(storiesOf(patterns('Shell (Security version)'), module))
```
SimonFinney: This is minor, but what are your thoughts around just referring to this as `Shell (Security)` and the other as `UIShell (Carbon)`?

jendowns:
Thanks @SimonFinney I just applied your feedback 👍

cassiesharp:
Looks good to me @SimonFinney

Issue #178, opened 2019-10-28
ComboButton overflow menu is incorrectly placed inside a scrollable parent


i0r1:
## Bug - ComboButton overflow menu is incorrectly placed inside a scrollable parent

**Expected behavior -**
Overflow menu is placed above/below the primary button

**Actual behavior -**
Overflow menu is placed way too far from the primary button

### Steps for reproducing
[CodeSandbox](https://codesandbox.io/s/codesandbox-7z9fl)

1. Scroll down
2. Click overflow menu icon

### Screenshots

![ComboButton](https://user-images.githubusercontent.com/23272304/67700238-58ad4980-f98c-11e9-8c7e-b3e4bc6b47c5.gif)

### Affected browsers

- [ ] Chrome 77

### Optional information

**Version -** 1.6.1

PR #232, opened 2019-11-19
fix(combo-button): placement in scrollable parent

j1mie:
## Affected issues

- Resolves partially https://github.com/carbon-design-system/ibm-security/issues/178

## Proposed changes

- Tweaks placement, creating a nested group to create the layout that the menu item's width is derived from

## Testing instructions

- Add a `ComboButton` inside a div without a fixed width
- Menu should no longer occupy 100% of wrapper div.

### DIFF
--- a/src/components/ComboButton/ComboButton.js
+++ b/src/components/ComboButton/ComboButton.js
@@ -124,7 +124,7 @@ const ComboButton = ({ children, className, direction }) => {
       ref={wrapper}
       data-floating-menu-container
     >
-      <div className={`this ${namespace}__group`}>
+      <div className={`${namespace}__group`}>
         {primaryActionWithProps}
 
         {overflowMenuItemWithProps !== undefined && (
--- a/src/components/ComboButton/ComboButton.js
+++ b/src/components/ComboButton/ComboButton.js
@@ -5,7 +5,7 @@
 
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
-import React, { Fragment, useRef, useState } from 'react';
+import React, { useRef, useState } from 'react';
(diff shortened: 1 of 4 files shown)


SimonFinney:
@j1mie As per discussion this morning, should we hold out on merging this until you investigate further?

j1mie:
Hey @SimonFinney - this PR resolves the issue of the `ComboButton` in a scrollable parent. I was having a separate issue with creating a floating menu in a table toolbar, which I would consider separate to the bug referenced.

src/components/ComboButton/ComboButton.js
```diff
@@ -124,34 +124,36 @@ const ComboButton = ({ children, className, direction }) => {
...
-          onClose={() => setIsOpen(false)}
-          onOpen={() => setIsOpen(true)}
-          renderIcon={() =>
-            isOpen ? overflowIcon(ChevronUp16) : overflowIcon(ChevronDown16)
-          }
-        >
-          {overflowMenuItemWithProps}
-        </OverflowMenu>
-      )}
+      <div className={`this ${namespace}__group`}>
```
SimonFinney: - [ ] Would you mind removing the `this` class if it's not doing anything?

jendowns: @SimonFinney I disagree on this... I feel like before, we've talked about attaching classes to wrappers, even if *we* arent' necessarily styling said wrappers, in order to allow end users to more easily target those wrappers, if needed. I believe this is Carbon's approach as well.

j1mie: Oops. Typo. Sorry!

jendowns:
@j1mie could you run `yarn test -u` one more time? 😅

SimonFinney:
Thanks @j1mie !

PR #240, opened 2019-11-27
feat(time-estimator): add component

### LINKED ISSUE #145: Time estimator / formatter
### Details

Absorb 'time estimator / formatter' component as seen in some designs to determine how long an action such as a 'getting started' flow is going to take.

Time-estimator-formatter

SimonFinney:
## Affected issues

- Resolves #145 

## Proposed changes

- Add `TimeEstimator` component
- Add helpers for Storybook Component Story Format (CSF)
- Use story as the basis for snapshot test


SimonFinney:
@petervachon Good point! Right now, this is pretty dumb as it only accepts a string so technically anything could be passed in.

I could document recommended usage directly in Storybook if you had any thoughts around it?

src/components/TimeEstimator/__tests__/TimeEstimator.spec.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @file Time estimator tests.
+ * @copyright IBM Security 2019
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { Default as TimeEstimator } from '../TimeEstimator.stories';
```
jendowns: This is pretty cool, but I wonder how well it would work when a component that has a bunch of knobs?

src/components/TimeEstimator/index.js
```diff
@@ -0,0 +1,34 @@
...
+    {children}
+  </span>
+);
+
+TimeEstimator.propTypes = {
+  /** Provide the contents of the `TimeEstimator` */
+  children: node.isRequired,
+
+  /** Provide an optional class to be applied to the containing node */
+  className: string,
```
jendowns: Don't we usually have a defaults for non-required props? (Usually the linter complains right?)

src/components/TimeEstimator/index.js
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+
+import { getComponentNamespace } from '../../globals/namespace';
+
+const namespace = getComponentNamespace('time-estimator');
+
+/**
+ * Time estimators display the estimated amount of time it takes to complete a flow.
+ */
+const TimeEstimator = ({ children, className, ...props }) => (
```
jendowns: `...props` is different from our consistent use of `...other` and Carbon's consistent use of `...rest` -- could we use one of those two instead? 😅

.storybook/index.js
```diff
@@ -70,11 +70,21 @@ Also refer to http://react.carbondesignsystem.com/?path=/story/${carbon.story}--
     : description,
 });
 
+/**
+ * Configuration for applying metadata information to stories.
+ * @param {string} title The title to apply to stories.
+ * @returns {Object<string, string>} The configuration containing information to apply.
+ */
+const meta = title => ({
```
jendowns: How does `meta` usage here differ from current story implementation for other components?

src/components/TimeEstimator/index.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @file Time estimator.
```
jendowns: `TimeEstimator` is missing an entry point here in the index, and then the component defined in `.../components/TimeEstimator/TimeEstimator.js`

src/components/TimeEstimator/index.js
jendowns: I wonder about calling this `TimeEstimator` because it kinda insinuates that the component is doing more than positioning an icon next to a string? 🤔 Can we just call it something else?

OR it would even be make more generic, really. Since it's not doing anything logic-wise. It could just be text next to any ol' icon. 🤔

petervachon:
Usage documentation in Storybook would be great! It looks liked we do not want to abbreviate unless the space is too tight, which in our use case it is not, so we would use `minute`

src/components/TimeEstimator/index.js
SimonFinney: Good catch - I thought I changed this!

src/components/TimeEstimator/index.js
SimonFinney: I was hoping you'd catch this so we could discuss it!

For this one, I replicated how Carbon was handling their optional class names, but it turns out they don't provide anything - github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Accordion/Accordion.js#L27

I spoke with Josh to understand if this was an explicit decision, but I don't think it was.

We have used an empty string in the past (`''`), but I'm not sure if that's correct, as it should really be a `null` value to truly define it as a non-value.

What are your thoughts here? I'm leaning towards having a default but setting it to `null`.

src/components/TimeEstimator/index.js
SimonFinney: This is another thing I've been thinking about and questioning, and I'd like to hear your thoughts - if we only ever intend to export a simple component, do you think an entry point is necessary?

src/components/TimeEstimator/index.js
SimonFinney: Good points! 

I think the 'text beside a time icon indicating how long a flow is going to take' is a common pattern so I have concerns with making it too generic, but I agree that the name could be misleading - perhaps `TimeIndicator` is more appropriate.

Would you have any suggestions? I'm also tagging @petervachon for input.

src/components/TimeEstimator/__tests__/TimeEstimator.spec.js
SimonFinney: True - I'd be interested to see how well it integrates with event handlers, but the real value I see is that we now have synchronization between what we display in Storybook and what we unit test.

.storybook/index.js
SimonFinney: This is a helper method so we can consistently export story `meta` using Storybook's component story format - https://storybook.js.org/docs/formats/component-story-format/#default-export

jendowns: Thank you for the context! 🎉

src/components/TimeEstimator/index.js
jendowns: Oh thank you -- yeah that makes sense. 🤔 I would prefer `null` here instead of not having a default, if that's all right with you.

src/components/TimeEstimator/index.js
jendowns: Yeah it does seem like overkill in some circumstances, like this one 😅 
As far as I know, there aren't any tests associated with this structure? That would be my only hesitation.

But honestly I don't feel strongly either way! So I'm cool with this approach. 👍

src/components/TimeEstimator/index.js
jendowns: Oh yeah I like `TimeIndicator` more! 👀 
@petervachon what do you think?

src/components/TimeIndicator/index.js
```diff
@@ -0,0 +1,38 @@
...
+import { node, string } from 'prop-types';
+import React from 'react';
+
+import { getComponentNamespace } from '../../globals/namespace';
+
+const namespace = getComponentNamespace('time-indicator');
+
+/**
+ * Time indicators display an estimated amount of time it takes to complete a flow. If there is unavailable space to accommodate the text, `minute` should be abbreviated to `min`.
+ */
```
jendowns: I like this info but I wonder if it shouldn't have more visibility in the storybook demo instead? (I don't think folks expect to find this type of usage info in source code)

SimonFinney: Absolutely! See https://github.com/carbon-design-system/ibm-security/pull/240/files#diff-c736752891da25bcea8e973d2b3a5798R21

When spiking the `addon-docs` work, I discovered that the plugin will generate the documentation based on the JSDoc for the component itself. The dream is to have the usage information contained within the component definition and pulled from there but unfortunately, there are still outstanding issues with integrating `addon-docs` with our current setup (see https://github.com/storybookjs/storybook/issues/8128).

This is more future-proofing it so it's ready for that integration, but I'm happy to remove this for now if that's thinking too far ahead.

PR #243, opened 2019-11-28
fix(error-page): add unused spacing values

SimonFinney:
## Proposed changes

- Use spacing values that had been defined for `ErrorPage` but not applied
- Override browser defaults to prevent issues with resets
- Refactor to reduce redundant variables


SimonFinney:
@jendowns Good point! Done

PR #244, opened 2019-11-29
feat(trending-card): add component

### LINKED ISSUE #181: Trending Card
## Trending Card

Absorb trending card into Carbon for IBM Security.

Trending Card

SimonFinney:
## Affected issues

- Resolves #181 

## Proposed changes

- Add component, stories, and unit tests


src/components/TrendingCard/_index.scss
```diff
@@ -0,0 +1,50 @@
...
+  &:focus {
+    @include focus-outline($type: outline);
+  }
+
+  &__title {
+    @include carbon--type-style($name: body-long-01);
+
+    margin-top: 0;
+    margin-bottom: $carbon--spacing-02;
+    color: $body__color__text;
```
jendowns: I'm not familiar with this token... 🤔 Why use it instead of `$text-01`?

SimonFinney:
> Thanks @SimonFinney! I just had a few comments -- also, I similar to `TimeIndicator`, I noticed that this component is not listed in alphabetical in the storybook sidebar; it appears at the bottom of the sidebar.

See https://github.com/carbon-design-system/ibm-security/issues/265

SimonFinney:
> Looking good! One open question I have is should we add-in truncation to the subtext? I need to find a few more examples of this in practice, but since it is used primarily to represent a list of reports or other objects leaving the open string maybe a little too open. Or should we not be as restrictive? Thoughts?

@petervachon Good point! This is how the original component currently behaves, at least for the title - I can add that. It appears that the subtitle is doesn't truncate:

Storybook

src/components/TrendingCard/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+TrendingCard.defaultProps = {
+  subtitle: null,
+  element: 'a',
+  className: null,
+};
+
+TrendingCard.propTypes = {
+  /** Specify the text of the title */
+  title: string.isRequired,
```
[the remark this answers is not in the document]
SimonFinney: I'm not strongly opposed to this - It all depends on whether we want consumers to throw any old thing in there!

I followed Carbon's notifications (http://react.carbondesignsystem.com/?path=/story/notifications--toast) as an example. The `title` is restrictive but the `subtitle` allows for additional customisation, for example, if someone wanted to add a `datetime` or other custom element.

src/components/TrendingCard/index.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Trending cards provide summary information of trending items with the ability to navigate to the details.
+ */
+const TrendingCard = ({ className, element, title, subtitle, ...other }) =>
+  React.createElement(
+    element,
+    { className: classnames(namespace, className), ...other },
+    <Fragment>
+      <h3 className={`${namespace}__title`}>{title}</h3>
+      {subtitle && <span className={`${namespace}__subtitle`}>{subtitle}</span>}
```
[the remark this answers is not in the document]
SimonFinney: The differences are subtle - my understanding is that `undefined` means the prop is declared and the value is yet to be assigned, but with `null` we're saying that the default value should be explicitly empty.

src/components/TrendingCard/_index.scss
SimonFinney: This is further enforcement of our own rules, and is perhaps overkill - `$body__color__text` refers to `$text-01` under the hood, so in this case, we're saying set the color of the `TrendingCard` text to be that of whatever our body color is.

src/components/TrendingCard/index.js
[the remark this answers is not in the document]
jendowns: Yeah that makes sense -- the main thing that came to mind for me is if someone wanted to bold something, especially if they were trying to trying to create some kind of naming convention for a trending card like -- `<span><strong>Headline category: </strong> headline text</span>` or something like that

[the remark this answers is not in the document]
SimonFinney: Good point!

petervachon:
Let's add single line truncation on the sub title, for now, it's the easiest solution and should help standardize until we can push this component through different use cases

src/components/TrendingCard/_index.scss
```diff
@@ -35,9 +35,13 @@
     @include focus-outline($type: outline);
   }
 
+  &__title,
+  &__subtitle {
+    @include text-overflow;
+  }
+
```
jendowns: @SimonFinney This keeps with truncation but I think to truly demonstrate it we'd need to add a width to the component's container in the storybook demo as well. And perhaps include some note in the info/documentation that explains that to consumers.

SimonFinney: That's a good point! It definitely makes sense to add a width to demonstrate the truncation, but I'd rather hold off from including it in the notes as it seems like a higher-level design decision for text in general. Thoughts?

jendowns: If you don't add a width, and you allow users to customize title/subtitle via the knobs, then they can add text that goes off screen pretty easily, depending in their window width. And then they may think something is wrong. 

If you add a width to the container then they will be able to see that value in the story source, as separate from the component. You could go further and put it inside of grid columns, as we've done with other components.

But I think leaving it to just expand off the page like this isn't idea... they'd expect full width components to overflow properly. This one doesn't scroll/overflow properly due to the `text-overflow` mixin, so it needs to demonstrate that.

jendowns: Oh wait, disregard my previous comment -- just saw you added width but weren't sure about adding notes.

That makes sense to me. I've just been leaning more on adding notes lately to help people where something isn't 100% clear with a knob (which seems like the case here, since the width is on the container and without that width, the component doesn't behave normally with typical overflow/scroll... it just keeps going off the page)

PR #249, opened 2019-12-04
feat(unstable-pagination): refactor pagination component

### LINKED ISSUE #223: Refactor existing Pagination component with building-blocks approach
### Details

- **Application / team -** 
- **Work item -** []()
- **Needed by -** 

### Goals / use cases

- [ ] 
- [ ] 
- [ ]

jendowns:
## Affected issues

- Resolves #223 
- 🚨 NOTE: does not include `inline` variation, `SkeletonPagination` variation, or tests. Those will come in a followup.

## Proposed changes

- create `UNSTABLE_Pagination` with structure and styles
- create `PageSelector` (`Select`) subcomponent that can be passed in as a child in a render function


jendowns:
Hey @SimonFinney this is ready to review when you have a chance. The main customizable part of this at the moment is the `PageSelector` (a `Select`). And I think a future enhancement would be to add another subcomponent that is a `NumberInput`, to be used in the same way.

I did not pull out the "page sizer" `Select` or the previous/next buttons, because they seemed so integral to the component structure. 🤔 In the storybook demo variation, I show how you can leave the "page sizer" `Select` out by leaving the `pageSizes` prop as undefined, but I don't believe the prev/next buttons should be removable. Anyway, let me know what you think!

src/components/UNSTABLE_Pagination/index.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @file Unstable pagination entry point.
+ * @copyright IBM Security 2019
+ */
+
+import UNSTABLE_Pagination from './Pagination';
+import PageSelector from './PageSelector';
+
+export { PageSelector, UNSTABLE_Pagination };
```
SimonFinney: Would you mind exporting the `UNSTABLE_Pagination` as the default export and the `PageSelector` as named? I believe this aligns with other components.

src/index.js
```diff
@@ -213,3 +213,9 @@ export OrderedList from 'carbon-components-react/lib/components/OrderedList';
 
 export TabContent from 'carbon-components-react/lib/components/TabContent';
 export UnorderedList from 'carbon-components-react/lib/components/UnorderedList';
+
+// Unstable experimental components
+export {
+  PageSelector,
+  UNSTABLE_Pagination,
```
SimonFinney: Would you mind pointing to Carbon's naming convention for this? I couldn't find it on a cursory search.

src/components/UNSTABLE_Pagination/_index.scss
```diff
@@ -0,0 +1,13 @@
...
+/// @group unstable-pagination
+/// @copyright IBM Security 2019
+////
+
+@import '../Component/mixins';
+
+@import 'mixins';
+
+@include security--component($name: unstable-pagination) {
+  @include unstable-pagination;
```
SimonFinney: If you don't mind, can we remove the `mixins` file and move the style declarations directly into the `security--component` declaration? I think we'll want to move away from `mixins` if we're not using them elsewhere.

src/components/UNSTABLE_Pagination/_mixins.scss
```diff
@@ -0,0 +1,148 @@
+////
+/// Unstable pagination mixins.
+/// @group unstable-pagination
+/// @copyright IBM Security 2019
+////
+
+@import '../../globals/index';
```
SimonFinney: Instead of importing the `../../globals/index` file, can we import only the required modules? This follows Carbon's approach and will help us reduce the Sass compilation time in the future.

You can leverage the `modular-styles` branch as a reference - https://github.com/carbon-design-system/ibm-security/tree/44/modular-styles

src/components/UNSTABLE_Pagination/Pagination.stories.js
```diff
@@ -0,0 +1,29 @@
...
+
+import { UNSTABLE_Pagination, PageSelector } from '../..';
+
+import { components } from '../../../.storybook';
+
+storiesOf(components('UNSTABLE Pagination'), module)
+  .addDecorator(story => <div style={{ width: '800px' }}>{story()}</div>)
+  .add('default', () => (
+    <UNSTABLE_Pagination totalItems={350} pageSizes={[10, 15, 20, 25]}>
+      {({ currentPage, onSetPage, totalPages }) => (
```
SimonFinney: I'm not familiar with this pattern - It looks like a render prop for `children`, which makes sense, but I'm curious to hear your reasoning behind it.

src/components/UNSTABLE_Pagination/_mixins.scss
SimonFinney: Would it be worth reusing Carbon's pagination styles here where we can (if possible) to avoid duplication and keeping it in sync with the original implementation?

src/components/UNSTABLE_Pagination/PageSelector.js
```diff
@@ -0,0 +1,66 @@
...
+import { Select, SelectItem } from '../Select';
+import { appendComponentNamespace } from '../../globals/namespace/index';
+import { namespace as paginationNamespace } from './Pagination';
+
+const namespace = appendComponentNamespace(
+  paginationNamespace,
+  'page-selector'
+);
+
+function PageSelector({ className, id, labelText, totalPages, ...rest }) {
```
SimonFinney: I believe we can reduce and refactor the logic here by doing something like the following:

```js
return [...Array(total).keys()].map(counter => <SelectItem key={counter} text={String(counter)} value={counter} />)`
```

In this case, we won't need to toggle off the `no-plusplus` rule either.

src/components/UNSTABLE_Pagination/Pagination.js
```diff
@@ -0,0 +1,248 @@
+/**
```
SimonFinney: 👏

src/components/UNSTABLE_Pagination/Pagination.js
```diff
@@ -0,0 +1,248 @@
...
+  const totalPages = Math.max(Math.ceil(totalItems / currentPageSize), 1);
+
+  const backButtonDisabled = disabled || currentPage === 1;
+  const forwardButtonDisabled = disabled || currentPage === totalPages;
+
+  function onSetPage(newPage) {
+    setCurrentPage(Number(newPage));
+  }
+
+  return (
```
SimonFinney: Similar to a comment in `_mixins.scss`, is there an opportunity for us to reuse the styles from Carbon?

src/components/UNSTABLE_Pagination/_mixins.scss
jendowns: I'm happy to only import required modules! But I'd rather not import the Carbon pagination styles as-is. This component is being refactored so I think it is reasonable to also re-examine the styles.

src/components/UNSTABLE_Pagination/Pagination.stories.js
jendowns: This allows the parent component to share information (current page, etc) with children, and then inversely allows info to be shared back (setting current page, etc). While it is a more hands-on approach for consumers, it's super flexible for them.

src/components/UNSTABLE_Pagination/_index.scss
jendowns: What do you mean by "if we're not using them else"? Isn't this a pattern that most components use right now?

src/components/UNSTABLE_Pagination/PageSelector.js
jendowns: Thank you! I'll refactor these lines... but I'm hesitant about the suggestion above for a couple reasons...
- main thing to me is that it is harder to read this way... spreading the `Array.keys()` iterator object like that seems a little obscure to me? And seeing a `map` there, at first, was a little confusing because, in context of what's going on here, we are trying to use a counter to create an array, not start with an anonymous array and create a different array.
- counter starts at 0, so we'll need to put counter + 1 in every instance of counter. Doesn't really change a whole lot but it would make it all more dense/repetitive, if we wanted to keep it a single line
- smaller point: `keys={counter}` may go against the React eslint requirement that you don't use an index in a key? 🤔

jendowns:
@SimonFinney could you take another look when you get a chance?

I made most of the changes you suggested & also added some notes to storybook. I'm really curious if you think the notes are helpful and/or can be improved?

Also about the styles... I had a question about removing the `mixin` file here: https://github.com/carbon-design-system/ibm-security/pull/249#discussion_r357189254

And then generally about the styles, I totally get the desire to want to reuse what Carbon has, but I also think it would be helpful for us to evaluate them... would you be open to looking at the styles I have introduced for this refactored component & letting me know what you think?

src/components/UNSTABLE_Pagination/_index.scss
SimonFinney: For context - Where a mixin or variable is not required by another module, I'd like to push for us to have a single entry point for components to reduce cognitive overload where necessary, similarly to https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/Accordion/_index.scss or https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/StringFormatter/_index.scss.

I'm fine with this being addressed outside of this PR though.

src/components/UNSTABLE_Pagination/_mixins.scss
SimonFinney: That's fair, as long they're not deviating too much from Carbon's original styles - I'm hoping this is something that can be swiftly contributed back to reduce our need for maintaining it. 🎉

src/components/UNSTABLE_Pagination/PageSelector.js
SimonFinney: I'm not strongly opposed to this approach, as they're both essentially doing the same thing. My only concern with this is that we're switching off a linting rule to achieve something that can be achieved with an alternative solution.

I personally find traditional `for` / `while` loops to be more error-prone and would advocate avoiding these as much as possible in favor of using built-in array methods like `map`.

src/components/UNSTABLE_Pagination/PageSelector.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  return (
+    <Select
+      className={classnames(namespace, className)}
+      hideLabel
+      id={`${namespace}__input-${id}`}
+      inline
+      labelText={labelText}
+      {...rest}
```
SimonFinney: `...rest` or `..other`, where did we land on this? 😆

src/components/UNSTABLE_Pagination/Pagination.stories.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * @file Unstable pagination stories.
+ * @copyright IBM Security 2019
+ */
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import { UNSTABLE_Pagination, PageSelector } from '../..';
```
SimonFinney: Where did we land on the naming? What was your preference between `unstable__Pagination` and `UNSTABLE__Pagination`?

src/components/UNSTABLE_Pagination/_mixins.scss
```diff
@@ -0,0 +1,152 @@
...
+
+@import 'carbon-components/scss/globals/scss/vars';
+
+@import '../../globals/layout/index';
+@import '../../globals/theme/index';
+@import '../../globals/type/index';
+
+@mixin unstable-pagination {
+  @include reset;
+  @include type-style('body-short-01');
```
SimonFinney: Can we use the `carbon--` prefix here and below for consistency with everywhere else?

src/components/UNSTABLE_Pagination/_mixins.scss
```diff
@@ -0,0 +1,152 @@
...
+////
+
+@import 'carbon-components/scss/globals/scss/vars';
+
+@import '../../globals/layout/index';
+@import '../../globals/theme/index';
+@import '../../globals/type/index';
+
+@mixin unstable-pagination {
+  @include reset;
```
SimonFinney: I wish there was currently a better way of us testing these imports, but we need to include the `css--reset` partial for this mixin.

src/components/UNSTABLE_Pagination/_mixins.scss
```diff
@@ -0,0 +1,152 @@
...
+      height: 100%;
+    }
+  }
+
+  &__left {
+    padding: 0 $carbon--spacing-05 0 0;
+  }
+
+  &__left &__text {
+    margin-right: rem(1px);
```
SimonFinney: What's the reasoning for using the `rem` method versus a `$carbon--spacing-*` variable or `carbon--mini-units` helper in places?

jendowns: Unfortunately there isn't a `$carbon--spacing-*` value that equals 1px, and using the `carbon--mini-units` helper would make this simple value MUCH harder to read 😅

src/components/UNSTABLE_Pagination/_index.scss
jendowns: Thanks for the explanation -- I'm fine changing this here & removing the mixins file 👍

src/components/UNSTABLE_Pagination/PageSelector.js
jendowns: We use both 😅 `...other` is more common in our codebase but Carbon uses `...rest almost exclusively.

src/components/UNSTABLE_Pagination/Pagination.stories.js
jendowns: Ah right, forgot to mention that in my comment last week -- 

I had trouble using `unstable` (lowercase) due to component conventions starting with an uppercase letter. 🤔 I was getting errors all over the place and didn't seem worth it to try to get around.
So do you mind if we keep the uppercase `UNSTABLE` for that reason? And maybe use two underscore instead of the one?

src/components/UNSTABLE_Pagination/_mixins.scss
SimonFinney: Good catch. I didn't see the `1px`! It looks like there's other hardcoded values that can benefit from this though? For example https://github.com/carbon-design-system/ibm-security/pull/249/files#diff-da6e5a620d3051e7e0e1e711c1862f87R54

src/components/UNSTABLE_Pagination/PageSelector.js
jendowns: @SimonFinney I think this is relevant to this style PR I noticed this morning, about writing code explicitly: `https://github.com/carbon-design-system/carbon/pull/4879`

I'm not familiar with error prone nature of `for` / `while` loops? I was thinking the opposite... that using `Array.keys()` in this manner could be more error prone (for future devs looking at the code in particular).

src/components/UNSTABLE_Pagination/_mixins.scss
SimonFinney: Also, there's a mix of the old spacing helpers and the newer spacing helpers too ( https://github.com/carbon-design-system/ibm-security/pull/249/files#diff-da6e5a620d3051e7e0e1e711c1862f87R54 ) - can we use the newer values so we won't have to worry about migrating them in future?

src/components/UNSTABLE_Pagination/Pagination.stories.js
SimonFinney: Absolutely!

src/components/UNSTABLE_Pagination/_mixins.scss
jendowns: @SimonFinney Unfortunately I can't tell where you are linking to? I try to go to those links & I get sent to the top of the "files" tab. I can't see any highlighted lines. Do you mind posting block code snippets? I could search for them then, if they are just in this file.

SimonFinney: 😆 Oh, sorry! That's annoying - I could've sworn that worked. Lines 53 - 54:

```scss
&__right &__text {
    margin-right: 1rem;
```

And 124:

```scss
padding: 0 2.5rem 0 $spacing-md;
```

jendowns: Thank you! I can update those 👍

src/components/UNSTABLE_Pagination/PageSelector.js
SimonFinney: Oh! I thought we picked up the convention from them originally 😆

src/components/UNSTABLE_Pagination/_mixins.scss
jendowns: @SimonFinney Could you elaborate on what you mean here?

src/components/UNSTABLE_Pagination/PageSelector.js
jendowns: Well, scratch what I said about them using `...rest` exclusively... I searched their codebase and found it was nearly half and half 😆

src/components/UNSTABLE_Pagination/Pagination.stories.js
```diff
@@ -0,0 +1,73 @@
...
+            onChange={event => onSetPage(event.target.value)}
+            totalPages={totalPages}
+            value={currentPage}
+          />
+        )}
+      </UNSTABLE_Pagination>
+    ),
+    {
+      info: {
+        propTables: [UNSTABLE_Pagination, PageSelector],
```
jendowns: @SimonFinney I found a way to specify which proptables are shown, in situations where a component (in this case, `PageSelector` is being used as a render prop, etc, in such a way that Storybook doesn't detect its usage when generating prop tables 👀 ) -- https://www.npmjs.com/package/@storybook/addon-info#options-and-defaults

src/components/UNSTABLE_Pagination/Pagination.js
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * `true` if total number of pages is unknown.
+   */
+  pagesUnknown: PropTypes.bool,
+
+  /**
+   * The total number of items.
+   */
+  totalItems: PropTypes.number.isRequired,
```
jendowns: @SimonFinney what do you think about requiring `totalItems`? Without it, there's nothing valuable to be displayed... and it would render as `NaN` if not provided. 🤔 That change would be "breaking" but I think this would be an improvement. What do you think?

src/components/UNSTABLE_Pagination/Pagination.stories.js
SimonFinney: Very cool - nice find!

src/components/UNSTABLE_Pagination/Pagination.js
SimonFinney: That makes sense, and I wouldn't worry too much about it being a breaking change since this is experimental / in preview mode - I haven't dug too much into the original source, but is this line related? I'm curious as to how Carbon currently handle this:

 https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Pagination/Pagination.js#L299

src/components/UNSTABLE_Pagination/_mixins.scss
SimonFinney: Sure! We need to import `https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/globals/scss/_css--reset.scss` because I believe that's where the `reset` mixin lives. For example, something like:

`@import 'carbon-components/scss/globals/scss/css--reset';`

In the current codebase, we're less likely to detect where these imports fail because we import the entire styles entry point in Storybook. If a module has been imported in a previous component it will resolve in a subsequent component due to Sass compiling sequentially. This is fine if consumers import our styles entry point, but if they attempt to import components individually, they'll likely come across unfound import errors.

src/components/UNSTABLE_Pagination/PageSelector.js
SimonFinney: That's a very good point - this is a much more valuable approach for future debugging and something I hadn't entirely considered. I'm torn between us being able to optimize productivity while also leveraging newer features of the language. I'm on board!

src/components/UNSTABLE_Pagination/Pagination.js
jendowns: The `pagesUnknown` prop isn't quite related to this... if you set that to true, then it won't show the total number of pages (like if you using a backend service) but it isn't related to how total items are treated. Here are 2 examples from their current storybook:

### `totalItems={null}` and `pagesUnknown={true}`
Screen Shot 2019-12-17 at 8 54 54 AM


### `totalItems={undefined}` and `pagesUnknown={true}`
Screen Shot 2019-12-17 at 8 50 17 AM

In both cases it seems like `totalItems` is implicitly required. 🤔

jendowns: Just a thought -- if we *did* keep it so that `totalItems` isn't required, what should be shown in the component if the number isn't provided? I'm guessing we would take out the `# - # of # items` line? 🤔

SimonFinney: 🤔 I'm unsure - my gut says the logic should be handled similarly to how Carbon currently does it, right? Although, to your point maybe, this is an opportunity for improvement

jendowns: Yeah that's what my point is above... it isn't handled in Carbon at all.

jendowns:
@SimonFinney This should be ready to review again 👍

src/components/UNSTABLE_Pagination/Pagination.js
SimonFinney: Thanks - I thought those lines in the original `Pagination` handled the logic you were referring to, but by the sounds of it they don't!

src/components/UNSTABLE_Pagination/PageSelector.js
SimonFinney: Got it! I'm fine with either but we need to pick one and stick to it at this stage 😆I'm good with `other` as this is used in React's documentation - https://reactjs.org/docs/jsx-in-depth.html#spread-attributes

src/components/UNSTABLE_Pagination/Pagination.stories.js
SimonFinney: This is a quick reminder to update the component name from `UNSTABLE_*` to `UNSTABLE__*`

jendowns:
@SimonFinney I made the changes you suggested in the comments above 👍

jendowns:
@SimonFinney Is this good to merge?

SimonFinney:
@jendowns Yep, everything seems in order here! 🎉

We should document the follow-up work to ensure this is validated using a second component as discussed, for example `TextInput`

PR #258, opened 2019-12-05
 feat(panel): deprecate `Panel` in favor of `PanelV2`

### LINKED ISSUE #224: Deprecate `Panel` for `PanelV2`
## Deprecate `Panel` for `PanelV2`

SimonFinney:
BREAKING CHANGE: `Panel` has been deprecated in favor of `PanelV2`

## Affected issues

- Resolves #224 

## Proposed changes

- Rename `PanelV2` to `Panel` and remove old `Panel`
- Update `Panel` examples to exclude `primaryButton` and `secondaryButton`
- Update migration documentation


src/components/Panel/Panel.stories.js
```diff
@@ -3,145 +3,236 @@
...
+          );
+        }
+      }
+      return <PanelStory />;
+    },
+    {
+      info: {
+        text: `
+        The \`Panel\` is an experimental component that is subject to change while it is being tested. If you need a stable component, please use the \`Panel\` instead.
+      `,
```
jendowns: Could you update this info text since the component is stable / not experimental in this context?

src/components/Panel/__tests__/Panel.spec.js
```diff
@@ -3,136 +3,38 @@
...
-            <PanelContainer
-              title={label}
-              subtitle={label}
-              closeButton={{ label, onClick: toggleActive }}
+  it('renders', () => {
+    const wrapper = mount(
+      <Panel
+        title={label}
+        subtitle={label}
+        closeButton={{ onClick }}
```
jendowns:
```suggestion
        closeButton={onClick}
```

src/components/Panel/Panel.stories.js
SimonFinney: Good catch!

src/components/Panel/__tests__/Panel.spec.js
SimonFinney: I don't even know how this happened? 🤣

src/components/Panel/Panel.stories.js
SimonFinney: Good call!

src/components/Panel/__tests__/Panel.spec.js
jendowns: I think it was already in the `PanelV2` so the question should be -- how/why did I do this? 😆

PR #263, opened 2019-12-09
fix(string-formatter): update prop tooltipDirection proptype, formatting

jendowns:
## Proposed changes

* Updated`tooltipDirection` prop to match Carbon's propType & info for the `TooltipDefintion`: https://github.com/carbon-design-system/carbon/blob/dce56816e35055fe861b15d29d2f2d1eddbfeb53/packages/react/src/components/TooltipDefinition/TooltipDefinition.js#L69

```js
  /**
   * Specify the direction of the tooltip. Can be either top or bottom.
   */
  direction: PropTypes.oneOf(['top', 'bottom']),
```
* change default of `className` to `null`
* rename the constant `children` to `content` so it is not confusion (`children` being the name of an important prop that is not used in this component)

### DIFF
--- a/src/components/StringFormatter/StringFormatter.js
+++ b/src/components/StringFormatter/StringFormatter.js
@@ -4,7 +4,7 @@
  */
 
 import classnames from 'classnames';
-import { bool, string, number } from 'prop-types';
+import { bool, string, number, oneOf } from 'prop-types';
 import React from 'react';
 
 import TooltipDefinition from '../TooltipDefinition';
@@ -22,7 +22,7 @@ const StringFormatter = ({
   width,
   ...other
 }) => {
-  const children = (
+  const content = (
     <span
       className={classnames(namespace, className, {
         [`${namespace}--truncate`]: truncate,
@@ -44,10 +44,10 @@ const StringFormatter = ({
       direction={tooltipDirection}
       tooltipText={value}
     >
-      {children}
+      {content}
     </TooltipDefinition>
   ) : (
-    children
+    content
   );
 };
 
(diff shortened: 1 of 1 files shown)


src/components/StringFormatter/StringFormatter.js
```diff
@@ -58,21 +58,21 @@ StringFormatter.propTypes = {
   /** @type {number} Number of lines to clamp value. */
   lines: number,
 
-  /** @type {string} The direction of the tooltip defintion. */
-  tooltipDirection: string,
+  /** Specify the direction of the tooltip. Can be either top or bottom. */
+  tooltipDirection: oneOf(['top', 'bottom']),
```
SimonFinney: There's a reason we replicate the `propType` instead of using `TooltipDefinition.propTypes.direction`, right? Is it because Storybook doesn't output the correct value otherwise?

jendowns: I'm not sure I follow but I can look into this 🤔 
I believe I was the one who originally added this prop, when I added tooltips to the `StringFormatter`, and so the `string` vs `oneOf` thing was a mistake, I think. But let me dig into this really quick.

jendowns: Oh yeah @SimonFinney just confirmed that when you use `TooltipDefinition.propTypes.direction`, storybook does not display correctly... (it shows the `propType` for the `direction` prop to be `custom` instead of `oneOf 'top' | 'bottom'` 👍

SimonFinney: Perfect - thanks for checking!

PR #272, opened 2019-12-16
feat(string-formatter): add updated truncation library

### LINKED ISSUE #251: `StringFormatter` - Tooltip should only be displayed on truncation
#### Context

Currently, the `StringFormatter` tooltip displays regardless of whether the text is truncated or not. This should be revised to only be available when the text is truncated.

SimonFinney:
## Affected issues

- Resolves #251 

## Proposed changes

- Add `react-truncate-markup` library and adjust `HeaderNotification` styles to accommodate
- Update Storybook-generated documentation and snapshot tests


SimonFinney:
@jendowns I wanted to get your initial thoughts on this as you had worked on it previously - can you see any issues with this approach?

PR #282, opened 2019-12-20
build(scripts): add debug scss script, wrap postCSS async in a try

jendowns:
In the process of debugging this issue https://github.com/carbon-design-system/carbon/issues/4926, I noticed that the scss errors were hard to locate/figure out. So I believe the changes here can help with future styles' debugging.

## Proposed changes

- wrap async postCSS build in a `try` to get rid of node warning for unhandled promise rejection (https://stackoverflow.com/questions/40500490/what-is-an-unhandled-promise-rejection)
- add separate scss debugging file that only runs scss compilation through `node-sass`, to help us catch errors more clearly -- otherwise, we get errors filtered with postCSS & that can obfuscate the source of the problem in some cases, like when I was working on https://github.com/carbon-design-system/carbon/issues/4926

## Testing instructions

- You can run the scss debug script with `yarn debug:scss`


jendowns:
@SimonFinney I changed this to a `build` type in the heading but feel free to change if you'd rather it be a `chore` or a `fix` 🤔

jendowns:
@SimonFinney did you get a chance to look at this?

scripts/debugScss.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @file debugScss.js
+ * @description Compiles scss directly with node-sass for debugging purposes.
+ */
+
+const path = require('path');
+const sass = require('node-sass');
+
+sass.render(
```
SimonFinney: Do we need a separate debug task for this or can it be incorporated into our webpack build?

jendowns: It's not really a required step of any build process though. It's more like an optional side step, if you happened to be getting mysterious Sass compilation errors like I was but postCSS was obfuscating them. 

If we added this (even a part of it) to the webpack build, it would probably increase compile time by a lot.

SimonFinney: Ah, I thought perhaps there was an option for `sass-loader` that could provide a custom formatted / verbose compilation error similar to this so we could also catch these errors in development

jendowns: @SimonFinney That's just it -- the errors in development were not helpful 😅

jendowns:
Hey @SimonFinney I was wondering if you had any more questions / feedback on this PR?

PR #285, opened 2019-12-24
feat(shell): add technical preview

### LINKED ISSUE #173: `Shell` refactor
### `Shell` refactor

Refactor the `Shell`, `Header`, and `Toolbar` API to provide the ability for it to be more composable.

The intention is to make the props for these components easier to understand, flexible, and simpler, similar to Carbon's `UIShell` (and perhaps even leveraging it) - http://react.carbondesignsystem.com/?path=/story/ui-shell--header-base

This would allow consumers to:

- Compose only what's required
- Pass in more complex data
- Support server-side rendering

#### Follow-up

Build for production

SimonFinney:
## Affected issues

- Resolves #173 

## Proposed changes

- Abstract `Header`, `Shell`, and `Toolbar` components and subcomponents into building blocks for  composeability
- Add 'recipes' for existing popover implementations - 'Notifications' and 'Profile'
- Leverage current styles of `Header`, `Shell`, and `Toolbar` and apply modifications while maintaining both implementations
- Update semantics to account for the addition of 'slots'
- Add stories demonstrating variations
- Update `Button` and `IconButton` to use `forwardRef` to assist with focus, and account for in migration documentation
- Update `addon-centered` configuration to be compatible with stories
- Reformat `.eslintrc`
- Update Browserslist configuration to remove 'Opera Mini'

## Outstanding

- Technical preview with stakeholders
- a11y and unit tests
- Fully deprecate previous implementation and styles


SimonFinney:
## Feedback

- Active logic can be built into the implementation to hide from the consumer
- Basic alignment and positioning slots - Middle and right, for example
- The `ShellHeaderActions` API is too verbose and should not be too restrictive to allow for other use cases
- Provide 'recipes' for more complex popovers

src/components/SecurityShell/SecurityShell.js
```diff
@@ -0,0 +1,264 @@
...
+
+import Button from '../Button';
+import HeaderNotification from '../SecurityHeader/HeaderNotification';
+import Icon from '../Icon';
+import IconButton from '../IconButton';
+import Link from '../Link';
+import ScrollGradient from '../ScrollGradient';
+
+const namespace = getComponentNamespace('header');
+const shellNamespace = getComponentNamespace('shell');
```
jendowns: IMO this could be called `legacyShellNamespace` to be crystal clear to anyone looking at this file in the future.

src/components/SecurityShell/SecurityShell.js
```diff
@@ -0,0 +1,264 @@
...
+
+const SecurityShellHeader = ({ children, ...other }) => (
+  <div className={`${namespace}__container ${securityShellNamespace}__header`}>
+    <header className={namespace} role="banner" {...other}>
+      {children}
+    </header>
+  </div>
+);
+
+const SecurityShellHeaderAction = ({
```
jendowns: Could you update these to function declarations instead? Thinking about Carbon's style direction with function declarations & that we should try to match that.

src/components/SecurityShell/SecurityShell.js
```diff
@@ -0,0 +1,264 @@
...
+
+      ref.current.focus();
+    }
+  };
+
+  return (
+    <>
+      {cloneElement(children, {
+        onClick,
+        onFocus: () => console.log('onFocus'),
```
jendowns: temporary?

src/components/SecurityShell/SecurityShell.js
```diff
@@ -0,0 +1,264 @@
...
+);
+
+const SecurityShellHeaderPopoverNotification = ({ ...other }) => (
+  <li className={`${namespace}__popover__list-item`}>
+    <HeaderNotification {...other} />
+  </li>
+);
+
+const SecurityShellHeaderPopoverNotifications = ({
+  ariaLabel,
```
jendowns: Since this component is more than the close button that will have this aria-label, I wonder if shouldn't call it `closeButtonAriaLabel` to be clearer.

src/components/SecurityShell/SecurityShell.js
```diff
@@ -0,0 +1,264 @@
...
+const SecurityShellSkipToContent = ({ ...other }) => (
+  <div className={`${shellNamespace}__skip-to-content`}>
+    <Button
+      className={`${shellNamespace}__skip-to-content__link`}
+      // eslint-disable-next-line jsx-a11y/tabindex-no-positive
+      tabIndex={1}
+      {...other}
+    />
+  </div>
+);
```
jendowns: For a skip-to-content link the target `href` and the label text is going to be very important, so they shouldn't be relegated to a spread attribute.

SimonFinney:
@jendowns Thanks for the review - As mentioned, this will be moved from a proof-of-concept into production this sprint so I'll definitely take all your feedback on board!

If you have time to look at the story file itself, and the API and ergonomics of that, that's where your feedback would be most valuable to me right now - Thank you!

jendowns:
Hey @simonfinney I’m not sure what you are referencing for “as mentioned” but with regards to the story, I didn’t have anything to add at this time.

SimonFinney:
@sbutler2901 For your consideration if you have some spare time

src/components/UNSTABLE__Shell/HeaderAction/index.js
```diff
@@ -0,0 +1,101 @@
...
+          [`${headerButtonNamespace}--notifications`]: isIconButton && hasBadge,
+          [`${headerButtonNamespace}--active`]: isIconButton && isActive,
+        }),
+        onClick: () => {
+          setActiveAction(!isActive ? id : null);
+
+          if (!isActive) {
+            ref.current.focus();
+          }
+        },
```
sbutler2901: `onBlur` will have to be added here to handle when the action loses focus. 
```
        onBlur: () => isActive && setActiveAction(null),
```
This worked for me when running the storybook.

SimonFinney: Good catch!

sbutler2901: @SimonFinney that is right in line with what I was thinking! But when I run it locally and select and account's dropdown it still removes the popover:

Screen Shot 2020-02-18 at 12 09 03 PM

sbutler2901:
@SimonFinney I'm doing some testing of your most recent changes. It seems to work fine (when clicking) for Chrome, but not Firefox or Safari. When tabbing to the account list section and selecting / deslecting with space all of the browsers seem to work.

I do notice that there are some cases when `current` is undefined when calling `current.focus();` which causes an error to be logged to the console.

On initial page load I can select the user profile button to cause the popover to appear. This seems to be how the browsers differ:

Chrome:
- It correctly puts focus on the profile button just the same as if I had tabbed to it and selected it. Event though the previously mentioned `current === undefined` error is logged.
Screen Shot 2020-02-25 at 2 30 10 PM

Firefox & Safari:
- The focus is not set to the popover button after selecting it and remains on `<body class="sb-show-main">…</body>`

Screen Shot 2020-02-25 at 2 35 03 PM

- If the user then clicks anywhere else on the page the popover will not disappear
- If the profile button is clicked to close the popover and then clicked again to open, the focus will be given to the profile button correctly. The popover will then close if the user clicks anywhere on the page, but if the account list accordion is selected the popover will be automatically closed.

ff-safari-issue

package.json
```diff
@@ -164,7 +164,8 @@
   },
   "browserslist": [
     "> 1%",
-    "not IE 11"
+    "not IE 11",
+    "not OperaMini all"
```
jendowns: 👏

src/components/Button/Button.js
```diff
@@ -16,35 +16,32 @@ export const namespace = getComponentNamespace('button');
...
-      className={buttonClasses}
-      disabled={disabled || loading}
-      kind={isGhostDanger || loading ? 'ghost' : kind}
-      renderIcon={loading ? InlineLoading : renderIcon}
-      size={!isLarge ? size : null}
-      {...other}
-    />
-  );
-};
+const Button = forwardRef(
```
jendowns: 👏

.storybook/index.js
```diff
@@ -44,17 +44,23 @@ const getComponentsCategory = bindCategory(COMPONENTS);
...
  * @returns {Object<string, Object>} The collection of stories.
  */
-const disableCentered = stories =>
-  stories.addParameters({
-    centered: {
-      disable: true,
-    },
-  });
+const disableCenteredStories = stories =>
+  stories.addParameters(disableCentered());
```
jendowns: Is there any way to leave out this change to reduce the subset of changes + files edited in this PR? 😅 
It is producing quite a bit of noise so I was wondering if it is necessary here?

src/components/SecurityHeader/HeaderNotification/constants.js
```diff
@@ -26,9 +26,6 @@ const propTypes = {
   /** @type {string} Product name. */
   product: PropTypes.string.isRequired,
 
-  /** @type {string} Time label. */
-  timeLabel: PropTypes.string.isRequired,
-
```
jendowns: Ahhhh is this prop no longer used? 👀

src/components/SecurityHeader/SecurityHeader.js
```diff
@@ -327,7 +327,7 @@ export default class SecurityHeader extends Component {
           </div>
         )}
         <ScrollGradient color={theme.inverse02}>
-          <ul className={`${namespace}__popover__list`}>
+          <div className={`${namespace}__popover__list`}>
```
jendowns: We really should try to keep semantic hierarchy. ☝️ 
(Though worth noting that only `ul`/`li` elements can be children of a `ul`)

src/components/UNSTABLE__Shell/HeaderAction/index.js
```diff
@@ -0,0 +1,117 @@
...
+      isActive &&
+      !popoverRef.current.contains(relatedTarget) &&
+      relatedTarget !== ref.current &&
+      setActiveAction(null)
+    );
+  }
+
+  return (
+    <>
+      {cloneElement(children, {
```
jendowns: Wondering if `cloneElement` is necessary here?

src/components/UNSTABLE__Shell/HeaderActions/index.js
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <div
+      className={`${appendComponentNamespace(
+        headerNamespace,
+        suffix
+      )} ${namespace}`}
+      {...other}
+    >
+      {Children.map(children, (child, index) =>
+        cloneElement(child, {
```
jendowns: Could we avoid `cloneElement` for easier debugging?

src/components/UNSTABLE__Shell/NotificationsPopover/index.js
```diff
@@ -0,0 +1,76 @@
...
+      <PopoverContent {...other}>
+        {label && <span className={`${namespace}__label`}>{label}</span>}
+        {onClear && (
+          <button
+            className={`${namespace}__button`}
+            aria-label={iconDescription}
+            onClick={onClear}
+          >
+            <Icon renderIcon={Close20} />
+          </button>
```
jendowns: Were you able to use a Carbon `Button` element with `renderIcon` instead?

src/components/UNSTABLE__Shell/PopoverAccount/index.js
```diff
@@ -0,0 +1,50 @@
...
+export const namespace = appendComponentNamespace(profileNamespace, 'body');
+
+function PopoverAccount({ account, children, className, id, ...other }) {
+  return (
+    <section
+      className={classnames(namespace, `${namespace}--account`, className)}
+      {...other}
+    >
+      <div className={`${namespace}__label`}>
+        {account}: <span className="selectable-text">{id}</span>
```
jendowns: Why doesn't the `selectable-text` class name have a namespace? (Should it?)

src/components/UNSTABLE__Shell/PopoverAccountItem/index.js
```diff
@@ -0,0 +1,45 @@
...
+const namespace = appendComponentNamespace(accountNamespace, 'account');
+
+function PopoverAccountItem({ children, id, onClick, ...other }) {
+  return (
+    <li className={namespace}>
+      <button
+        className={`${namespace}__button`}
+        onClick={onClick}
+        value={id}
+        {...other}
```
jendowns: Did you try to use a Carbon `Button` here?
(The reason I mention that here + in another popover subcomponent is for the built-in a11y attributes, mainly. They are very useful & I would rather take advantage of those + add style overrides than use a plain `button` element to avoid style overrides -- does that make sense? And do you have concerns?)

src/components/UNSTABLE__Shell/PopoverAccountItem/index.js
```diff
@@ -0,0 +1,45 @@
...
+import { func, node, string } from 'prop-types';
+import React from 'react';
+
+import { appendComponentNamespace } from '../../../globals/namespace';
+
+import { namespace as accountNamespace } from '../PopoverAccount';
+
+const namespace = appendComponentNamespace(accountNamespace, 'account');
+
+function PopoverAccountItem({ children, id, onClick, ...other }) {
```
jendowns: For `id` can we utilize Carbon's instance id generation tool? (used in `PanelV2` here: https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/PanelV2/PanelV2.js#L28)

src/components/UNSTABLE__Shell/PopoverFooter/index.js
```diff
@@ -0,0 +1,28 @@
...
+import { node } from 'prop-types';
+import React from 'react';
+
+import { appendComponentNamespace } from '../../../globals/namespace';
+import namespace from '../Popover';
+
+function PopoverFooter({ children, ...other }) {
+  return (
+    <footer
+      className={appendComponentNamespace(namespace, 'footer')}
```
jendowns:
```suggestion

const footerNamespace = appendComponentNamespace(namespace, 'footer');

function PopoverFooter({ children, ...other }) {
  return (
    <footer
      className={footerNamespace}
```

Could we do this? Likely may need to export the namespace for tests 🤔

src/components/UNSTABLE__Shell/PopoverHeader/index.js
```diff
@@ -0,0 +1,28 @@
...
+import { node } from 'prop-types';
+import React from 'react';
+
+import { appendComponentNamespace } from '../../../globals/namespace';
+import popoverNamespace from '../Popover';
+
+function PopoverHeader({ children, ...other }) {
+  return (
+    <section
+      className={appendComponentNamespace(popoverNamespace, 'header')}
```
jendowns: Similar suggestion as in `PopoverFooter` -- can we set this up where the namespace in available to exported for tests?

src/components/UNSTABLE__Shell/ProfilePopover/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+import { appendComponentNamespace } from '../../../globals/namespace';
+
+import ProfileImage from '../../ProfileImage';
+import PopoverHeader from '../PopoverHeader';
+
+import popoverNamespace from '../Popover';
+
+export const namespace = appendComponentNamespace(popoverNamespace, 'profile');
+const headerNamespace = appendComponentNamespace(namespace, 'header');
```
jendowns: Could this namespace be imported from `PopoverHeader`?

src/components/UNSTABLE__Shell/Shell.stories.js
```diff
@@ -0,0 +1,299 @@
...
+
+        <HeaderAction>
+          <Button kind="secondary">Sign in</Button>
+        </HeaderAction>
+      </HeaderActions>
+    </ShellHeader>
+  </UNSTABLE__Shell>
+);
+
+export const skipToContent = () => (
```
jendowns: For skip to content stories in particular, it is helpful to include content to skip to (i.e., not just a "main" target but also something inside, to demonstrate to users how tab order jumps to the skipped-to area)

src/components/UNSTABLE__Shell/ShellHeaderName/index.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * @file Header 'name'.
```
jendowns:
```suggestion
 * @file Header name.
```

src/components/UNSTABLE__Shell/ShellReturnToBanner/index.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * @file 'Return' navigation.
```
jendowns: Can this reflect the component name?

src/components/UNSTABLE__Shell/ShellReturnToBanner/index.js
```diff
@@ -0,0 +1,68 @@
...
+      )}`}
+      href={href}
+      style={{
+        backgroundImage: `url(${dataUri(
+          '../../../images/aurora-banner@2x.png'
+        )})`,
+      }}
+      {...other}
+    >
+      <Icon className={`${legacyNamespace}__icon`} renderIcon={ArrowLeft16} />
```
jendowns: Just an idea -- instead of `Link` with an `Icon` inside (likely involves adding placement styles), you can use --

```jsx
<Button
  href={href}
  renderIcon={ArrowLeft16}
  {...etc}
/>
```

The `href` prop turns it into a semantic `a`/anchor element & then the icon placement is handled for you.

src/components/UNSTABLE__Shell/ShellSkipToContent/index.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * @file 'Skip' navigation.
```
jendowns: Skip to content link

src/components/UNSTABLE__Shell/ShellToolbar/index.js
```diff
@@ -0,0 +1,42 @@
...
+const namespace = appendComponentNamespace(shellNamespace, 'toolbar');
+
+function ShellToolbar({ children, ...other }) {
+  const [activeAction, setActiveAction] = useState(null);
+
+  return (
+    <nav className={`${toolbarNamespace} ${namespace}`} {...other}>
+      <div className={`${toolbarNamespace}__group`}>
+        {Children.map(children, (child, index) =>
+          cloneElement(child, {
```
jendowns: Can we avoid `cloneElement` here?

src/components/SecurityHeader/HeaderNotification/constants.js
SimonFinney: Yeah! It didn't look like it was ever used in `HeaderNotification` 😆

.storybook/index.js
SimonFinney: I could add it in a separate PR if you'd prefer, but at the time it was required for this work to disable `@storybook/addon-centered` for both the old story format and CSF

src/components/UNSTABLE__Shell/PopoverAccountItem/index.js
SimonFinney: This needs to be a consumer-supplied value as it signifies a user account ID, similarly to the previous implementation

src/components/UNSTABLE__Shell/PopoverAccount/index.js
SimonFinney: It should! Because we're leveraging the current implementation's styles, I wanted to avoid touching anything that may cause a noticeable breaking change - I'll need to create a task for fully consolidating the styles once we're ready to deprecate the old `Header`, `Shell`, and `Toolbar`

src/components/UNSTABLE__Shell/ShellReturnToBanner/index.js
SimonFinney: Good thought! I originally tried this and it didn't play so nice out-of-the-box - 

Button

To remain close to the original implementation and to position the icon on the left as per designs, the `Link` seemed more appropriate:

Link

src/components/UNSTABLE__Shell/NotificationsPopover/index.js
SimonFinney: Good point! I'll use the `IconButton` here

src/components/UNSTABLE__Shell/Shell.stories.js
SimonFinney: Good point! Done

src/components/UNSTABLE__Shell/PopoverFooter/index.js
SimonFinney: Here, I wanted to keep everything as private as possible until we find the necessity to expose it as an export.

I can imagine we may eventually need these for unit tests, but I'd rather do that once we identify that we do, if that makes sense. Otherwise, we run the risk of exposing it to consumers (I'm not sure how real a concern this is), but if so, it becomes harder to remove down the line.

src/components/UNSTABLE__Shell/PopoverHeader/index.js
SimonFinney: Resolving thread in favor of https://github.com/carbon-design-system/ibm-security/pull/285#discussion_r384741145

src/components/UNSTABLE__Shell/HeaderAction/index.js
SimonFinney: Because the consumer can provide a number of things here, and we need to apply particular classes and handlers to manage the state between actions, I don't think there's any way of avoiding this. What's the alternative approach?

src/components/UNSTABLE__Shell/PopoverAccountItem/index.js
SimonFinney: To your point, though, there are places like `HeaderActions` and `ShellToolbar` that could benefit from this!

src/components/SecurityHeader/SecurityHeader.js
SimonFinney: Yeah, I agree, and I wasn't proud of this! 😆Because we want this area to be configurable for consumers to provide notifications or content in the absence of notifications, it no longer made sense to restrict it to a `ul`

src/components/UNSTABLE__Shell/ProfilePopover/index.js
SimonFinney: Good catch!

src/components/UNSTABLE__Shell/HeaderActions/index.js
SimonFinney: Resolving thread in favor of https://github.com/carbon-design-system/ibm-security/pull/285#discussion_r384737204

src/components/UNSTABLE__Shell/ShellToolbar/index.js
SimonFinney: Resolving thread in favor of https://github.com/carbon-design-system/ibm-security/pull/285#discussion_r384737204

src/components/UNSTABLE__Shell/PopoverAccountItem/index.js
SimonFinney: Resolving thread as per Slack conversation

SimonFinney:
@sbutler2901 I've made some changes that I believe resolve the issues outlined in your comment above - let me know your thoughts!

SimonFinney:
@jendowns That's an interesting point on naming!

I've kept it closely aligned with the original, as it's currently consistent with the Cloud Pak for Security `shell` repo, but perhaps we can crowd-source the discussion and any suggestions - @dbrugger , @petervachon , @sbutler2901 

While this is happening, I've addressed some of your feedback and provided some discussion points for outstanding questions. I know you mentioned you'd like to do another pass - I'd like to ensure we get final changes in this week, so we can get this into consumers' hands by the end of sprint.

SimonFinney:
Thanks @jendowns ! Would you mind confirming which of these issues exist in the current `Shell` implementation, and which of these have been introduced with the refactor? That will help push this along

SimonFinney:
☝️ My Chrome extension is down at the moment, so this is my best guess at consolidating these - let me know if and where you disagree!

The next step here would be to create follow-up defects so these can be resolved across the board.

jendowns:
@SimonFinney As I posted in Slack, I think it's very important to address top-level a11y issues in this refactor. Please let me know if you could like some help with any of these! 👍

PR #289, opened 2020-01-02
fix(theme-docs): filter for color tokens when generating themes.md

### LINKED ISSUE #288: Themes docs: table of generated tokens includes more than color tokens.
## Bug - 

**Expected behavior -**

Expect the generated table at https://github.com/carbon-design-system/ibm-security/blob/dev/docs/themes/themes.md to only show color tokens.

**Actual behavior -**

Generated table at https://github.com/carbon-design-system/ibm-security/blob/dev/docs/themes/themes.md includes type tokens. Moreover, when working on #281, the themes file automatically added spacing and layout tokens as well. It appears to be grabbing all tokens, not just color-related tokens.

### Steps for reproducing

* See current `themes.md` file for type variables (scroll to bottom): https://github.com/carbon-design-system/ibm-security/blob/dev/docs/themes/themes.md
* Spacing and layout tokens getting added here: https://github.com/carbon-design-system/ibm-security/pull/287/commits/13f3a68acea49029813f87f72862f720e863b033 (scroll down to automatic changes applied to `themes.md` file)

jendowns:
## Affected issues

- Resolves #288 

## Proposed changes

- filter for color tokens when generating `themes.md` file (since Carbon now includes all tokens, including type and spacing, into theme maps)
- update `themes.md` file


jendowns:
Hey @SimonFinney would you have a chance to review this please? (I keep having to add caveats when I share this file to people -- i.e., "please ignore the non color values")

SimonFinney:
Thanks for your patience - I'll add some suggestions now!

scripts/themes/index.js
```diff
@@ -21,16 +21,21 @@ const { value: themeMap } = renderSync({
...
 // Loops through each of the supported themes, finding and formatting the correct color hex values and names.
 themes.forEach(theme => {
   const { value: tokenObject } = themeMap[theme];
 
   Object.keys(tokenObject)
+    // TODO - currently only handling color tokens.
+    .filter(tokenValue => tokenObject[tokenValue].type === 'SassColor')
     .map(tokenValue => tokenObject[tokenValue])
     .forEach(({ value }, index) => {
       let { hex: color = 'colors' } = value;
```
SimonFinney: - [ ] We can remove this default assignment because we're only outputting color values

scripts/themes/index.js
```diff
@@ -21,16 +21,21 @@ const { value: themeMap } = renderSync({
 const themes = Object.keys(themeMap);
 
 // Generates the data for themes and tokens to template with Handlebars.
-const tokens = Object.keys(themeMap[themes[0]].value).map(token => ({
-  themes: [],
-  token,
-}));
+const tokens = Object.keys(themeMap[themes[0]].value)
+  // TODO - currently only handling color tokens.
+  .filter(token => themeMap[themes[0]].value[token].type === 'SassColor')
```
SimonFinney: I'd like to pull this out into a helper method above for reuse here and on line 38 - For example:

`const filterColor = ({ type }) => type === 'SassColor';`

And its usage - `.filter(token => filterColor(themeMap[themes[0]].value[token]))`

scripts/themes/index.js
```diff
@@ -21,16 +21,21 @@ const { value: themeMap } = renderSync({
 const themes = Object.keys(themeMap);
 
 // Generates the data for themes and tokens to template with Handlebars.
-const tokens = Object.keys(themeMap[themes[0]].value).map(token => ({
-  themes: [],
-  token,
-}));
+const tokens = Object.keys(themeMap[themes[0]].value)
+  // TODO - currently only handling color tokens.
```
SimonFinney: I don't initially understand what this comment and the comment on line 37 is describing - Is there another task to be completed here?

jendowns: From this comment (and possibly related Slack comments earlier this month): https://github.com/carbon-design-system/ibm-security/issues/288#issuecomment-570322154

> It may just be best to ignore any non-color-related variable for now and only show color variables in themes.md

That is, we may want to include these non-color tokens from the theme map in the future. But for now, the easiest thing is to filter them out.

scripts/themes/index.js
jendowns: You don't want to keep it just in case?

scripts/themes/index.js
jendowns: I'm having trouble seeing the value of abstraction here. `type === 'SassColor'` in the location it is used (to avoid looking elsewhere, to see what is actually being filtered for) is not preferable to having a method called `filterColor`?

scripts/themes/index.js
SimonFinney: The purpose of the value was to replicate Carbon's themes (https://carbon-elements.netlify.com/themes/examples/preview/) - if we're filtering them out, I don't think we need this anymore

scripts/themes/index.js
SimonFinney: I think alternatively to reuse, it might be more appropriate to find an alternative solution where the colors are filtered before they get to these points to avoid duplication

scripts/themes/index.js
SimonFinney: Thanks for clarifying - I recommend removing `TODO` from the comment here to avoid confusion, and being more explicit about the code. For example - 

```suggestion
  // Filter only color tokens from the theme.
```

jendowns: Wait so you want to remove ‘TODO’ which we use elsewhere to help us find things that need to be revisited, and only describe what the line does in the comment? 😅 (Also consider above you asked for a helper ‘filterColor’ so that seems redundant?

scripts/themes/index.js
jendowns: Should I add a ‘TODO’ as well for additional work to be done? 😅

jendowns:
@SimonFinney Hopefully this PR is good to go now? I kept `TODO`s because we use those throughout the codebase to indicate somethings to be revisited.

PR #291, opened 2020-01-02
fix(ibm-type): remove ibm-type feature flag & helpers

### LINKED ISSUE #255: `2.x` - Remove `ibm-type`
#### Context

`ibm-type` was originally included for backward compatibility, and should be removed in this major release.

jendowns:
## Affected issues

- Resolves #255 

## Proposed changes

- remove `ibm-type` feature flag, `$type__name` + helpers, and related outdated documentation (since it is no longer supported in Carbon anyway, so as a migration note it wouldn't be useful now)
- update feature flag demo in docs


src/globals/type/_index.scss
```diff
@@ -72,42 +72,3 @@ $type-style-deprecation-map: (
   @warn deprecate($actual: 'type-scale-item($scale-item: #{$scale-item})', $expected: 'carbon--type-style($name: #{$name})');
   @include carbon--type-style($name: $name);
 }
-
```
SimonFinney: Could you also remove the `type-scale-item` helper method along with the deprecation maps and imports above?

jendowns:
thanks @SimonFinney -- this is ready to review again 👍

src/globals/type/_index.scss
```diff
@@ -4,110 +4,11 @@
...
-    }
-
-    @each $type-class, $type-style-name in $type-style-deprecation-map {
-      .#{$type__name}-#{$type-class} {
-        @include carbon--type-style($name: $type-style-name);
-      }
-    }
-  }
-}
+@include carbon--type-reset();
```
SimonFinney: What does this line do? Also, will the font face here be output multiple times if this file is imported more than once?

README.md
```diff
@@ -86,7 +86,7 @@ Carbon for IBM Security takes advantage of feature flags to conditionally enable
 
 ```scss
 $security--feature-flags: (
-  ibm-type: false,
+  grid-columns-16: true,
```
SimonFinney: 🎉

src/globals/type/_index.scss
jendowns: From the usage docs: https://github.com/carbon-design-system/carbon/tree/master/packages/type#usage

jendowns:
@SimonFinney Were there any changes you wanted to see with this PR?

SimonFinney:
@jendowns Yes, I believe the `font-face` and `reset` mixins will be imported and called multiple times with this current setup as there's no `import-once` guard in place. This will in turn cause bloat in the CSS output  -

- https://github.com/carbon-design-system/carbon/blob/master/packages/type/scss/font-face/_mono.scss
- https://github.com/carbon-design-system/carbon/blob/master/packages/type/scss/_reset.scss

jendowns:
@SimonFinney How should we handle this then? 🤔

SimonFinney:
We currently have some of these in https://github.com/carbon-design-system/ibm-security/blob/dev/src/platform/body/_mixins.scss so perhaps adding the `font-face` there could work!

Otherwise, it might make sense for all of those to live as a guard in `type/index` instead - what are your thoughts?

jendowns:
Thanks @SimonFinney! I appreciate the context... and I liked that first suggestion so I went with that. Changes are in now. Let me know what you think 👍

PR #308, opened 2020-01-13
fix(panel): add a11y props and attributes

### LINKED ISSUE #278: `2.x` - `PanelV2` should have required accessibility props for header
`PanelV2` should have the same required accessibility props as the `Modal`.

There needs to be a required `aria-label` prop applied to the header of the panel.
And then there needs to be an `id` prop (not required, with a valid default value) that is applied to the header of the panel.
Finally, the `children` of `PanelV2` should be wrapped in `role="region"` and have an `aria-labelledBy` value that references the `id` added to the panel's header.

A related change: the `PanelContent` component should not have `role="tabpanel"`. The recommended change to apply `role="region"` to wrap the `children` of the `PanelV2` will replace the role that is currently applied to `PanelContent`.

jendowns:
## Affected issues

- Resolves #278 

(NOTE: I went ahead and made changes off of `dev` instead of just off `2.x)

## Proposed changes

- use same a11y props and attributes that Carbon uses in the `Modal` component: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js 

## Testing instructions

- Run the storybook deployment in Chrome, view `PanelV2` and `Panel` in isolation, and then run the DAP tool to ensure that there are no violations (I suggested running when closed & then rerunning when open)


src/components/Panel/PanelContainer.js
```diff
@@ -3,13 +3,18 @@
  * @copyright IBM Security 2019
  */
 
+/* eslint-disable no-useless-computed-key, react/require-default-props */
```
SimonFinney: What are these rules disabling?

src/components/Panel/PanelContainer.js
```diff
@@ -3,13 +3,18 @@
...
+
 import Close20 from '@carbon/icons-react/lib/close/20';
 
 import classnames from 'classnames';
 import PropTypes, { func } from 'prop-types';
 import React, { Component, createRef, Fragment } from 'react';
 import { createPortal } from 'react-dom';
 
+import requiredIfGivenPropExists from 'carbon-components-react/es/prop-types/requiredIfGivenPropExists';
```
SimonFinney: Very cool! 👏

SimonFinney: More often than not, we use Carbon's `lib` output here over `es`

src/components/Panel/PanelContainer.js
```diff
@@ -142,18 +152,50 @@ export default class PanelContainer extends Component {
...
+      rootNode,
+
+      // Required because it may include `aria-label`, etc.
+      ...other
     } = this.props;
 
     const hasFooter = renderFooter || primaryButton;
 
+    const ariaLabel =
+      title || this.props['aria-label'] || panelAriaLabel || subtitle;
```
SimonFinney: I'm curious to hear your thoughts about the differences between the `aria-label` and `panelAriaLabel` props? It initially feels like we should go with one or the other.

Also, perhaps the ordering here should be more like the following, so if a consumer supplies an `aria-label` that takes ultimate preference over everything else:

```js
this.props['aria-label'] || title || subtitle;
```

src/components/Panel/PanelContent.js
```diff
@@ -18,11 +18,7 @@ const namespace = getComponentNamespace('panel');
  */
 const PanelContent = props => {
   const classes = classnames(`${namespace}__body--content`, props.className);
-  return (
-    <div tabIndex="0" role="tabpanel" className={classes}>
-      {props.children}
-    </div>
-  );
+  return <div className={classes}>{props.children}</div>;
```
SimonFinney: 👏

src/components/Panel/PanelContainer.js
```diff
@@ -142,18 +152,50 @@ export default class PanelContainer extends Component {
...
+
+    const getAriaLabelledBy = title ? this.paneltitleId : this.panelSubtitleId;
+
+    const hasScrollingContentProps = hasScrollingContent
```
SimonFinney: I'm unsure about this, purely because we may not be able to accurately determine the length of the content being inserted into the panel. Is this to prevent the panel from being focusable if the content is not overflowing?

As a very basic example, if you reduce the height of the viewport, the panel will become scrollable after initialization.

src/components/Panel/PanelContainer.js
jendowns: The proptypes block --

```jsx
  ['aria-label']: requiredIfGivenPropExists(
    'hasScrollingContent',
    PropTypes.string
  ),
```

-- contains both a computed property and a prop without a default

src/components/Panel/PanelContainer.js
jendowns: I strongly thing we should keep the same logic as Carbon's `Modal`, to create a similar expectation with how the `Panel` is meant to be used. Their accessibility team made this decision here -- https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L362

```
    const ariaLabel =
      modalLabel || this.props['aria-label'] || modalAriaLabel || modalHeading;
```

(where `modalLabel` is virtually the same as the `title` here)

jendowns: As for the difference between `aria-label` and `panelAriaLabel` -- my understanding is that Carbon was trying to catch a wide net with the `Modal`'s a11y updates, providing as much flexibility as possible for folks. `aria-label` is a valid attribute (that would be used here) but it's possible that folks would prefer to use React props.

src/components/Panel/PanelContainer.js
jendowns: There's no way to 100% guarantee, like with the `ScrollGradient` -- you have to trust the consumer to know if they plan to add a lot of content or not. Otherwise, adding attributes that may ultimately be unnecessary, in content, (by just adding all of this attributes by default, for example) would likely be disruptive.

Here's where Carbon added the same logic: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Modal/Modal.js#L367

As with the aria-label logic, I think it makes sense to create the same expectations in the `Panel` that Carbon has created for the `Modal`.

src/components/Panel/PanelContainer.js
jendowns: @SimonFinney Actually disregard my last comment about `panelAriaLabel` -- I added it because I saw it used in the `Modal`, but it appears based on the git history that their equivalent, `modalAriaLabel`, is 2 years old so they are keeping it around for backwards compatibility: https://github.com/carbon-design-system/carbon/blame/4a0c095f6086c78f9bc4d2b1d7a92d89cb79d990/packages/react/src/components/Modal/Modal.js#L69

Likewise I'll go ahead and remove `panelAriaLabel` 👍

jendowns:
Thank you @SimonFinney! I just made a couple changes you suggested above & also responded to your questions -- let me know what you think.

src/components/Panel/PanelContainer.js
SimonFinney: Thanks for clarifying! I trust Carbon's decision considering they have a dedicated a11y resource, but I worry that if an `aria-label` is supplied intentionally then it's always going to be superseded by the `title`, which may not be the intention.

For example, the `aria-label` could be declared something like  'IP risk panel', while the `title` could be an IP address like '127.0.0.1', which I feel doesn't provide as much semantic context.

But... perhaps I'm over-thinking this! 😆I'd love to hear your thoughts!

src/components/Panel/PanelContainer.js
SimonFinney: For additional context, here's an example of the same `Panel` with the same data model, but 2 different results - 

### Without scrolling

1-2_Panel

### With scrolling

2-2_Panel

In this case, I hypothesize that consumers might need to hack together logic to get the height of the content in the panel relative to the rest of it after fetching it, and also account for resizing to determine whether `hasScrollingContent` needs to be applied.

I'm not strongly opposed to this idea, I'm just concerned that this may be a recurring issue for consumers adopting this change and that they might have to implement a solution similar to the above.

I'm wondering if there's a suitable person that works with the panel that we can bounce this off?

src/components/Panel/PanelContainer.js
SimonFinney: Awesome! I wonder if we should set `no-useless-computed-key` as a global warning given that `requiredIfGivenPropExists`, seems like it could be a valuable helper method.

src/components/Panel/PanelContainer.js
jendowns: @SimonFinney  If you set `hasScrollingContent` to `true`, you would have to provide an `aria-label`. Otherwise you have a bunch of other aria props and a tabindex that aren't right without the label. For a consumer, maybe it would be appropriate for them to set `hasScrollingContent` to `true` on their end and ensure they provide a valid label, without even trying to access height of content, which wouldn't be performant.

But I just wanted to also reiterate that this is how things are implemented in the `Modal` currently, and I think it makes sense to keep them consistent. I don't see a reason here to deviate from the upstream library...

src/components/Panel/PanelContainer.js
jendowns: I think in this case, `aria-label` is meant to be used if you don't have a `title`. Which some folks don't have, for designs or whatever the case may be.

SimonFinney: Ah, that makes sense - Thanks for clarifying!

src/components/Panel/PanelContainer.js
SimonFinney: I don't think we should deviate from the upstream library, but I'm definitely conscious of the effect this may have on existing consumers, dynamic content as highlighted above, and E2E tests.

Perhaps, the best route here is to have `hasScrollingContent` default to `true` as you suggested on Slack!

src/components/Panel/PanelContainer.js
```diff
@@ -112,6 +118,10 @@ export default class PanelContainer extends Component {
     });
   }
 
+  panelInstanceId = `panel-${getInstanceId()}`;
```
SimonFinney: This is minor, but can we omit the `panel` prefix here from the static variable names (`panelInstanceId`, etc.) as it's already implied that we're in the scope of the `Panel` as per the `footer` and `header` below?

```suggestion
  instanceId = `${namespace}--${getInstanceId()}`;
```

src/components/Panel/PanelContainer.js
```diff
@@ -142,18 +152,48 @@ export default class PanelContainer extends Component {
       secondaryButton,
       subtitle,
       title,
+      hasScrollingContent,
+
+      // Explicitly define so they aren't in `...other`.
+      disableEscape,
```
SimonFinney: I wasn't able to find where this is or isn't being used - can you confirm that we need this?

src/components/Panel/PanelContainer.js
jendowns: I can definitely update some of the variable names, but `panel-${getInstanceId()}` adds is a bit different from `${namespace}--${getInstanceId()}`. Namespace is already getting added later. So with the first you end up with a nice id like `security--panelv2__title--panel-1`, and with the second you get something longer like `security--panelv2__title--security--panelv2-1`

src/components/Panel/PanelContainer.js
jendowns: @SimonFinney Yeah did mention that but unfortunately it would be a breaking change 😅 It would throw a proptypes error because if you set `hasScrollContent={true}`, you have to also provide an `aria-label`. So this would cause issues for existing implementations.

jendowns:
@SimonFinney I think I addressed all of your feedback except for the comment about setting `hasScrollingContent={true}` ---

I need to look into that more for the reason I posted above (i.e., it would be a breaking change by requiring `aria-label` be provided)

But also, as I was testing it, even when providing `aria-label` when setting `hasScrollingContent={true}`, I was getting a proptypes error. I think this is because our `Panel`/`PanelV2` stories actually utilize a `Portal`, so the proptypes error is based on "hidden" / not yet visible content, if that makes sense. 🤔 But I need to dig into it further. To be safe, I *may* just need to revamp this a bit.

src/components/Panel/PanelContainer.js
SimonFinney: Oh! I wonder in that case is it enough to remove the `panel` prefix entirely?

src/components/Panel/PanelContainer.js
SimonFinney: As I understand it, if there are any (hypothetical) tests expecting to be able to keyboard-tab to the content, this will also affect it, right?

src/components/Panel/PanelContainer.js
jendowns: I'm okay with either way, but honestly I kinda like it reading like "panel 2" or whatever, in the modifier.

src/components/Panel/PanelContainer.js
jendowns: Yes when writing a test for this you'd have to ensure that `hasScrollingContent={true}` before running the `userEvent` tab test otherwise it would fail.

jendowns:
@SimonFinney I believe this is ready to review again.

I made the following changes:
* Per your comment in your first review (https://github.com/carbon-design-system/ibm-security/pull/308#discussion_r367330911), I went ahead and changed the label for the `ariaLabel` applied to the panel wrapper to be `this.props['aria-label'] || title || subtitle` (after thinking about it, your point makes more sense)
* I kept `hasScrollingContent` as `false` by default so as not to introduce a breaking change (as mentioned/explained here: https://github.com/carbon-design-system/ibm-security/pull/308#discussion_r368030758)
* I removed the spread attributes on the wrappers that I had previously added in one of my first comments -- there was strangeness with trying to prevent a user-provided `aria-label` from being overridden, so I decided to leave the spread attr out entirely -- if we want to add that, we could totally do it in a separate PR & evaluate it then.
* I updated the `PanelV2` stories to include a custom `aria-label` to show it being applied in context.

However, full disclosure -- if you set `hasScrollingContent` to `true`, you will still see the following prop type error in development:

```
checkPropTypes.js:20 Warning: Failed prop type: You must provide a value for `aria-label` in `PanelV2` if `hasScrollingContent` exists.
    in PanelV2 (created by PanelV2Story)
    in PanelV2Story (created by storyFn)
    in div (created by Story)
    in Story (created by storyFn)
    in Theme (created by Wrapper)
    in Wrapper (created by storyFn)
    in Container (created by storyFn)
    in storyFn
    in ErrorBoundary
```

I believe this is an error related to Carbon's `requiredIfGivenPropExists` utility. You can verify this by adding `hasScrollingContent` to the `Modal` story as well. Even if one of our panels or the Carbon `Modal` has the required `aria-label` prop, the proptypes error will still appear in the console. But I think its still worthwhile to keep this check, since it is so important for keeping the panel accessible. The proptype error will only appear in development, and if `hasScrollingContent` is `true`... and currently, it would appear for the `Modal` as well if the same prop is applied.

src/components/PanelV2/PanelV2.stories.js
```diff
@@ -39,8 +38,9 @@ const { interactive01, text04 } = theme;
 const closeButtonLabel = 'Close';
 
 const props = {
-  title: label,
-  subtitle: label,
+  title: 'Example title',
```
SimonFinney: This is minor, and perhaps a follow-up, but it would be worth exposing these as knobs for consumers to experiment with

jendowns: Thanks! Yeah I'll put that in a quick follow up after sprint exit 👍

PR #309, opened 2020-01-13
fix(trending-card): use Link component in test

jendowns:
Tests generated the following error for `TrendingCard` --

```
    console.error node_modules/react-dom/cjs/react-dom.development.js:530
      Warning: <Link /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.
          in Link (created by TrendingCard)
          in TrendingCard (created by Default)
          in Default
```

## Proposed changes

- ~use HTML element `a` instead of `Link` for snapshot test~ make sure to import the `Link` component to use as a custom component in test

### DIFF
--- a/src/components/TrendingCard/__tests__/TrendingCard.spec.js
+++ b/src/components/TrendingCard/__tests__/TrendingCard.spec.js
@@ -7,6 +7,7 @@ import { render } from '@testing-library/react';
 import React from 'react';
 
 import { Default as TrendingCard } from '../TrendingCard.stories';
+import Link from '../../Link';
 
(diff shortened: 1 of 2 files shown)


SimonFinney:
@jendowns This test validates that we can pass an additional React element outside of the default `a` tag and that it renders correctly.

At a glance, we most likely need to dig into the linting rules here and adjust accordingly!

jendowns:
Oh thanks @SimonFinney! I just pushed an update that uses the `Link` component instead -- to make sure it imports the `Link` & also passed it in as a react component.

PR #312, opened 2020-01-14
feat(a11y): add DAP and Axe matchers to Jest, add TrendingCard tests

### LINKED ISSUE #248: Add AAT with Axe and DAP for accessibility testing
#### Context

See: https://github.com/carbon-design-system/carbon/pull/4621

We should follow Carbon core's example by adding an AAT configuration to be able to test components in this library with Axe and DAP.

#### Acceptance criteria

- [ ] add AAT configuration
- [ ] add a11y tests to the `Card` component

jendowns:
## Affected issues

- Resolves #248 

## Proposed changes

- add `@ibm/aat` & `axe-core` for a11y testing
- add Jest matchers similar to Carbon's: https://github.com/carbon-design-system/carbon/tree/ac14f81df750dde81d125e6a2a036e476723b5b9/tasks/jest/matchers
- add a11y tests to `TrendingCard`

## Testing instructions

To check that the DAP and Axe tests are actually checking the `TrendingCard`, I suggest pulling down this branch and adding something critically invalid to the `TrendingCard`, like `role="blah"`, and then running `yarn test` to see if the DAP & Axe checkers throw an error.


src/components/TrendingCard/__tests__/TrendingCard.spec.js
```diff
@@ -19,4 +19,20 @@ describe('TrendingCard', () => {
...
+    it('should have no Axe violations', async () => {
+      const { container } = render(<TrendingCard />);
+      await expect(container).toHaveNoAxeViolations();
+    });
+
+    it('should have no DAP violations', async () => {
+      const { container } = render(
+        <main>
+          <TrendingCard />
+        </main>
```
jendowns: Context: DAP tests should be wrapped in a landmark element, otherwise they will throw a violation for orphaned content.

aat/aat.js
```diff
@@ -0,0 +1,46 @@
+module.exports = {
```
SimonFinney: Will this require an authentication token or how does this work?

jendowns: Nope! It doesn't require an auth token & is currently working without one.

SimonFinney: Amazing!

package.json
```diff
@@ -86,6 +87,7 @@
     "@babel/register": "^7.4.0",
     "@commitlint/cli": "^8.0.0",
     "@commitlint/config-conventional": "^8.0.0",
+    "@ibma/aat": "https://able.ibm.com/tools/dist/ibma-aat-2.0.3-alpha.tgz",
```
SimonFinney: 👏

config/jest/setupAfterEnv.js
```diff
@@ -0,0 +1,22 @@
+/**
```
SimonFinney: We can include these definitions in config/jest/index.js` to avoid an additional configuration file

.gitignore
```diff
@@ -26,5 +26,5 @@ src/globals/grid/css-gridish/bx-grid.sketch
 # Generated by npm@5 - project currently uses Yarn.
 package-lock.json
 
-
-
+# AAT test results from DAP/Axe
+results/
```
SimonFinney: This should be `.aat/results`, right, or am I missing something? 🤔

```suggestion
.aat/results
```

aat/aat.js
SimonFinney: 👏

config/jest/matchers/toHaveNoAxeViolations.js
```diff
@@ -0,0 +1,78 @@
+/**
+ * Copyright IBM Corp. 2016, 2018
```
SimonFinney: Because we leverage `babel-jest`, we should be able to use the `import` syntax for these files

config/jest/matchers/toHaveNoDAPViolations.js
```diff
@@ -0,0 +1,24 @@
+/**
```
SimonFinney: 👏

src/components/TrendingCard/__tests__/TrendingCard.spec.js
SimonFinney: Thanks for the context! If we always know that we'll need to wrap these in a `main`, would it be worth including that as part of the `toHaveNoDAPViolations` matcher?

src/components/TrendingCard/__tests__/TrendingCard.spec.js
```diff
@@ -19,4 +19,20 @@ describe('TrendingCard', () => {
       render(<TrendingCard element={Link} />).container.firstChild
     ).toMatchSnapshot();
   });
+
+  describe('automated accessibility testing', () => {
```
SimonFinney: I'd like to see if there's a way for us to better integrate these with our snapshot tests, so that we can test a11y against  the same snapshots - For example the default `TrendingCard`, and the custom link variation.

src/components/TrendingCard/__tests__/TrendingCard.spec.js
jendowns: I thought about that but there will likely be instances where we are testing a component that is already inside of a landmark element, so wrapping everything in a `main` may cause issues. I think it's better to adjust per component since it's such a small thing (plus if you forgot to wrap in a landmark element when you were supposed to, it's a good way to catch if you forgot to check the tests? 😅 )

.gitignore
jendowns: The `results/` folder is where the tests are going though...

Looking at the config:

```
 // optional - Where the scan results should be saved.
  // Default: results
  outputFolder: '.aat/results',
```

AAT appears to be ignoring the `outputFolder` being set. 🤔 
Just gonna... change the outputFolder to the default to avoid confusion. Weird that it ignores that though.

src/components/TrendingCard/__tests__/TrendingCard.spec.js
jendowns: I'm having trouble seeing what you are saying...

I thought with this setup, we are already using the story defined in the storybook. And all tests are running through that. I don't know how you'd take a snapshot file's output and run that through AAT? And I wonder even if you should, since they are different? 🤔 Maybe you could post an example of what you mean?

src/components/TrendingCard/__tests__/TrendingCard.spec.js
SimonFinney: That sounds good! Let's run with this for now and we can tweak later down the line if needed

src/components/TrendingCard/__tests__/TrendingCard.spec.js
SimonFinney: Thanks - I understand your confusion! What I think I'm trying to tease out is whether we should have a11y tests separate from the snapshot tests, or if they should be integrated.

For example, this component only has 2 variations, 'default' and 'custom link', but that in turn would add 4 a11y tests. I can imagine a situation in which we have multiple variations with snapshots and that would increase the testing suite exponentially. I'm wondering if we can do something like the following to reduce the maintenance burden:

```js
it('renders', async () => {
  const { container } = render(<TrendingCard />);
  
  expect(container.firstChild).toMatchSnapshot();

  await expect(container).toHaveNoAxeViolations();
});
```

jendowns: Thanks @SimonFinney I see what you mean now... and I question whether it would be helpful to group these checks into a single testing block. When you say this would "increase the testing block exponentially" are you referring to the testing blocks alone or something else? 

The blocks are helpful for group things together in a way that quickly scannable & follows Carbon's example (similar to: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Loading/Loading-test.js#L20). It also makes things easier to upset in the future, because you only have one related test to update per block, in case something changes. Putting more things in one block feels like putting too much in a function, if that makes sense. 😅

SimonFinney: I certainly agree about having too much happening within testing blocks, but I feel like for snapshot tests it's less of an issue as they're validating the render output of the component, similar to a11y tests.

I suppose the point I'm getting at is that we're going to need to define these a11y tests for each variation, to account for all edge-cases. Looking at Carbon's `Loading` example, they specify a11y tests for the default example, but they are most likely missing coverage [here] (https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Loading/Loading-test.js#L97) and [here](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Loading/Loading-test.js#L111). It seems like a missed opportunity to not reuse these variations given that they're intended to be an accurate representation of each of the component's states.

If you feel like it's outside the scope of this PR, I can create an issue to investigate how we might integrate these better, but I foresee a lot of repetition and maintenance burden ahead if we don't have some synchronicity between the 2.

jendowns: Thanks @SimonFinney that's a good point -- do you mind if we go the route of making a separate PR to explore this? 🙏  I'd like to see how you tackle it. Because with regards to maintenance burden, I'm having trouble seeing how putting things in the same block wouldn't make things harder to extend/revise in the future (as in, sacrificing that flexibility for the sake of avoiding a little repetition).

PR #315, opened 2020-01-15
feat(filter-panel): create composable version with subcomponents

### LINKED ISSUE #236: Request - Composable FilterPanel with sub components
### Details

- **Application / team -** Cloud Pak for Security
- **Needed by -** 2020 Q1

### Proposal

It has long been pointed that the current FilterPanel implementation is not flexible enough to fulfill all filter panel requirements. The primary shortcomings are: 
  1. Consumers have to compose a complex and repetitive data object for the component to render correctly. 
  2. Not all filter panels needs fit the Category → Subcategory → Filter structure.
  3. Consumers desire to provide more controls other than checkboxes for filters.

To follow the updated APIs of complex components, it makes sense to expose FilterPanel subcomponents to allow consumers to build their own FilterPanels. Here is a potential implementation that I've been toying with:

```jsx
<FilterPanel labelText="">

  <FilterSearch labelText="Find filter" onChange={handleFilterSearchChange}>
    {/* render filter results here */}
   </FilterSearch>
  
  <FilterGroup labelText="Data source">
    <FilterCheckbox labelText="" count={1} id="" checked={false} onChange={onChange} />
    <FilterCheckbox labelText="" count={1} id="" checked={true} onChange={onChange} />
    <FilterCheckbox labelText="" count={1} id="" checked={false} onChange={onChange} />
  </FilterGroup>

  <FilterGroup labelText="Observed data">
    <FiltersAccordion labelText="User account">
      <FiltersAccordionItem labelText="User ID">
        <FilterCheckbox labelText="" count={1} id="" checked={false} onChange={onChange} />        
        <FilterCheckbox labelText="" count={10} id="" checked={true} onChange={onChange} />
      </FiltersAccordionItem>
    </FiltersAccordion>
  </FilterGroup>

</FilterPanel>
```

### Goals / use cases

- [ ] Expose subcomponents
- [ ] Make the FilterPanel composable like other complex components such as the UIShell and DataTable components
- [ ] Give more flexibility for consumers to manage their own filter data (instead of prescribing the data structure)
- [ ] Make it easy to introduce of new filtering controls (sliders, radio buttons, graphs)

diego-codes:
## Affected issues
- Resolves #236 

## Proposed changes

- Creates a new variation of the `FilterPanel` component that accepts a `children` prop instead of `filterData` to render a composable panel
- Creates and exports new subcomponents `FilterAccordion`, `FilterAccordionItem`, `FilterCheckbox`, `FilterGroup`, `FilterLabel`, `FilterSearch`
- Creates tests for new subcomponents
- Create a storybook preview demonstrating new filter panel capabilities
- Maintain existing filter panel functionality with the `filterData` prop with a deprecation warning

## Testing instructions

- Check out the Storybook default `FilterPanel` preview


src/components/FilterPanel/FilterPanel.js
```diff
@@ -1,107 +1,82 @@
...
 import { getComponentNamespace } from '../../globals/namespace';
-import * as defaultLabels from '../../globals/nls';
+import deprecatedProp from '../../globals/prop-types';
+
+import FilterAccordion from './FilterAccordion';
+import FilterAccordionItem from './FilterAccordionItem';
+import FilterCheckbox from './FilterCheckbox';
+import FilterGroup from './FilterGroup';
+import FilterLabel from './FilterLabel';
+import FilterSearch from './FilterSearch';
```
areddon: I do not think that you would want to import all of these components inside `FilterPanel` as some of them may not be used at all. If the consumer uses filterData and the `LegacyFilterPanel`, then they would not use any of the other subcomponents.

diego-codes: That's a good point. I was trying to follow how the `DataTable` components are exported, which adds all subcomponents to the `DataTable` component object. @SimonFinney and @jendowns , do you have a preference?

jendowns: @diego-codes I'm cool with keeping this similar to the `DataTable` but the one thing I would ask is to keep `FilterPanel` as a default export, for backwards compatibility in case someone is currently importing it directly.

diego-codes: Okay, I exported the `FilterPanel` as the default component. 

@jendowns it doesn't seem like you feel strongly for adding the subcomponents to the `FilterPanel` object. As I am looking through this, I agree with Alex. One of the benefits of creating composable components is that consumers only import what they need and ignore what they don't need. Adding subcomponents to the `FilterPanel` component would negate this benefit by shipping code to users' browsers that does not get used.

jendowns: @diego-codes Yeah that's right, I don't feel strongly about it & you both make a compelling argument. 👍

src/components/FilterPanel/LEGACY_FilterPanel/FilterSelector/FilterSelector.js
```diff
@@ -0,0 +1,38 @@
...
+import React from 'react';
+
+import { getComponentNamespace } from '../../../../globals/namespace';
+
+import { filterFiltersPropTypes } from '../FilterPanelUtilities';
+import FilterCheckbox from '../../FilterCheckbox';
+
+export const namespace = getComponentNamespace('filter-selector');
+
+const FilterSelector = ({ filter, onChange }) => (
```
jendowns: Two things --

- I was wondering why `FilterCheckbox` has to be wrapped like this, as it looks like just an `onChange` is getting added?

- Also, could you add an `...other` if you want to keep this?

src/components/FilterPanel/FilterPanel.js
```diff
@@ -1,107 +1,60 @@
...
-  };
+  const { title, children, className, filterData } = props;
+
+  if (filterData) {
+    return <LegacyFilterPanel {...props} />;
+  }
 
   return (
-    <aside className={namespace}>
+    <aside className={classnames(namespace, className)}>
```
jendowns: Instead of an `aside` could we use a `section`? The reason is because the DAP tool flags this for the following --

```

ASIDE elements or elements having a WAI-ARIA 'complementary' role must have a label specified with 'aria-label' or 'aria-labelledby'.?
--
ASIDE element or element having a WAI-ARIA 'complementary' role is missing a label specified with 'aria-label' or 'aria-labelledby'.

```

Related to this checkpoint: https://www.ibm.com/able/guidelines/ci162/bypass_blocks.html

src/components/FilterPanel/FilterAccordionItem/FilterAccordiontItem.js
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * Optional color for the scroll gradient.
+   */
+  scrollGradientColor: PropTypes.string,
+
+  /**
+   * Optional count of unique values.
+   */
+  count: PropTypes.number,
```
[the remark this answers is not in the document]
diego-codes: Oops, it's meant to be a number. I'll fix it there instead.

src/components/FilterPanel/FilterCheckbox/_mixins.scss
```diff
@@ -0,0 +1,39 @@
...
+    @include carbon--type-style($name: caption-01);
+
+    // Add visual parentheses.
+    &::before {
+      content: '(';
+    }
+
+    &::after {
+      content: ')';
+    }
```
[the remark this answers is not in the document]
diego-codes: What about this makes you nervous? The parentheses are visual flourishes that don't carry any content meaning.

src/components/FilterPanel/LEGACY_FilterPanel/FilterSearch/FilterSearch.js
```diff
@@ -0,0 +1,188 @@
...
+        };
+      }, {});
+  };
+
+  render() {
+    const {
+      filterData,
+      labels,
+      filterSearchLabel,
+      noFiltersResultsLabel,
```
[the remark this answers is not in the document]
diego-codes: This is a legacy component, so I'd rather not change it more than necessary. Also, this component is not exposed to consumers. It's only used internally, in which case adding an `other` spread does not seem useful. What do you think?

src/components/FilterPanel/LEGACY_FilterPanel/FilterSearch/FilterSearch.js
```diff
@@ -0,0 +1,188 @@
...
+
+  /** @type {object} Externalized global labels */
+  labels: defaultLabels.propType,
+};
+
+FilterSearch.defaultProps = {
+  onChange: () => {},
+  initialSearchValue: '',
+  filterSearchLabel: '',
+  noFiltersResultsLabel: '',
```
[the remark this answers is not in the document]
diego-codes: I believe this is how the props are set up in the current filter search, I just moved this file. I'm not sure why GH is saying that I created this file.

src/components/FilterPanel/FilterCheckbox/_mixins.scss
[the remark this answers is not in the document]
diego-codes: This approach is actually how GH adds visual flourishes as well: 
image

src/components/FilterPanel/LEGACY_FilterPanel/FilterSelector/FilterSelector.js
diego-codes: 1. This component is a legacy component and is expected to work with a filter object that includes specific properties likes `enabled`, `id`, `count`, and `name` among others. To reduce the amount of duplicate code, I am reusing the new `FilterCheckbox` component to achieve the visuals, while maintaining the legacy data requirements.

2. Same comment about not needing to add `...other` to legacy components.

src/components/FilterPanel/FilterCheckbox/_mixins.scss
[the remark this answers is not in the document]
jendowns: The GH diff situation is more extreme to me, because that's even less of a visual flourish... there doesn't seem to be any indication that a given line is being added or reviewed, except through visual indicators with colors + the `:before` content? I tried to see what it reads as in VoiceOver, and it just reads the line minus +/- or any context about adding/removing the line.

With the count parentheses, VoiceOver reads as "Filter label 10" or whatever, where 10 is just part of the line... that seems kinda weird to me and not as helpful as it could be. It would be better if there was a pause and then "10 items" or something of that nature. You could add screen-reader only text inside the parenthesis, after the count number, to achieve this? 🤔

src/components/FilterPanel/index.js
```diff
@@ -1,6 +1,12 @@
 /**
- * @file Filter panel entry point.
- * @copyright IBM Security 2019
+ * @file Filter panel and subcomponents entry point.
+ * @copyright IBM Security 2020
  */
+export { default } from './FilterPanel';
```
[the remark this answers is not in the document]
diego-codes: I'm pretty sure that what you're suggesting is technically bad syntax and it isn't supposed to work:
https://esdiscuss.org/topic/re-export-default#content-2

src/index.js
```diff
@@ -45,7 +45,15 @@ export {
...
-export FilterPanel from './components/FilterPanel';
+export {
+  default,
+  FilterAccordion,
+  FilterAccordionItem,
+  FilterCheckbox,
+  FilterGroup,
+  FilterLabel,
+  FilterSearch,
+} from './components/FilterPanel';
```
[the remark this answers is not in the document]
diego-codes: I'm pretty sure that what you're suggesting is technically [bad syntax](https://esdiscuss.org/topic/re-export-default#content-2) and it isn't supposed to work which is why VS code freaks out about this.
image

versus 
image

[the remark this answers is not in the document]
diego-codes: It appears that Carbon uses this syntax as well: https://github.com/carbon-design-system/carbon/blob/3c0e001eea395a828a66871a6b169a45e9321459/packages/components/src/globals/js/components.js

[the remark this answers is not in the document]
jendowns: Hey @diego-codes I'm totally open to changing the structure of these exports. That's an interesting conversation you linked to and hopefully we can find one that's a little newly than 5 years old. But anyway, in this case, I'm just asking for consistency with the rest of the file. 👍

src/components/FilterPanel/index.js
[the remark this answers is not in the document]
jendowns: Oh actually I mean, for consistency with other files:

```suggestion
export default from './FilterPanel';
```

src/components/FilterPanel/FilterCheckbox/_mixins.scss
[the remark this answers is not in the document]
diego-codes: Okay, I've added a `countLabel` prop that I think will with this.

src/index.js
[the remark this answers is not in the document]
diego-codes: That's fair. I'll make it consistent.

src/components/FilterPanel/FilterAccordionItem/_mixins.scss
```diff
@@ -34,32 +39,37 @@
...
 
     &__content {
-      padding: 0 0 0 $filter-subcategory__spacing__padding;
+      padding: 0 0 0 $filter-accordion-item__spacing__padding;
     }
   }
 
+  // Reduce the width of the label so long labels with counts overflow out of the panel.
+  #{$carbon--accordion}__title {
+    width: calc(100% - 2.8em);
```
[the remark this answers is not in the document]
diego-codes: Yeah, I meant to use `em`. The value comes from the space that the arrow next to the accordion item title takes. Unfortunately, this is the only way I have found to prevent this from happening: 

image

diego-codes:
@jendowns I believe I've resolved all of your feedback. This is ready to review again.

SimonFinney:
@j1mie I know you mentioned you needed a customized `FilterPanel` for designs in your offering - would you be able to review this to ensure it's suitable for your needs?

src/components/FilterPanel/FilterAccordion/FilterAccordion.js
```diff
@@ -0,0 +1,70 @@
...
+  ...other
+}) => (
+  <FilterGroup
+    className={classnames(namespace, className)}
+    titleClassName={`${namespace}__title`}
+    title={title}
+    count={count}
+    countLabel={countLabel}
+  >
+    <Accordion {...other}>{children}</Accordion>
```
jendowns: @diego-codes I was thinking more that the `{...other}` would go on the parent/root.

src/components/FilterPanel/FilterSearch/FilterSearch.js
```diff
@@ -1,188 +1,58 @@
...
+  return (
+    <div
+      className={namespace}
+      onBlur={hideResultsIfLostFocus}
+      onFocus={() => setDisplayResults(true)}
+    >
+      <Search
+        size="sm"
+        className={classnames(`${namespace}__input`, className)}
+        {...other}
```
jendowns: Could you put the spread on the parent/root element?

src/components/FilterPanel/FilterAccordion/FilterAccordion.js
diego-codes: So I was thinking that since the `FilterAccordion` is a wrapper around the `Accordion` component, it would be more useful to expose the `Accordion` to consumers instead. What do you think?

src/components/FilterPanel/FilterSearch/FilterSearch.js
diego-codes: Same thinking about exposing the `Search` component to consumers.

src/components/FilterPanel/FilterAccordion/FilterAccordion.js
jendowns: Good point 👍

src/components/FilterPanel/FilterSearch/FilterSearch.js
jendowns: Yeah that makes sense 👍

diego-codes:
Thanks, @jendowns! Not a problem, I definitely want more eyes on this!

diego-codes:
Thanks for the feedback, @j1mie. To answer your questions:

* Overall, I feel that by making both APIs accessible through the same component makes it more seamless to transition between the two, especially when the legacy filter panel is removed. However, I realize that this approach makes the updates less visible than having a `FilterPanelV2` that is clearly very different from the legacy one. What do y'all think?

* Though the `FilterPanel` is less complex than the new `DataTable`, I agree that I can do better with documentation.

* Yeah... I'm not sure what the best way to go about it is. Visually everything seems to be well aligned until the outline is displayed on the accordion. Do you have a suggestion on how to make the focus state look better?

diego-codes:
Thanks, @jendowns. That was a good suggestion! I actually just updated the width and positioning of the `:before` element which gave me more freedom to widen the visible outline without increasing its height. 

This is what it looks like now:
image
What do you think @j1mie and @jendowns?

src/components/FilterPanel/FilterAccordion/FilterAccordion.js
```diff
@@ -0,0 +1,70 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import { Accordion } from '../../Accordion';
+import { getComponentNamespace } from '../../../globals/namespace';
+import FilterGroup from '../FilterGroup';
+
+const namespace = getComponentNamespace('filter-accordion');
+
+const FilterAccordion = ({
```
diego-codes: What do y'all think of these subcomponent names? I think `FilterPanelAccordion` would make more sense, but it's just so long. Especially for the `FilterAccordionItem` which would be `FilterPanelAccordionItem`. However, I see that the `SummaryCard` subcomponents all have `SummaryCard` in front of them. So maybe length doesn't matter?

jendowns: Honestly I don't particularly mind the additional length, especially if it adds clarity 👍

diego-codes: Okay, I'll replace all `Filter...` with `FilterPanel...`

diego-codes:
@j1mie and @SimonFinney this PR is ready for your review/approval.

jendowns:
@SimonFinney Any feedback/concerns before we merge this? 👀

package.json
```diff
@@ -102,6 +102,7 @@
     "@storybook/theming": "^5.2.4",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
+    "@testing-library/user-event": "^8.0.3",
```
SimonFinney: 👏

SimonFinney:
@diego-codes Thanks for your patience! I'm curious to know if it's intentional that the count in the accordion is now a heavier font weight, or if this was an oversight?

## Legacy

LEGACY_FilterPanel

## Refactored

FilterPanel

diego-codes:
Good catch @SimonFinney. It's a bit of both. It was an oversight originally, but I kept it because it broke up the counts visually now that accordions can display a count as well. It also matches the type hierarchy of the text it goes with. What do you think? I can change it to regular weight if you think that's better.

SimonFinney:
@diego-codes I have no strong preference, and I agree with your rationale - I wanted to point it out in case it was an oversight!

PR #321, opened 2020-01-21
fix(header): prevent header popover's width from spanning browser window

sbutler2901:


sbutler2901:
@jendowns thanks for the feedback, I didn't realize Addons was taking a mobile first strategy here! I was replicating what we have done in CLX. I'll see if I can improve this PR based on your suggestions.

SimonFinney:
@cameroncalder / @petervachon would ye have any thoughts in how the header popover should behave at medium breakpoints?

petervachon:
For the popover, I would keep the same behavior on medium and large breakpoints and only go full-width for the small breakpoint.

PR #324, opened 2020-01-23
fix(panelv2): update tests to use a11y matchers and user events

jendowns:
## Affected issues

- ~BLOCKED BY #308~ No longer blocked!
- follow up to #308 

## Proposed changes

- improve `PanelV2` tests by adding a11y tests + checking for user events like clicking to close the panel & tabbing through panel content


jendowns:
@diego-codes I just added you as a reviewer, if you have time --

Keeping in mind that current there are two DAP violations right now and that #308 will ultimately help with that, I was wondering if you could take a look at these tests and let me know if you agree with the structure etc? And if you think there are gaps? Thank you so much 🙏

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,121 @@
...
+import { cleanup, render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
 
 import { Button, PanelV2, PanelContent } from '../../..';
 
 describe('PanelV2', () => {
-  it('renders', () => {
-    const wrapper = mount(
+  afterEach(cleanup);
```
diego-codes: I'm pretty sure you don't need to do clean up anymore. The package does this already.

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,121 @@
...
+      <PanelV2
+        closeButton={{
+          label: 'test close',
+          onClick: onCloseMock,
+        }}
+      />
+    );
+
+    userEvent.click(getByLabelText(/test close/i));
+    expect(onCloseMock).toHaveBeenCalled();
```
diego-codes: Sometimes I take the assertion a bit further by using `toHaveBeenCalledTimes(1)` just in case so it doesn't call it multiple times.

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,121 @@
...
+          label: 'test close',
+        }}
+        renderFooter={() => <Button>test footer button</Button>}
       >
-        <PanelContent>{label}</PanelContent>
+        <PanelContent>test content</PanelContent>
       </PanelV2>
     );
-    expect(wrapper).toMatchSnapshot();
+    await expect(document.body).toMatchSnapshot();
```
diego-codes: For this test, what exactly are you wanting to test for that can't be achieved with specific assertions or just looking at a storybook story? The reason I'm asking is that snapshots are so challenging to maintain and identify when real bugs appear. We've started to move away from them.

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,121 @@
...
+      >
+        <PanelContent>test content</PanelContent>
+      </PanelV2>,
+      {
+        // DAP requires a landmark '<main>' in the DOM:
+        container: document.body.appendChild(main),
+      }
+    );
+    await expect(document.body).toHaveNoAxeViolations();
+    await expect(document.body).toHaveNoDAPViolations('PanelV2');
```
diego-codes: I'm not familiar with this assertion what does this line mean?

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,121 @@
...
+        closeButton={{
+          label: 'test close',
+        }}
+      />
+    );
+
+    expect(queryByText(/test title/i)).not.toBeInTheDocument();
+  });
+
+  it('should cycle panel elements in tab order', () => {
```
diego-codes: Does this actually test `PanelV2` logic or is it testing some focus trap component or tool that's not actually part of the panel code? If its the latter, it might be more appropriate to test the focus trap component if it is an internally developed one.

src/components/PanelV2/__tests__/PanelV2.spec.js
jendowns: @diego-codes I was keeping this one out of habit, to be honest. 😅 But you make a good point so I'll go ahead and remove it.

src/components/PanelV2/__tests__/PanelV2.spec.js
jendowns: I added Axe and DAP matchers in this PR, based on similar matchers Carbon added a month or so ago: https://github.com/carbon-design-system/ibm-security/pull/312
Carbon's matchers: https://github.com/carbon-design-system/carbon/tree/ac14f81df750dde81d125e6a2a036e476723b5b9/tasks/jest/matchers

src/components/PanelV2/__tests__/PanelV2.spec.js
jendowns: This is meant to test tab order, but it wouldn't be able to effectively test focus trap anyway because there's no interactive elements outside of this panel in the mock/render. 🤔

src/components/PanelV2/__tests__/PanelV2.spec.js
diego-codes: Oh, I didn't catch that they were two different assertions. Nice! Looks useful.

src/components/PanelV2/__tests__/PanelV2.spec.js
diego-codes: Do you expect the tab order to work any differently than the default browser behavior? If not, this test is testing jsdom instead of any panel logic. 

It looks like the `Portal` component relies on a package called `focus-trap-react` so my answer would not test for focus trapping. However, I see that the trap is activated by the `focusTrap` prop on the `Portal` component so that might be a reason to test even though it is just passing the prop directly to the focus trap. It's your call.

My general guidance is to only test logic that is located in the exact file that you have imported to test (`PanelV2`). If the logic is located in another file in your project, then that file should be tested separately (`Portal`). If the logic comes from a third-party package or library, then it is their responsibility to test their own code (`focus-trap-react`) not yours.

jendowns: Thanks that makes sense -- do you still think this test is worthwhile though, as-is? (And then something separate for focus trap, in the Portal tests most likely)

diego-codes: I think it's good. I just realized that it tabs to the content area first, which makes sense why you'd want to tab through.

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,103 @@
...
+  test('should have no Axe or DAP violations', async () => {
+    const main = document.createElement('main');
+    render(
+      <PanelV2
+        title="test title"
+        subtitle="test subtitle"
+        closeButton={{
+          label: 'test close',
+        }}
+        renderFooter={() => <Button>test footer button</Button>}
```
SimonFinney: To be certain and until we remove them entirely, we should also validate that the `primaryButton` and `secondaryButton` props are not throwing a11y violations

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,103 @@
...
+          onClick: onCloseMock,
+        }}
+      />
+    );
+
+    userEvent.click(getByLabelText(/test close/i));
+    expect(onCloseMock).toHaveBeenCalledTimes(1);
+  });
+
+  test('should not render panel when `isOpen` is set to `false`', () => {
```
SimonFinney: 👏

src/components/PanelV2/__tests__/PanelV2.spec.js
```diff
@@ -3,38 +3,103 @@
...
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
 
 import { Button, PanelV2, PanelContent } from '../../..';
 
 describe('PanelV2', () => {
-  it('renders', () => {
-    const wrapper = mount(
+  test('should have no Axe or DAP violations', async () => {
```
SimonFinney: I'm curious about the choice of `test` versus `it` here?

src/components/PanelV2/__tests__/PanelV2.spec.js
SimonFinney: @diego-codes I'd like to understand more about your decisions to move away from snapshot tests if you wouldn't mind elaborating?

I personally still find value from them as long as they're being used in moderation and in conjunction with a more comprehensive testing suite, but I'd love to hear more about the challenges you've faced with them!

src/components/PanelV2/__tests__/PanelV2.spec.js
jendowns: 😆 I guess it go in my head while reading the React testing library docs & see all of their examples in that format. Though honestly I do kinda like `test` or `it` here, especially because I've flattened out the tests a bit. But do you have a preference?

(One more note actually -- it make help to do new structure with `test` to help us differentiate right away between new/re-evaluated tests and old tests?)

SimonFinney: I've a feeling `it` is just an alias for familiarity between other testing libraries. I like that idea about differentiating!

jendowns:
@SimonFinney I believe I addressed your feedback above & this is ready to review again! Changes from #308 have been integrated as well, so there are no more DAP violations here.

PR #327, opened 2020-01-24
fix(UNSTABLE__pagination): add a11y and unit tests

### LINKED ISSUE #294: Unstable Pagination - add tests
### Proposal

Add unit tests for the new `UNSTABLE__Pagination` component.

jendowns:
## Affected issues

- Resolves #294 

## Proposed changes

Add tests for `UNSTABLE__Pagination` that check...
* interactions (changing the page, changing the items per page)
* missing props (what happens when you don't provide `totalItems`, or a child `PageSelector`)
* `pagesUnknown`
* accessibility against Axe and DAP tools
* checking tab cycle


src/components/UNSTABLE__Pagination/__tests__/UNSTABLE_Pagination.spec.js
```diff
@@ -0,0 +1,413 @@
...
+
+    // Still expect default "of 2 pages" text not to be shown,
+    // because `pagesUnknown={true}` (despite the fact that `totalItems` is provided):
+    expect(queryByText(/of 2 pages/i)).not.toBeInTheDocument();
+
+    // Still expect default text "1-5 items" with 5 being derived from page size,
+    // not the total items (despite the fact that `totalItems` is provided):
+    expect(getByText(/1–5 items/i)).toBeInTheDocument();
+  });
+});
```
jendowns: @diego-codes if you have a moment I was wondering if you could take a look at this last test and let me know what you think...

I'm trying to check if that certain text is display if `pagesUnknown` is `true`, *whether or not* `totalItems` is provided. I think it's important to check both situations, where `totalItems` is provided or not, because `totalItems` is this case is superceded by `pagesUnknown`.

I went with a `rerender` here, in the same test. At first I just rerendered with a valid `totalItems` value, thinking that my previous props would be carried over. But then when things didn't look right, I opted to redefined *everything* in this rerender. Does that make sense to do or would you recommend I split the test?

diego-codes: Hi @jendowns! These two scenarios are two use cases so I would write two separate tests. It's better to think of tests as use cases instead of "features" because we end up with bloated tests like so.

jendowns: Thank you that makes sense! Honestly at this point I'm not sure when/why I'd ever use `rerender`? 😅

diego-codes: Most test cases don't require it. When you truly need it, you'll know 😆… It's useful when the change of a prop over time is required to demonstrate a test case.

src/components/UNSTABLE__Pagination/__tests__/UNSTABLE_Pagination.spec.js
```diff
@@ -0,0 +1,412 @@
...
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
+import { UNSTABLE__Pagination, PageSelector } from '../../..';
+
+import { namespace } from '../Pagination';
+
+describe('UNSTABLE_Pagination', () => {
+  test('should have no Axe or DAP violations', async () => {
```
SimonFinney: Would it make sense to move the a11y tests into it's own `describe` block? I know previously we had blocks for rendering, events, etc. so I'm wondering if we should apply a similar approach here

jendowns: I started out with nested blocks but they cluttered up the file a lot. Personally I'd prefer to leave them out and keep things flat. 🤔 Plus I think at this point, add blocks would be just to see certain indented output in the CI `yarn test` output?

SimonFinney: I'm of 2 minds here - I feel like it's nice to have the grouping to visually separate the tests out, but admittedly it becomes difficult with larger files.

I wonder if there's an alternative approach to reducing the cognitive burden here... My only initial suggestion is to break them out into separate files, but perhaps that's overkill. What are your thoughts here?

jendowns: Yea the visual separation of the extra `describe` blocks isn't easy to distinguish due to the clutter begin/end of the block creates, to me. And at the same time, I don't suggest splitting these tests out into multiple files... it's 400-ish lines right now & a split would increase the cognitive load of having both open at the same time (updating/adding tests would take longer as well).

Again, to me this flat, single file of 400-ish lines was easier to write, update, and read without the clutter or file splitting. 👍 More and more I'm in the camp of less abstraction = easier.

jendowns: Though on the topic of multiple tests files, I am going to make a new file for a couple `PageSelector` tests, I think. At least to check for a namespace class, since that class is relied on so heavily for styles not breaking etc.

SimonFinney: True, I thought it might be overkill...

I agree! That absolutely makes sense to have a suite specifically for the `PageSelector` - this should reduce this file's length and ensure we're only covering the behavior and logic of the `Pagination` here, and help us identify bloat in the process.

src/components/UNSTABLE__Pagination/__tests__/PageSelector.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { PageSelector } from '../../../..';
+
+import { namespace } from '../PageSelector';
+
+describe('PageSelector', () => {
+  test('should have root class with security namespace', () => {
```
SimonFinney: Would you mind elaborating on the purpose of this test?

jendowns: I elaborated on it earlier before I made the change & I believe you said you liked the idea: https://github.com/carbon-design-system/ibm-security/pull/327#discussion_r373153604

> Though on the topic of multiple tests files, I am going to make a new file for a couple PageSelector tests, I think. At least to check for a namespace class, since that class is relied on so heavily for styles not breaking etc.

I'm not sure what more can be said about it. The point about the importance of the class, and making sure it doesn't get deleted due to how heavily it is relied on for styles, is the main/only thing.

SimonFinney: Apologies for any confusion or misdirection - I like the idea of breaking out the `PageSelector` component and testing the behavior and logic of that, but I don't believe this kind of test is entirely necessary.

jendowns: Okay... unfortunately I don't see any "behavior" tests for the `PageSelector` to break out of the other test file. Every test in the main `Pagination` file is related to the pagination component's behavior, as a whole. And the `PageSelector` is just a Carbon `Select`, which they test on their end. So I will delete this file.

jendowns:
@SimonFinney Looking back at this, I believe I addressed your feedback? Can we move this one along?

jendowns:
@SimonFinney just a reminder that it appears I have addressed your feedback. Can you revisit this PR so we can include tests for this component? Thank you

src/components/UNSTABLE__Pagination/__tests__/Pagination.spec.js
```diff
@@ -0,0 +1,414 @@
...
+    userEvent.tab();
+
+    // The "next" page button:
+    expect(getByLabelText(/test next button/i).closest('button')).toHaveFocus();
+
+    userEvent.tab();
+
+    // Loop complete.
+    // The "items per page" select element:
+    expect(pageSizer).toHaveFocus();
```
SimonFinney: 👏

src/components/UNSTABLE__Pagination/__tests__/Pagination.spec.js
```diff
@@ -0,0 +1,414 @@
...
+            labelText="test page selector"
+            onChange={event => onSetPage(event.target.value)}
+            totalPages={totalPages}
+          />
+        )}
+      </UNSTABLE__Pagination>
+    );
+
+    // Expect to see default text for page 1:
+    expect(getByText(/1–5 of 10 items/i)).toBeInTheDocument();
```
SimonFinney: Is it sufficient to omit this assertion, as it's covered in the previous test?

SimonFinney:
@jendowns Thanks for your patience! I'm trying to rehash my memory based on the progress we've made in other PRs...

I love how this is broken out into specific use cases, too. Can we ensure we're covering all props as per other PRs? For example `className` and `...other`? Let me know your thoughts.

src/components/UNSTABLE__Pagination/__tests__/Pagination.spec.js
jendowns: This assertion is setting a baseline / default for the next series of user events and assertions.
I don't think it is necessary to valid the text, however. It is done many times throughout this testing suite so multiple tests would fail if that text suddenly changed. 👍

jendowns:
Thanks @SimonFinney! I just pushed extra tests for custom class names and the spread attribute. Should be good to go now 👍 

Also I found the previous review from last month -- https://github.com/carbon-design-system/ibm-security/pull/327#discussion_r370842417
Looks like that one was resolved 👍

PR #337, opened 2020-01-29
fix(summary-card): add a11y and unit tests

jendowns:
### Tests
- adds tests a11y and unit tests for `SummaryCard` and any sub components that didn't have any tests previously
(NOTE I did not edit any existing tests for the `SummaryCardContainer`, etc, though I wanted to propose a couple updates to those in a follow-up PR. I did go ahead and remove the `SummaryCardBatchActions` snapshot test because it wasn't checking anything worthwhile.)

### Components
- exported namespaces to uses in tests
- fixed an a11y issue in the `SummaryCardAction` (see https://github.com/carbon-design-system/ibm-security/pull/337#discussion_r373093310)


src/components/SummaryCard/SummaryCardAction/SummaryCardAction.js
```diff
@@ -83,6 +83,8 @@ class SummaryCardAction extends Component {
                       iconDescription={closeButtonIconDescription}
                       onClick={this.toggleOpen}
                       kind="ghost"
+                      title={closeButtonIconDescription}
+                      aria-label={closeButtonIconDescription}
```
jendowns: This is in response to an Axe error for the `SummaryCardAction`. While we can definitely add a new prop to allow folks to customize these values further, the easiest thing to do at the moment is use the `closeButtonIconDescription` value.

jendowns:
For the record these tests do not fail locally. 🙃 
For some reason, a `SummaryCardAction` with `loading={true}` is still "interactive" when CircleCI runs, even though a loading `SummaryCardAction` is basically a `Button` with `disabled={true}` 🙃 I'll investigate...

package.json
```diff
@@ -99,9 +99,9 @@
     "@storybook/addons": "^5.2.4",
     "@storybook/react": "^5.2.4",
     "@storybook/theming": "^5.2.4",
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^9.3.2",
-    "@testing-library/user-event": "^8.0.3",
+    "@testing-library/jest-dom": "4.2.4",
+    "@testing-library/react": "9.3.2",
+    "@testing-library/user-event": "8.1.0",
```
jendowns: The trick to fix the tests was to update `@testing-library/user-event`
I suggest we use strict version numbers for each of these so that we can ensure that both our local environment and Circle CI use the same package versions for tests.

jendowns:
@SimonFinney These tests have been updated to include prop tests in subcomponents (except for `SumamryCardAction`, which is just a Carbon component passed through). And the merge conflicts have been resolved 👍

src/components/SummaryCard/SummaryCardBatchActions/__tests__/SummaryCardBatchActions.spec.js
```diff
@@ -9,10 +9,14 @@ import React from 'react';
 import { SummaryCardBatchActions } from '../../../..';
 
 describe('SummaryCardBatchActions', () => {
-  it('renders', () => {
-    expect(
-      render(<SummaryCardBatchActions onCancel={jest.fn()} totalSelected={0} />)
-        .container.firstChild
-    ).toMatchSnapshot();
+  test('should pass through extra props via spread attribute', () => {
```
SimonFinney: Will we need a11y tests for these? Also, we'll most likely also want to cover `translateWithId`

src/components/SummaryCard/SummaryCardBody/__tests__/SummaryCardBody.spec.js
```diff
@@ -0,0 +1,28 @@
...
+ * @file Summary card body tests.
+ * @copyright IBM Security 2020
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { SummaryCardBody } from '../../../..';
+
+describe('SummaryCardBody', () => {
```
SimonFinney: Will we need a11y tests for these also?

src/components/SummaryCard/SummaryCardFooter/__tests__/SummaryCardFooter.spec.js
```diff
@@ -0,0 +1,32 @@
...
+ * @copyright IBM Security 2020
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { SummaryCardFooter } from '../../../..';
+
+describe('SummaryCardFooter', () => {
+  test('should accept a custom class name', () => {
```
SimonFinney: Will we need a11y tests for these also?

src/components/SummaryCard/SummaryCardHeader/__tests__/SummaryCardHeader.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  test('should accept a status as a `node`', () => {
+    const { getByText } = render(
+      <SummaryCardHeader title="test title" status={<span>test status</span>} />
+    );
+    expect(getByText('test status')).toBeVisible();
+  });
+
+  test('should truncate `titel` content when `truncate` is `true`', () => {
```
SimonFinney:
```suggestion 
  test('should truncate `title` content when `truncate` is `true`', () => {
```

src/components/SummaryCard/SummaryCardBatchActions/__tests__/SummaryCardBatchActions.spec.js
SimonFinney: Will we need a11y tests for these?

src/components/SummaryCard/SummaryCardHeader/__tests__/SummaryCardHeader.spec.js
```diff
@@ -0,0 +1,44 @@
...
+ * @file Summary card header tests.
+ * @copyright IBM Security 2020
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { SummaryCardHeader } from '../../../..';
+
+describe('SummaryCardHeader', () => {
```
SimonFinney: https://github.com/carbon-design-system/ibm-security/pull/337/files#r387011400

src/components/SummaryCard/SummaryCardSelect/__tests__/SummaryCardSelect.spec.js
```diff
@@ -19,4 +19,15 @@ describe('SummaryCardSelect', () => {
     );
     expect(document.querySelector('label')).toHaveClass('custom-class');
   });
+
+  test('should pass through extra props via spread attribute', () => {
```
SimonFinney: Same comment as https://github.com/carbon-design-system/ibm-security/pull/337/files#r387011400

jendowns: All tested together inside `SummaryCard`

src/components/SummaryCard/SummaryCardHeader/__tests__/SummaryCardHeader.spec.js
jendowns: All tested together inside `SummaryCard`

src/components/SummaryCard/SummaryCardFooter/__tests__/SummaryCardFooter.spec.js
jendowns: All tested together inside `SummaryCard`

src/components/SummaryCard/SummaryCardBody/__tests__/SummaryCardBody.spec.js
jendowns: All tested together inside `SummaryCard`

src/components/SummaryCard/SummaryCardBatchActions/__tests__/SummaryCardBatchActions.spec.js
jendowns: This is a Carbon prop though? Also, the a11y test is in `SummaryCard`

jendowns:
Hey @SimonFinney I responded to your questions above. Please take another look when you get a chance. To reiterate -- due to the complexity of the components together (and also the simplicity of some subcomponents like `SummaryCardFooter` which mean nothing by themselves & would not be used alone) the `SummaryCard` tests houses tests for all of the subcomponents, to demonstrate how it would be used in product.

jendowns:
@SimonFinney In my previous comments I wasn't 100% clear --

`SummaryCard` houses a big a11y test for `SummaryCard` + all of its small subcomponents like `SummaryCardBody`, `SummaryCardFooter`, etc. It made sense to do this because those subcomponents are small and meant to be used together.

Likewise for your comments about `SummaryCardBatchActions` etc and subcomponents related to batch action + container functionality, those a11y tests were already added to the `SummaryCardContainer` test file in #340. And likewise it makes sense to test batch actions, etc., inside of their intended parent component.

src/components/SummaryCard/__tests__/SummaryCard.spec.js
```diff
@@ -0,0 +1,178 @@
...
+  SummaryCardAction,
+  SummaryCardFooter,
+  Tooltip,
+} from '../../..';
+
+import { carbonPrefix } from '../../../globals/namespace';
+import { namespace as headerNamespace } from '../SummaryCardHeader/SummaryCardHeader';
+
+describe('SummaryCard', () => {
+  test('should have no Axe or DAP violations`', async () => {
```
SimonFinney: I'm torn between whether we should be testing the a11y of this component and sub-components as a whole or in individual unit tests - personally, it feels right to do all of these individually, but I'd like to get your thoughts!

src/components/SummaryCard/SummaryCardAction/__tests__/SummaryCardAction.spec.js
```diff
@@ -0,0 +1,113 @@
...
+ * @copyright IBM Security 2020
+ */
+
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
+import { SummaryCardAction } from '../../../..';
+
+describe('SummaryCardAction', () => {
```
SimonFinney: Will we need tests for `hasIconOnly` and `onClick` also?

src/components/SummaryCard/__tests__/SummaryCard.spec.js
jendowns: The `SummaryCard` a11y test would really mean nothing without the subcomponents included.
And to test the subcomponents individually, after having already included them together as intended inside the `SummaryCard` a11y test, seems redundant to me. 

What are you seeking to gain by rendering and testing each subcomponent individually, rather than doing it this way?

src/components/SummaryCard/SummaryCardAction/__tests__/SummaryCardAction.spec.js
jendowns: Those are `Button` props, passed through to the inner `Button`... 😅 
The only thing we'd really be checking with `hasIconOnly` is if the boolean is doing its thing?

With `onClick`, what specifically would you be testing for / what do you feel is missing here?

SimonFinney: Because these are consumer-facing props, if they were removed, we wouldn't know about it

src/components/SummaryCard/__tests__/SummaryCard.spec.js
SimonFinney: The distinction here is that as we're including these with our unit tests, my assumption is that we would be testing the accessibility of each individual component or unit.

If we want to test them as a whole, I believe that's more in line with integration testing, and we should define a strategy of how we want to proceed with that.

src/components/SummaryCard/SummaryCardAction/__tests__/SummaryCardAction.spec.js
jendowns: I'm not sure what you mean here -- if the prop is removed on our end, then we would know about it because we removed it? 🤷‍♀ 

But I can add a test to move this along. Not exactly sure what I am able to check, since we don't control the icon here, but I'll work something out.

src/components/SummaryCard/__tests__/SummaryCard.spec.js
diego-codes: Because the `SummaryCard` subcomponents depend on the structure of the parent `SummaryCard` and would not be used elsewhere, I agree that it does not make sense to test each subcomponent individually. As @jendowns said, the subcomponents will lose their semantic meaning without living inside of the `SummaryCard`.

To handle these types of situations in DE, we usually have a `Component.test.js` file for unit tests and a `Component.integration.test.js` file for integration tests that test how interacting with one component might alter the behavior of another (usually a child affecting its parent)

jendowns:
@SimonFinney I added tests in `SummaryCardAction` for `hasIconOnly` and `onClick`

src/components/SummaryCard/__tests__/SummaryCard.spec.js
SimonFinney: I agree it's pointless to unit test individually if we want to verify overall semantics - if we have one component that's a `ul` and a subcomponent that's a `div`, we won't identify that this is an issue if these aren't integrated.

I'm trying to tease out the best approach here, as there are a number of options I can think of off the top of my head - These include:

- having dedicated integration tests
- the catch-all approach in `SummaryCard.spec.js`
- an approach where the `a11y` tests for sub-components live as a hybrid integration / unit test in their own testing suite (for example `SummaryCardAction.spec.js`) and require the components needed to verify the semantics (potentially `SummaryCard` and `SummaryCardAction`)

The a11y tests are unique in that we're validating the rendered output, so it's harder for us to point to any methodology and say 'this is the right way to do it' - I'm definitely not suggesting we go with one or the other, more so that we should consider and discuss the possibilities before making a call on it.

src/components/SummaryCard/SummaryCardAction/__tests__/SummaryCardAction.spec.js
SimonFinney: We may know about it, or we may not if we don't catch it! 😆The unit tests are in place to protect the functionality a component is expected to provide in case we introduce a change that breaks it.

src/components/SummaryCard/SummaryCardAction/__tests__/SummaryCardAction.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('should pass through extra props via spread attribute', () => {
+    const { queryByTestId } = render(
+      <SummaryCardAction data-testid="test-id" />
+    );
+    expect(queryByTestId('test-id')).toBeVisible();
+  });
+
+  test('should apply assistive text and icon-only class name when `hasIconOnly` is `true`', () => {
```
SimonFinney: 👏

src/components/SummaryCard/SummaryCardAction/__tests__/SummaryCardAction.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    );
+    // Expect icon description to be applied as assistive text:
+    expect(queryByText(/test icon/i)).toBeVisible();
+    // Expect button to have Carbon's icon-only class:
+    expect(queryByText(/test icon/i).closest('button')).toHaveClass(
+      `${carbonPrefix}btn--icon-only`
+    );
+  });
+
+  test('should invoke click mock when button is clicked', () => {
```
SimonFinney: 🎉

src/components/SummaryCard/__tests__/SummaryCard.spec.js
jendowns: Considering the above, what are you seeking to change here that is currently missing / insufficient? 

I think the conversation is getting very meta now, and we should step back and evaluate the effectiveness of the code.

If the test was broken down and split, that could make it less effective (as you described yourself with the UL and DIV example, and as Diego and I also described above).

If the test is left as-is... then it remains effective, right?

In that case, can we make a final call based on the effectiveness of the test? It appears to be effective, and there are no more effective alternatives.

jendowns: Considering the above, what are you seeking to change here that is currently missing / insufficient? 

I think the conversation is getting very meta now, and we should step back and evaluate the effectiveness of the code.

If the test was broken down and split, that could make it less effective (as you described yourself with the UL and DIV example, and as Diego and I also described above).

If the test is left as-is... then it remains effective, right?

In that case, can we make a final call based on the effectiveness of the test? It appears to be effective, and there are no more effective alternatives.

jendowns:
@SimonFinney I believe I have addressed all of you actionable feedback here? Hopefully we can move this along? Thank you 🙏

PR #338, opened 2020-01-29
docs(readme): update codesandbox link

### LINKED ISSUE #303: CodeSandbox - update version in demo to latest stable release
## Details

The CodeSandbox demo/template linked from the README (https://codesandbox.io/s/codesandbox-nmmqp) is on `1.1.0`, but we've had a number of stable releases since then.

We should update the CodeSandbox demo/template to the latest stable release 👍

jendowns:
## Affected issues

- Resolves #303 

## Proposed changes

- update CodeSandbox link to use instance with updated version of the library

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@
 If you're just getting started and looking for React components, take a look at [our Storybook](https://ibm-security.carbondesignsystem.com).
 
 If you just want to try out Carbon for IBM Security, you can also use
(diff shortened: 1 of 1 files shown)


jendowns:
Thanks @SimonFinney I just added a `^` to the version. I wonder if using a `*` will match version `2.0.0`?

SimonFinney:
That's a good point! I'm not entirely sure - I wonder if because `2.0.0` is currently a prerelease it may not pick that up? I think this is great for now!

PR #340, opened 2020-01-31
fix(summary-card-container): update tests

### LINKED ISSUE #339: SummaryCardBatchActions - buttons in tab order even when they appear "hidden"
## Bug - summary card batch action buttons are always in tab order, even when hidden

**Expected behavior -**

The batch actions toolbar for the `SummaryCardContainer` (called `SummaryCardBatchActions`) should not be interactive when it is "hidden" from view. You shouldn't be able to tab those hidden batch actions buttons.

**Actual behavior -**

The batch actions buttons in `SummaryCardBatchActions` are in the tab order even when they appear "hidden"

### Steps for reproducing

Go the storybook and see that you can tab to "hidden" batch actions buttons: https://ibm-security.carbondesignsystem.com/?path=/story/patterns-summarycard--multiselect

jendowns:
~🚫 Blocked by #339 🚫 
The test `should cycle elements in tab order` will fail until "hidden" batch actions buttons are removed from tab order.~

No longer blocked! 🎉 

## Summary of changes

Resolves #339 

* Removes snapshots in favor of user-centered testing structure based on visible text, labels, etc.
* Adds a11y tests
* Had to add a dynamic `tabIndex` based on `shouldShowBatchActions` prop to both tests and the story (NOTE: this would have to be something a user adds on their end, to ensure their batch actions are accessible & don't disrupt tab order)


src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
```diff
@@ -61,39 +166,120 @@ describe('SummaryCardContainer', () => {
...
+      />
+    );
 
-    it("doesn't show batch actions when nothing is selected", () => {
-      expect(batchActions).not.toHaveClass(batchActionsActiveClass);
-    });
+    // We can't check `not.toBeVisible` because the batch actions
+    // are absolutely positioned & not hidden/removed from the DOM.
+    // So instead, we query the default batch actions text with 0 selected:
+    expect(queryByText(/0 item selected/i)).toBeInTheDocument();
```
jendowns: This will change with #341

src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
```diff
@@ -61,39 +166,120 @@ describe('SummaryCardContainer', () => {
...
+    expect(queryByText(/2 items selected/i)).toBeInTheDocument();
 
-    it('resets selected summary cards', () => {
-      click(select);
+    // Clear batch actions selections:
+    userEvent.click(getByText(/Cancel/i));
 
-      click(container.getByText('Cancel'));
+    // Expect the batch actions text to be updated:
+    expect(queryByText(/0 item selected/i)).toBeInTheDocument();
```
jendowns: This will change a bit with #341 (will need to check `isVisible`?)

src/components/SummaryCard/SummaryCardSelect/__tests__/SummaryCardSelect.spec.js
```diff
@@ -4,17 +4,17 @@
...
 
-import props from '../_mocks_';
-
 describe('SummaryCardSelect', () => {
-  it('renders', () => {
-    expect(
-      render(<SummaryCardSelect className="className" {...props} />).container
-        .firstChild
-    ).toMatchSnapshot();
+  test('should be checked when clicked', () => {
```
jendowns: I don't think this hurts to check but it is technically redundant due to the tests for `SummaryCardContainer` 🤔

src/components/SummaryCard/SummaryCard.stories.js
```diff
@@ -146,6 +146,7 @@ storiesOf(patterns('SummaryCard'), module)
                 action('SummaryCardBatchAction onClick')(selectedSummaryCards)
               }
               renderIcon={Delete16}
+              tabIndex={getBatchActionProps().shouldShowBatchActions ? 0 : -1}
```
jendowns: @SimonFinney I'm curious if you know of a way *we* could add this on our end, without requiring users to add this in their implementation? 🤔 Otherwise any custom `SummaryCardBatchAction` they add will always have `tabIndex={0}` and therefore disrupt tab order when batch actions aren't even showing...

We may just need to add more documentation here.

src/components/SummaryCard/SummaryCardSelect/__tests__/SummaryCardSelect.spec.js
SimonFinney: Good point - in the case of the `SummaryCardContainer`, we're testing the internal logic / negotiation of selected checkboxes. As we pass through our props to the `Checkbox` component this should also be covered by their unit tests. 🤞

I think it's also beneficial to ensure the appropriate `className` is being applied here, and an example of an `...other` prop, as these are behaviors based on consumer input.

src/components/SummaryCard/SummaryCard.stories.js
SimonFinney: We'd need to investigate this further - I'm assuming we could add something to the `SummaryCardContainer` that handles this but I'm not entirely sure just yet 🤔

src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
```diff
@@ -61,39 +139,171 @@ describe('SummaryCardContainer', () => {
...
+
+    userEvent.tab();
 
-    batchActions = getByTestId(batchActionsId);
-    select = getByTestId(`${selectId}${summaryCards[0].id}`);
+    // Loop complete.
+    // The first summary card select:
+    expect(getByLabelText(/test select 1/i)).toHaveFocus();
+
+    userEvent.tab();
```
SimonFinney: Would you mind elaborating on the purpose of this last tab?

src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
```diff
@@ -61,39 +139,171 @@ describe('SummaryCardContainer', () => {
...
+    // Currently there's no way to load the page
+    // with summary cards pre-checked AND batch action toolbar updated.
+    // Therefore, we have to select both cards first:
+    userEvent.click(getByText(/test select 0/i));
+    userEvent.click(getByText(/test select 1/i));
+
+    userEvent.tab();
+
+    // The batch action button:
+    expect(getByText(/test batch action/i)).toHaveFocus();
```
SimonFinney: Because we leverage the `DataTable`s batch actions here, I'm wondering what your thoughts are on where our responsibilities should end for what we should be testing?

It still feels a bit unclear to me now, but I'd like to understand the boundaries so we're not tightly-coupled to upstream functionality when writing unit tests.

src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
jendowns: No purpose for it -- looks like it was left there by mistake.

src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
jendowns: For sure -- I'm not necessarily testing the functionality of the data table batch action button in isolation, in this case; I'm testing it here in relation to tab order (and interactivity) within the summary card container. That's not something necessarily controlled in the Carbon's batch action component -- it's moreso how we use it. I think that's a big distinction in this case. 

I think something out of scope would be testing that the component can take in a custom class, or have other typical Carbon button / batch action button props.

src/components/SummaryCard/SummaryCardSelect/__tests__/SummaryCardSelect.spec.js
jendowns: Good point -- it looks like `id` and `labelText` props I'm using here (which I believe they are required in the `Checkbox` anyway) are passed through the spread attribute in `SummaryCardSelect` (so they are already checking `...other`)

I'll add a test for a custom class 👍

jendowns:
Thank you @SimonFinney! I believe I have addressed your feedback so this is ready to review again 👍

src/components/SummaryCard/SummaryCardContainer/__tests__/SummaryCardContainer.spec.js
SimonFinney: Great point!

PR #346, opened 2020-02-04
fix(trending-card): reformat tests and remove snapshots

jendowns:
Followup to #309 & #312 

## Proposed changes

- reformat a11y tests to match what's done for `PanelV2` & current open PRs
- remove snapshot tests
- create new custom link test that checks for passed-through attributes and href & also checks that custom classes are consolidated/handled

NOTE: I stopped importing the `TrendingCard` from the story because that structure did not allow customizing props like `title` etc in the tests. So I think it's important that the tests be independent in this case. Plus I think it would be helpful that their details (prop values etc) aren't abstracted in other files.


src/components/TrendingCard/__tests__/TrendingCard.spec.js
```diff
@@ -3,36 +3,63 @@
...
-    expect(
-      render(<TrendingCard element={Link} />).container.firstChild
-    ).toMatchSnapshot();
-  });
+  test('should accept a custom link', () => {
+    const { getByText, queryByTestId } = render(
+      <TrendingCard
+        title="test title"
+        className="test-trending-card-class"
+        element={({ children, className, ...other }) => (
```
SimonFinney: Would you mind clarifying on where these props are originating from? If I understand correctly, these will currently resolve as `undefined`

src/components/TrendingCard/__tests__/TrendingCard.spec.js
```diff
@@ -3,36 +3,63 @@
...
+          <Link
```
SimonFinney: If possible, I wonder if here it would also be valuable to test that the `Link` component itself has been rendered.

src/components/TrendingCard/__tests__/TrendingCard.spec.js
jendowns: It doesn't resolve to `undefined`... this enables you to use a custom component & pass through relevant props. 

It renders like this if you run `debug()` in this test:

```jsx
<body>
  <div>
    <a
      class="bx--link test-custom-link-class security--trending-card test-trending-card-class"
      data-testid="test-data-id"
      href="#test-href"
    >
      <h3
        class="security--trending-card__title"
      >
        test title
      </h3>
    </a>
  </div>
</body>
```

This passes through the `TrendingCard` children (i.e., "test title"), and it allows you to combine class names of `TrendingCard` AND the custom `Link`.

I can write a test for just `element={Link}` but this was meant to test how the custom element and the `TrendingCard` can share/combine certain things, like class name.

src/components/TrendingCard/__tests__/TrendingCard.spec.js
jendowns: That seems more appropriate for the `Link` component? (What would you check -- the presencee of a class name? that definitely seems more `Link`-test specific). 

There are a couple checks in this block that include `closest('a')` to target the anchor element, coming from the `Link`, as well.

jendowns:
Hey @SimonFinney I updated this PR to check all props & also I answered your questions above 👍

src/components/TrendingCard/__tests__/TrendingCard.spec.js
SimonFinney: Thank for clarifying!

jendowns: Hey @SimonFinney I believe an `elementType` can be an actual React component: https://reactjs.org/docs/typechecking-with-proptypes.html (also `https://github.com/facebook/prop-types/pull/211`)

And in this case, it is ideal to check like this because it is mimicking how one would pass in a custom link component from like a routing library, for example. They may not do just  `element={Link}`--  there may be a number of extra props they may need to pass through in their routing component, which is what this test is mimicking.

Do you think something is missing here?

SimonFinney: My understanding here is that `elementType` expects the element / component type, for example `Link` and not a component to be rendered, like `<Link prop="prop" />`.

This can be a standard HTML tag or a React component, as the internal logic of the `TrendingCard` generates the correct component and the props from that (See `createElement` - https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/TrendingCard/index.js#L18).

This logic was replicated from how Carbon's `UIShell` handles its links - https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/UIShell/Link.js#L17

Let me know if you need further clarification!

jendowns: @SimonFinney I'm confused what your point is above? Let me put what I said in another way...

```jsx
// Defined in another file
const Link(() => <Link prop={prop} />);

// This test
<TrendingCard element={Link} />
```

is the same as 

```jsx
// This test
<TrendingCard element={({prop}) => <Link prop={prop} />
```

Here's another example in use, with `CustomLink`: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Button/Button-story.js#L118

Unless you are saying you want the test simplified? Please let me know.

SimonFinney: Exactly! Given the internal logic, we only need to ensure that the correct component is rendered

jendowns:
@SimonFinney I updated the `element` prop test.

jendowns:
Just a reminder @SimonFinney that I updated the `element` test here & I believe that was the last piece of feedback? Let me know - thanks! 👍

Issue #347, opened 2020-02-04
DataDecorator - renders empty primaryButton in footer


jendowns:
## Bug - 

**Expected behavior -**

When no `primaryButton` is provided to the `DataDecorator`, it should not render that button inside the connected `Panel`.

**Actual behavior -**

If you don't provide a `primaryButton`, the `DataDecorator`'s connected `Panel` will still render an empty button.

### Steps for reproducing

Don't pass any footer-related props to the `DataDecorator` -- no `renderFooter` or `primaryButton`. And then run the storybook & see there's an empty button in the panel's footer.

PR #348, opened 2020-02-04
fix(data-decorator): add unit and a11y tests, remove snapshots

jendowns:
Follow up to #347 (no longer blocked)

## Proposed changes

- `Decorator` + `DataDecorator` -- add a11y tests & update unit tests to remove snapshots, add user events testing


jendowns:
@SimonFinney This PR has been updated to include more tests for the core `Decorator`, to ensue that its various props work as expected. Props tests in `Decorator` should be reflected in `DataDecorator`, with the exception of certain `Panel`-related functionality...

Also worthy of note: while testing the accessibility of this component, I found a gap in the way the score is communicated. If a score is provided to the `Decorator`, there is no way to communicate that score (or the relevance/severity of it within the set thresholds) to someone using a screenreader. The severity colors of the icon are visual-only features.

So I added a `scoreDescription` prop (function). By default it can be used to take the `score` and last item in the provided `scoreThreshold` to add a `title` to the inner severity icon. 

I also made sure this prop, plus the `scoreThreshold`, are passed through in the `DataDecorator`.

Deriohig:
> @SimonFinney This PR has been updated to include more tests for the core `Decorator`, to ensue that its various props work as expected. Props tests in `Decorator` reflect the same functionality that's in `DataDecorator`, with the exception of certain `Panel`-related functionality...
> 
> Also worthy of note: while testing the accessibility of this component, I found a gap in the way the score is communicated. If a score is provided to the `Decorator`, there is no way to communicate that score (or the relevance/severity of it within the set thresholds) to someone using a screenreader. The severity colors of the icon are visual-only features.
> 
> So I added a `scoreDescription` prop (function). By default it can be used to take the `score` and last item in the provided `scoreThreshold` to add a `title` to the inner severity icon.
> 
> I also made sure this prop, plus the `scoreThreshold`, are passed through in the `DataDecorator`.

I think that title is only available to people with a mouse. Will need to add `aria-label` for screen reader support (or `...spread`). This could be a follow up PR imo, as the scope of this issue is testing.

jendowns:
Good point @Deriohig! In that case an `aria-label` would be more appropriate than `title`? 🤔

Deriohig:
> Good point @Deriohig! In that case an `aria-label` would be more appropriate than `title`? 🤔

I think they provide different value. 
Title: in-context tooltip with cursor
aria-label: screen reader support
However, with any browser supported API I would default to a `...rest` approach (except classnames since they may need concatenation with other classnames), as then you don't need to selectively support them if they pop up in the future. Again, this is out of scope for this issue but I can create a new one if it help?

jendowns:
Yeah @Deriohig feel free to open a new one! I wonder if it's worthwhile though to just add an `aria-label` alongside a `title`? 🤔 Was doing a bit of reading... the `aria-label` would cover screen readers and would take precedence over `title` when it comes to assistive tech.

And then the `title` could stay there to help with visual + cursor info? 🤔

PR #349, opened 2020-02-04
fix(data-decorator): do not render primary button if prop not provided

### LINKED ISSUE #347: DataDecorator - renders empty primaryButton in footer

jendowns:
## Affected issues

- Resolves #347 

## Proposed changes

- do not render the primary button in the footer if no `primaryButton` prop is provided

## Testing instructions

- best way to test this is in the storybook. I suggest removing the `renderFooter` prop in the story -- this will make it so there are no footer props for buttons (including no `primaryButton`). The reason this wasn't showing up in the storybook sooner is `renderFooter` as always superseding the empty primary button anyway, since the `primaryButton` prop is deprecated.


src/components/DataDecorator/__tests__/__snapshots__/DataDecorator.spec.js.snap
```diff
@@ -165,11 +165,6 @@ exports[`DataDecorator renders correctly 1`] = `
         "WIZARD_TERTIARY_SECONDARY_TEXT": "Information will not be saved upon exit",
       }
     }
-    primaryButton={
```
jendowns: Just wanted to point out that this has been in the snapshot test for a while now, but it was the a11y & tab tests in #348 that actually caught the issue.

To me this illustrates how insufficient the snapshot tests are, due to the way they work (i.e., there is not a lot of info here & it's easy not to understand this means there's an empty button) and also how ineffectively we use them. 😅

PR #351, opened 2020-02-05
fix(scroll-gradient): update linear-gradient logic to work in safari

### LINKED ISSUE #328: Scroll Gradient is wrong color in Safari
## Bug - 
When looking at the FilterPanel component when Safari, the gradients are grey, not white. 

**Expected behavior -**
Text should like like it's fading away' under the gradient in the Cool Gray 10 theme.
Working (Chrome)
image

**Actual behavior -**
Text appears behind grey gradient
image

Happens with horizontal scrolling also. 

### Steps for reproducing

1.  Open Safari on OSX (or iOS)
2.  Navigate to https://ibm-security.carbondesignsystem.com/?path=/story/components-scrollgradient--vertical-scrolling

### Affected browsers

[What's my browser?](http://www.whatsmyua.com) and [browserl.ist supported browsers](http://browserl.ist/?q=%3E+1%25%2C+not+IE+11)

- [ ] Safari 13.0.3 on Mac OSX 10

jendowns:
## Affected issues

- Resolves #328 

## Proposed changes

- add `hex-rgb` dependency to handle converting a HEX value to a valid RGB (whether the hex is 3 or 6 characters)
- update the linear-gradient in `ScrollGradient` to conform to the way Safari interprets "transparent" colors (explanation here: https://github.com/carbon-design-system/carbon/pull/4859#issue-351979766)
- update `color` prop desc to specify that we really want to see a valid HEX, RGB, or RGBA color value.
- update story to NOT use css properties, which would break this linear gradient

## Testing instructions

- open in storybook in Safari and update the scroll gradient colors


src/components/ScrollGradient/ScrollGradient.js
```diff
@@ -181,8 +182,36 @@ class ScrollGradient extends Component {
...
+      // Generate array from comma separated string:
+      const colorCodeArray = colorCode.split(',');
+
+      // Make sure code is the right length:
+      if (colorCodeArray.length === 3 || colorCodeArray.length === 4) {
+        rgbCode = `${colorCodeArray[0]}, ${colorCodeArray[1]}, ${
+          colorCodeArray[2]
+        }`;
+      } else {
+        console.warn(colorPropWarning);
```
jendowns: I'm doing a couple extra steps here because I thought it might be useful to be helpful & tell people if the RGB/RGBA is not valid 🤔 

So instead of just taking `colorCode` and using that as the `rgb` value, I processed it a little to verify that it has 3 or 4 parts (red, green, blue, and maybe alpha though alpha will be dropped).

Let me know if you disagree and think I should just run with `codeCode`, without checking its length.

src/components/ScrollGradient/ScrollGradient.js
```diff
@@ -181,8 +182,36 @@ class ScrollGradient extends Component {
...
```
jendowns: I'm doing a couple extra steps here because I thought it might be useful to be helpful & tell people if the RGB/RGBA is not valid (i.e., too long or too short) 🤔 

So instead of just taking `colorCode` and using that as the `rgb` value, I processed it a little to verify that it has 3 or 4 parts (red, green, blue, and maybe alpha though alpha will be dropped).

Let me know if you disagree and think I should just run with `codeCode`, without checking its length.

SimonFinney:
@jendowns Apologies, I'm only getting around to this after the discussion on Slack! I may be doing something dumb here, but upon switching the theme, the value is still leveraging the 'Gray 100' value:

ScrollGradient

Is this correct? I have a suspicion that the deploy preview may not be up-to-date.

jendowns:
@SimonFinney The deploy is correct, it's just that the use of custom properties is not adaptable enough. The custom property for the color is not being used directly in the component -- instead, the component is converting the custom property to a rgb value and inserting that into the style tag. This seems like a limitation of the custom property implementation & I'm not sure how to get around it.

jendowns:
@SimonFinney I know you were exploring how best to handle the custom properties for this PR so the storybook theming does not break -- do you have any updates on that exploration?

SimonFinney:
@jendowns Good question! I did some initial exploration in the `328/scroll-gradient` branch - it's incomplete, but feel free to check it out!

As I recall, I believe there might be some performance concerns with calculating these values on the fly but this would need to be validated.

jendowns:
Closing, as there is no path for this currently.
This PR was focused on getting the colors to work correctly for end-users in Safari. But with this fix, the themes swapped via CSS variables in storybook wont work. If the priority is the demo, then a different approach needs to be taken.

PR #355, opened 2020-02-07
fix(tearsheet): allow text reflow

### LINKED ISSUE #354: TearsheetSmall does not wrap a string that overflows
### Screenshots

bug

SimonFinney:
## Affected issues

Resolves #354 

## Proposed changes

Allows small tearsheet text to break


SimonFinney:
Tagging @i0r1 for consideration

PR #358, opened 2020-02-07
fix(error-page): update error page tests and remove snapshots

jendowns:
## Proposed changes

- add a11y test and test props
- remove snapshot tests
- apply correct `style` formatting for custom `backgroundImage`


src/components/ErrorPage/ErrorPage.js
```diff
@@ -62,14 +62,18 @@ const ErrorPage = ({
...
+    formattedBackgroundImage = errorIllustrations[statusCode]
       ? { ...errorIllustrations[statusCode] }
       : { ...errorIllustrations.default };
+  } else {
+    formattedBackgroundImage = { backgroundImage };
   }
 
   return (
-    <section className={classes} style={backgroundImage} {...other}>
+    <section className={classes} style={formattedBackgroundImage} {...other}>
```
jendowns: In the process of adding a test for `backgroundImage` prop, I realized that custom background images are not put in the correct `style` attribute format (i.e., inside an object).

PR #359, opened 2020-02-07
docs: start a doc on testing components

jendowns:
## Proposed changes

- Add a doc with some testing instructions


docs/testing/README.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+#### Test structure
+
+- Each significant variation of a component should have its own test block.
+- Use an **asynchronous** test block, and **await** each accessibility rule matcher.
+
+#### DAP requirements
+
+- DAP requires that components be rendered inside a valid landmark element, which is why the `container` is a `main` node. This extra step may not be required if the component is already wrapped in a `main` or another significant HTML landmark element.
```
SimonFinney: I'm curious about this part here - have we identified many components in which this step is not required?

It seems like as we're testing components in isolation, this may never be the case - what are your thoughts?

docs/testing/README.md
```diff
@@ -0,0 +1,124 @@
...
+- DAP requires a unique id per test within a given component. (Hence, "ComponentName" and "ComponentName with variation")
+
+## User events
```
SimonFinney: 👏Awesome! What other kinds of unit tests have you discovered that we need to consider documenting for component contributors?

docs/testing/README.md
jendowns: My suspicion is that the `Shell` is the biggest example. But either way I think it is worthwhile to keep defining a `main` wrapper... even if we had to explicitly do it for every component... because it forces the test writer to understand how/why/what DAP requires here. If we were to ever abstract that away, then we'd be basically ignoring the landmark rule & keeping it out of sight, out of mind.

docs/testing/README.md
jendowns: Keyboard events would be the other example here but personally I haven't had success using them. (Specifically I've wanted to simulate "Enter", "Spacebar", and arrow keys)

jendowns: We would also need to have a section on keyboard input (outside of "Enter", "Spacebar" and arrow keys used to traverse elements), which I think @diego-codes did for his `FilterSearch` tests.

SimonFinney: Good point! I think it'd be super helpful to show an example of testing consumer-facing props too, for example, `className` or `...other`

docs/testing/README.md
```diff
@@ -0,0 +1,124 @@
+# Testing
```
SimonFinney: Would you mind also linking out to this document from `CONTRIBUTING.md`?

docs/testing/README.md
```diff
@@ -0,0 +1,189 @@
...
+  // The close button:
+  expect(getByLabelText(/test close/i)).toHaveFocus();
+});
+```
+
+If the component is a single interactive element, like a single `<a>` or `<button>` then a tab cycle test may not be necessary. The `TrendingCard` is an example of a component that does not require a tab cycle test.
+
+## Props
+
+### Custom class names
```
SimonFinney: 👏 Thanks for adding this! I wonder if we should also include general guidance around unit testing props - Something along the lines of 'If a component exposes a prop for consumer input, that input should be validated via an individual unit test to ensure the expected behavior is outputted'

.github/CONTRIBUTING.md
```diff
@@ -42,6 +42,8 @@ Once it's done building, you can start editing source code or creating new compo
 
 ## 5. Test your JavaScript code
 
+Please review the [testing guidelines](./docs/testing) for details on how and what to test.
```
SimonFinney: 🚀

docs/testing/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+## What to test
+
+### Accessibility
+
+Automated accessibility tests are required for significant component variations and should use existing `toHaveNoAxeViolations()` and `toHaveNoDAPViolations()` Jest matchers.
+
+Use this structure for automated accessibility tests:
+
+```js
```
SimonFinney: I haven't been able to verify this, but using the `jsx` tag instead of `js` here may give us syntax highlighting for the React components

SimonFinney: As an update, I verified and we get extra syntax highlighting for React components with `jsx`

jendowns: While you may see like a word or two with an extra color in the short snippet below, the difference is very stark with the longer tab order test (and would be with similarly long snippets that are heavy on `expect`s):

### js highlighting (current)
Screen Shot 2020-03-04 at 3 49 47 PM

### jsx highlighting
Screen Shot 2020-03-04 at 3 50 10 PM

jendowns:
@SimonFinney Is this good as a first draft? Regarding the syntax highlighting, I think we should keep these as `js` highlighting, as they are js snippets taken from a js file. They may content some jsx but with some longer snippets, the jsx highlights can appear incorrect (see: https://github.com/carbon-design-system/ibm-security/pull/359#discussion_r387956805). I think it would be preferable to have a little fewer colors than have the wrong highlighting? 👍 Let me know & I can wrap this one up.

docs/testing/README.md
SimonFinney: Ah! I obviously didn't look very far 😆That's kind of odd though - it looks like highlighting gets busted by the time it hits that closing tag... Does wrapping the `(<Button>test footer button</Button>)` in parentheses do anything for it?

SimonFinney:
Almost! Let me know your thoughts on https://github.com/carbon-design-system/ibm-security/pull/359#discussion_r383185778

docs/testing/README.md
jendowns: Unfortunately not, from what I can tell. I'm just going to the files tab in this PR, click on "edit this file", editing it, and then looking at the markdown preview.

Screen Shot 2020-03-04 at 4 16 18 PM

docs/testing/README.md
jendowns: Would you like to take a stab at that in a follow-up PR?

docs/testing/README.md
SimonFinney: Thanks! Finally got it -

Documentation

jendowns: OH oops! Now I see what you meant in your previous comment, about the footer button. 😆 Updated now.

docs/testing/README.md
SimonFinney: Sure! Let me create a stub

jendowns: Awesome, thank you!

PR #360, opened 2020-02-07
docs: set escapeHTML to false for decorator and datadecorator storybook knobs

### LINKED ISSUE #356: Bug - Decorator story is displaying HTML entities instead of escaped characters
**Expected behavior -**
The `Decorator` should display the intended HTML entity character instead of its code.

**Actual behavior -**
The `Decorator` component is not displaying escaped characters properly. For example, if one provides:  
```
<Decorator value="\"C:\\Program Files\\Confer\\RepMgr.exe\"" />
``` 
as the value of a decorator, 
```
&quot;C:\\Program Files\\Confer\\RepMgr.exe&quot;
``` 
is displayed.

### Steps for reproducing

1.  Go to the decorator preview in Storybook
2.  Paste in `"C:\\Program Files\\Confer\\RepMgr.exe"` in the value knob for the preview
3.  Notice the odd formatting of the string.

#### Input
image

#### Output
image

jendowns:
## Affected issues

- Resolves #356 

## Proposed changes

- turn off `escapeHTML` option for `Decorator` and `DataDecorator` storybook knobs

### DIFF
--- a/src/components/DataDecorator/DataDecorator.stories.js
+++ b/src/components/DataDecorator/DataDecorator.stories.js
@@ -35,6 +35,12 @@ const storyProps = () => ({
   renderFooter: () => <Button size="large">Custom footer</Button>,
 });
 
-storiesOf(patterns('DataDecorator'), module).add('Default', () => (
-  <DataDecorator {...storyProps()} />
-));
+storiesOf(patterns('DataDecorator'), module).add(
+  'Default',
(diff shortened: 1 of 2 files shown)


diego-codes:
I didn't realize is was knobs that was causing the behavior. Since it is not related to the actual component, I don't think it's necessary to expose Storybook by not escaping HTML.

PR #361, opened 2020-02-10
fix(accordion): add alignment test, fix custom class addition

jendowns:
## Proposed changes

- add test to check for "start" alignment by default (opposite of Carbn's default)
- add test to check for custom class name passed through spread attribute

Also:
- fix custom class addition in `Accordion` 😳

### DIFF
--- a/src/components/Accordion/__tests__/Accordion.spec.js
+++ b/src/components/Accordion/__tests__/Accordion.spec.js
@@ -1,0 +1,40 @@
+/**
+ * @file Accordion tests.
+ * @copyright IBM Security 2020
+ */
+
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { carbonPrefix } from '../../../globals/namespace';
+
+import { Accordion, AccordionItem } from '../../..';
+
+describe('Accordion', () => {
+  test('should have icon aligned at start by default', () => {
+    const { getByLabelText, getByText } = render(
+      <Accordion>
+        <AccordionItem title="test title" iconDescription="test icon label">
+          test content
+        </AccordionItem>
+      </Accordion>
+    );
+    // Expect accordion to have correct alignment class:
+    expect(getByText(/test title/i).closest('ul')).toHaveClass(
+      `${carbonPrefix}accordion--start`
+    );
+    // Expect the icon to be first inside the accordion item heading button:
+    expect(getByText(/test title/i).closest('button').firstChild).toBe(
+      getByLabelText(/test icon label/i)
+    );
+  });
+
+  test('should pass custom class through spread attribute', () => {
+    const { getByText } = render(
+      <Accordion className="custom-class">
+        <AccordionItem title="test title">test content</AccordionItem>
+      </Accordion>
+    );
+    expect(getByText(/test title/i).closest('ul')).toHaveClass('custom-class');
+  });
+});
--- a/src/components/Accordion/__tests__/Accordion.spec.js
+++ b/src/components/Accordion/__tests__/Accordion.spec.js
@@ -19,10 +19,6 @@ describe('Accordion', () => {
         </AccordionItem>
       </Accordion>
     );
-    // Expect accordion to have correct alignment class:
-    expect(getByText(/test title/i).closest('ul')).toHaveClass(
-      `${carbonPrefix}accordion--start`
-    );
     // Expect the icon to be first inside the accordion item heading button:
     expect(getByText(/test title/i).closest('button').firstChild).toBe(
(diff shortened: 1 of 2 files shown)


src/components/Accordion/__tests__/Accordion.spec.js
```diff
@@ -0,0 +1,40 @@
...
+          test content
+        </AccordionItem>
+      </Accordion>
+    );
+    // Expect accordion to have correct alignment class:
+    expect(getByText(/test title/i).closest('ul')).toHaveClass(
+      `${carbonPrefix}accordion--start`
+    );
+    // Expect the icon to be first inside the accordion item heading button:
+    expect(getByText(/test title/i).closest('button').firstChild).toBe(
```
SimonFinney: The line's a bit blurred, but I wonder if it's sufficient to test for the `--start` class as that's what we set ourselves? It nearly seems like the assertion on line 27 could be a Carbon responsibility.

src/components/Accordion/__tests__/Accordion.spec.js
```diff
@@ -0,0 +1,40 @@
...
+      getByLabelText(/test icon label/i)
+    );
+  });
+
+  test('should pass custom class through spread attribute', () => {
```
SimonFinney: 👏

src/components/Accordion/__tests__/Accordion.spec.js
jendowns: About testing for class name:
I do have a test for the `--start` class on line 24, but this test is more in line with the spirit of the guiding principles of the testing library. To that end, I've been trying to test by what the user sees more than checking for "invisible" things like class names -- in this case, trying to check if the user sees the icon "first" in the order of the accordion heading. There might be a better way to check this, but I prefer something in this spirit, versus just checking for class name.

About testing for something Carbon likely already checks:
I don't believe they necessarily check for this particular use case... this is moreso testing our implement of the `Accordion`, which takes Carbon's component and sets the alignment to `start`. I think this test could be useful in catching situations where:
(1) we accidentally made a style change that affects the alignment
(2) we accidentally made a component implementation change that remove "start as default"
(3) something happens upstream from Carbon that changes the alignment/order that we are expecting to see

SimonFinney: I agree with your rationale, and I really like the idea that we have a 'hook' that allows us to detect if something is broken in Carbon when we upgrade. 🎉

As I'm typing this, I'm realizing the `--start` class could actually end up being noise for us to test for, more so than the second assertion I requested be removed - Hypothetically if Carbon updates that class name but the rendering remains the same, our unit tests fail, but that's actually a false positive... if they change the HTML structure of the component significantly, our unit tests will fail too. This may not necessarily mean that the component is broken nor do I know the exact likelihood of something like this actually happening, but it's something that's definitiely been on my mind. 😆

I'm of 2 minds here, I really love this user-facing philosophy and approach to testing and I think it's working great for our own pieces so far. The question I have is, where does our responsibility end? This is a somewhat unique challenge we face in situations like this because we extend Carbon's base components. I'm looking for ideas in which we can safely account for situations like the above.

I'm still feeling out and rationalizing what we should be testing for and what we should ignore, so I appreciate the discussion and patience while I probe, and we work through all this!

jendowns: Yeah good point -- I'll remove that class name check on line 24 👍 Regardless of what the class, we are more concerned with the order. The class could update and it shouldn't matter to us in this case.

About this question --
> The question I have is, where does our responsibility end?
Still trying to figure this out myself honestly 😆

SimonFinney:
My bad! One thing I realized, because we pass through `...other` props, we should probably test for that too.

src/components/Accordion/__tests__/Accordion.spec.js
jendowns: Just saw this comment https://github.com/carbon-design-system/ibm-security/pull/361#issuecomment-584843349 --

> My bad! One thing I realized, because we pass through ...other props, we should probably test for that too.

The test above `should pass custom class through spread attribute` checks `...other`. Since the custom class name isn't added to anymore / pull out, it's included in the spread attribute. That's why the test is worded kinda funny i.e., "pass custom class through spread attribute"

I guess technically it not required since the custom class is a Carbon thing, however writing the test is what made me realize that we weren't passing through custom classes. 😆

SimonFinney: Nice! Thanks, I just realised this too - I think this should cover it

PR #362, opened 2020-02-10
fix(button): update testing structure, props docs, and add a11y test

jendowns:
## Proposed changes

- add tests for functionality we've added to the `Button` -- loading state, ghost-danger, etc. 
- fixed props so that `kind` has the correct values in propTypes & doesn't generate prop errors when `kind="ghost-danger"`


jendowns:
@SimonFinney I believe this PR is ready to review 👍

jendowns:
@SimonFinney I can add a separate test for the `disabled` prop but we don't have access to any `size`-related classes, except for the  deprecated `largeText`.

`size` is just passed in, and we don't change any classes: https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/Button/Button.js#L48

For `kind`, I did test `ghost-danger`. Otherwise `kind` is just passed in & we don't have access to those classes: https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/Button/Button.js#L46

jendowns:
Hey @SimonFinney disregard my last comment -- I was able to add additional tests for the `disabled` state and the `kind` + `size` props in my last 2 commits. 👍 Could you take another look when you get a chance? Thank you!

src/components/Button/Button.js
```diff
@@ -8,13 +8,17 @@ import PropTypes from 'prop-types';
...
+import { ButtonKinds } from 'carbon-components-react/lib/prop-types/types';
 import InlineLoading from '../InlineLoading';
 
 import { getComponentNamespace } from '../../globals/namespace';
 import deprecatedProp from '../../globals/prop-types';
 
 export const namespace = getComponentNamespace('button');
 
+// Add our `ghost-danger` kind to the existing array of button kinds:
+ButtonKinds.push('ghost-danger');
```
SimonFinney: 👏

PR #364, opened 2020-02-10
fix(card): update card and skeleton tests, update default alt attr value

jendowns:
## Proposed changes

- update tests for `Card` and `CardSkeleton` with a11y tests, click event test, and checking for a custom class
- remove snapshots as they are no longer needed

ALSO:
- updated default value for `label` attribute, which becomes an `alt` attribute for the header image... since the value isn't required, it needs to default to an empty string.


jendowns:
@SimonFinney I updated this PR to include more prop tests, so it should be ready to review 👍

PR #365, opened 2020-02-10
fix(external-link): add a11y and attributes tests, remove snapshots

jendowns:
## Proposed changes

- add a11y test, add test for custom class and extra attributes
- remove snapshot tests, as they are no longer needed


jendowns:
Thanks @SimonFinney I just added specific tests for `children` and `href` props 👍

PR #368, opened 2020-02-11
fix(pill): update tests, remove old snapshots

jendowns:
## Proposed changes

- add a11y test, custom class name test, and test to make sure the `Pill` is non-interactive
- remove old snapshots


src/components/Pill/__tests__/Pill.spec.js
```diff
@@ -1,21 +1,51 @@
...
+import { render } from '@testing-library/react';
 import React from 'react';
+import userEvent from '@testing-library/user-event';
 
-import Pill from '../';
+import { Pill } from '../../..';
 
-describe('Pill', () => {
-  let pill;
+describe('PanelV2', () => {
```
SimonFinney:
```suggestion
describe('Pill', () => {
```

src/components/Pill/__tests__/Pill.spec.js
```diff
@@ -1,21 +1,51 @@
...
+    const main = document.createElement('main');
+    render(<Pill value="127.0.0.1" type="IP" />, {
+      // DAP requires a landmark '<main>' in the DOM:
+      container: document.body.appendChild(main),
+    });
+    await expect(document.body).toHaveNoAxeViolations();
+    await expect(document.body).toHaveNoDAPViolations('Pill');
+  });
+
+  test('should add a custom class', () => {
```
SimonFinney: 👏Good! I believe we'll also need tests that verify the `inline`, `type`, and `value` props

src/components/Pill/__tests__/Pill.spec.js
```diff
@@ -1,21 +1,51 @@
...
+        <Pill value="127.0.0.1" type="IP" />
+        <button>test button</button>
+      </div>
+    );
+
+    userEvent.tab();
+
+    // Expect only the extra test button to be interactive:
+    expect(getByText(/127.0.0.1/i).parentNode).not.toHaveFocus();
+    expect(getByText(/test button/i)).toHaveFocus();
```
SimonFinney: Great! I think we can end the assertion here 💯

src/components/Pill/__tests__/Pill.spec.js
jendowns: 🤔 `inline` & `type` don't change... seems more appropriate to check in the `Decorator` anyway?
Also, `value` is included in every text but I can make a specific one...

src/components/Pill/__tests__/Pill.spec.js
jendowns: Hmm I don't agree. How can we be sure that order isn't at play here somehow, unless we do a complete loop?

src/components/Pill/__tests__/Pill.spec.js
jendowns: Actually about `value`,  think of it this way:

I'm querying `getByText(/127.0.0.1/i)` multiple times. It wouldn't work if the `value` prop wasn't working as expected. 👍 In this way, checking what the user sees is already demonstrating the functionality/internals of the component.

jendowns:
@SimonFinney I believe I addressed your feedback above 👍

src/components/Pill/__tests__/Pill.spec.js
SimonFinney: Would you mind elaborating on what you mean by the order? Initially, this felt redundant due to the `button` being placed after the `Pill` in the render.

src/components/Pill/__tests__/Pill.spec.js
SimonFinney: The reason I'm calling these out is that they're props that can be provided and configured by consumers, but I feel like this again goes back to this grey area where we're still defining our boundaries when passing props down to a child component! 

If a consumer applies the `inline` prop at the `Pill` level, I believe that needs to be verified as passed-through correctly? I think it's safe to disregard something like the `noIcon` prop we apply to the `Decorator` from the `Pill` because it's not reliant on consumer logic.

I would be more comfortable if we had explicit tests for something like `value` because if for some reason we decide to change the `getByText` query in any of the other assertions, we're no longer ensuring the prop works as expected.

Let me know your thoughts on the above!

src/components/Pill/__tests__/Pill.spec.js
jendowns: Honestly not sure what I meant here. 👍 I'll remove it.

jendowns:
Thanks @SimonFinney I just applied your feedback above 👍

PR #369, opened 2020-02-12
 fix(loading-message): update tests, move spread attribute 

jendowns:
## Proposed changes

- update tests, remove old snapshot tests
- move the `...other` spread attribute to the inner `Loading` icon


src/components/Loading/LoadingMessage/LoadingMessage.js
```diff
@@ -16,14 +16,16 @@ const LoadingMessage = ({
   withOverlay,
   ...other
 }) => {
-  const loading = <Loading active={active} small={small} withOverlay={false} />;
+  const loading = (
+    <Loading active={active} small={small} withOverlay={false} {...other} />
```
jendowns: `...other` needs to be moved here to pass through `Loading` component related props, like `description`.

src/components/Loading/LoadingMessage/LoadingMessage.js
```diff
@@ -16,14 +16,16 @@ const LoadingMessage = ({
...
+  );
 
   const overlayClasses = classnames('bx--loading-overlay', className, {
     'bx--loading-overlay--stop': !active,
   });
 
   return withOverlay ? (
```
diego-codes: Does the `LoadingMessage` not render the `children` (the message) if the overlay is turned off? Shouldn't it always render the message?

src/components/Loading/LoadingMessage/__tests__/LoadingMessage.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      <LoadingMessage description="test description">
+        test message
+      </LoadingMessage>
+    );
+    // Expect the loading animation to appear first:
+    expect(container.firstChild).toBe(
+      getByLabelText(/test description/i).parentNode
+    );
+    // Expect the loading message to appear second/last:
+    expect(container.lastChild).toBe(getByText(/test message/i));
```
diego-codes: I'm not sure what the value of testing the order of the elements is. It essentially tests React's logic of making sure that elements are rendered in order.

I suggest focusing on the outputs that are different due to different props and/or props combinations because they test the logic on how you programmed the component. For this component, I would consider testing:
- The spread of the other props, which you already did 😄 
- The conditional overlay class name based on the `active` prop. I would test the output of setting the prop value to `true` and `false`.
- The conditional render based on the `withOverlay` prop. I would also test the output of setting the prop value to `true` and `false`.

src/components/Loading/LoadingMessage/LoadingMessage.js
Deriohig: out of scope for this PR, but I feel a loader would need to also support aria attributes for screen readers. This spread only supports `className` and `description`

src/components/Loading/LoadingMessage/__tests__/LoadingMessage.spec.js
Deriohig: agree, think Diego has this PR covered in terms of expected testing coverage.

src/components/Loading/LoadingMessage/LoadingMessage.js
jendowns: I'm not sure about this honestly 😅 I did not originally work on this component, and I don't recall what the requirements are/where. Do you mind if I capture this as a separate issue so I can double-check with some folks?

src/components/Loading/LoadingMessage/LoadingMessage.js
jendowns: @Deriohig The `...other` spread attribute in this case would cover everything you mentioned -- it's just passing through anything extra that's not explicitly called out. So it would pass through any number of aria attributes you gave it to the inner `div.bx--loading` element.

Deriohig: Apologies, you are correct. I was using the storybooks accepted props instead of the carbon-component-react source code as my reference for the `<Loading/>` component's available props, should have dug deeper. I wonder if the support of native attributes should be documented in the storybook? Might be useful to consider.

jendowns: No worries @Deriohig I think in general the storybook needs work. 😬 I'm going to open a separate issue to address this for sure.

jendowns:
Thanks for the feedback @diego-codes & @Deriohig! I have added some new props tests & also revised some existing tests.

Two things I want to capture separately from this PR --
* Address @diego-codes question about `children` being rendered without an overlay here: https://github.com/carbon-design-system/ibm-security/pull/369#discussion_r389143035 (I would need to double-check requirements here but my initial reaction is that yes, it `children` should be rendered regardless of whether there's an overlay)
* Improve the storybook, as it currently does not show the knobs at all for `LoadingMessage` & the documentation is unclear.

src/components/Loading/LoadingMessage/LoadingMessage.js
diego-codes: Yep! Here it is #444

jendowns: Thank you!

src/components/Loading/LoadingMessage/__tests__/LoadingMessage.spec.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  test('should show a custom loading message via `children`', () => {
+    const { queryByText } = render(
+      <LoadingMessage>test message</LoadingMessage>
+    );
+    expect(queryByText(/test message/i)).toBeVisible();
+  });
+
+  test('should apply a custom class', () => {
```
diego-codes: It looks like you have the same tests here: https://github.com/carbon-design-system/ibm-security/pull/369/files#diff-6bca8213755a4e7a871fd06f980454edR60-R65

I like this one better 😄

PR #370, opened 2020-02-12
fix(time-indicator): update tests and remove old snapshots

jendowns:
## Proposed changes

- update tests
- remove old snapshots


src/components/TimeIndicator/__tests__/TimeIndicator.spec.js
```diff
@@ -1,15 +1,33 @@
 /**
  * @file Time indicator tests.
- * @copyright IBM Security 2019
+ * @copyright IBM Security 2020
  */
 
 import { render } from '@testing-library/react';
 import React from 'react';
 
-import { Default as TimeIndicator } from '../TimeIndicator.stories';
```
SimonFinney: I think this would be a great opportunity for us to leverage Storybook's Component Story Format (https://medium.com/storybookjs/component-story-format-66f4c32366df#bc76) to maintain parity between our Storybook stories and our axe and DAP tests - it might make less sense for the traditional unit tests below as we need to validate input and output (like the custom class in this example), but for our a11y tests, we'll want to ensure that the rendered output of the examples we provide are standards-compliant.

It's less of an issue for a smaller component like this, but for components where we have 3 - 5 examples or variations, this drastically increases the overhead.

jendowns: I wrote about why I'm doing this here: https://github.com/carbon-design-system/ibm-security/pull/346#issue-370962016

>  I stopped importing the TrendingCard from the story because that structure did not allow customizing props like title etc in the tests. So I think it's important that the tests be independent in this case. Plus I think it would be helpful that their details (prop values etc) aren't abstracted in other files.

We need to be able to customize props so that each test render can include exactly what we need, per situation/test.

SimonFinney: I think it's less necessary here because all we supply is a `children` prop - I'll experiment with the `Shell` refactor to see if there's any value in leveraging CSF without adding too much abstraction.

jendowns: Yeah something else to consider about abstraction in general --

Test renders should only render what is necessary, and should not be abstracted in another file. This blog post talks about the perils of abstraction a bit & I like the explanation: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing (nesting is the main topic but IMO the abstraction being talked about here + the abstraction in nesting are similar in that they make tests harder to maintain)

SimonFinney: While I definitely agree about unnecessary abstraction for simple components, I'd still like to determine it's value with regards to a11y testing and leveraging the CSF format. I believe some further exploration is needed and I can take that on to better understand it.

Given the nature of how the a11y tests are currently structured (pass a rendered component in), it seems like it has the potential to be a low-cost approach to validating the markup of story examples we currently advertise.

PR #375, opened 2020-02-13
fix(filter-panel): add a11y tests to subcomponents, use test keyword

jendowns:
~Blocked by #376~ No longer blocked

## Proposed changes

- add a11y tests to `FilterPanel` & all subcomponents


src/components/FilterPanel/FilterPanelAccordion/__tests__/FilterPanelAccordion.spec.js
```diff
@@ -9,19 +9,29 @@ import { render } from '@testing-library/react';
 import FilterPanelAccordion from '../FilterPanelAccordion';
 
 describe('FilterPanelAccordion', () => {
-  it('renders with a title', () => {
+  test('should have no Axe or DAP violations', async () => {
+    const main = document.createElement('main');
```
diego-codes: Just a suggestion, it might be useful to create a [custom renderer](https://testing-library.com/docs/react-testing-library/setup#custom-render) for a11y tests that handle the creation of the main element and setting it as a container.

We do this in DE when we need to wrap our container with a Redux provider to work. I'm happy to share the example offline if you're curious.

src/components/FilterPanel/FilterPanelAccordion/__tests__/FilterPanelAccordion.spec.js
```diff
@@ -9,19 +9,29 @@ import { render } from '@testing-library/react';
...
+    render(<FilterPanelAccordion title="test accordion title" />, {
+      // DAP requires a landmark '<main>' in the DOM:
+      container: document.body.appendChild(main),
+    });
+    await expect(document.body).toHaveNoAxeViolations();
+    await expect(document.body).toHaveNoDAPViolations('FilterPanelAccordion');
```
diego-codes: Do all of Axe / DAP tests have the same `expect` statements? What if this whole test function is converted into a utility that reduces the code duplication? Something like:

```jsx
testA11y(<FilterPanelAccordion title="test accordion title" />);
```

It may not be the best idea if the test logic is different in all a11y tests, but I thought I'd make the suggestion.

src/components/FilterPanel/FilterPanelAccordion/__tests__/FilterPanelAccordion.spec.js
jendowns: Good suggestion! Two reasons I'm hesitant to do that...
1. There may be situations where this isn't necessary, like if we end up having a component with its own sufficient landmarks. (Haven't come across one yet. For some reason, I imagine the `Shell` would apply here?)
2. Less of a reason, but -- this seems like a good way to make people remember the landmark requirements. 😆

src/components/FilterPanel/FilterPanelAccordion/__tests__/FilterPanelAccordion.spec.js
jendowns: Yeah they all use expect statements. And the only significant detail/difference is that the DAP tests must always have their own unique id (so there can't be multiple `FilterPanelAccordion` tests -- this comes up because some complex components have multiple a11y test blocks. You'll see `Component` and `Component with variation` as ids)

But I imagine even with DAP's unique ID requirement, a utility could still be used. 🤔

jendowns:
Thank you @diego-codes! Those are good suggestions above & having written a number of repetitive a11y tests now, I feel like maybe now it's time to revisit how they are structured. A utility would definitely help with that. I added your 2 suggestions to this new issue here: https://github.com/carbon-design-system/ibm-security/issues/427 (feel free to edit that description if its missing any details etc)

PR #377, opened 2020-02-13
fix(filter-panel-accordion-item): reorder elems so li are children of ul elems

### LINKED ISSUE #376: FilterPanelAccordionItem - Axe violation due to improper nesting of div elems inside ul
**Expected behavior -**

`FilterPanelAccordionItem' should not throw Axe violations.

**Actual behavior -**

`FilterPanelAccordionItem` throws Axe violations due to improper nesting. There should not be `div` elements inside of a `ul`.

Here is the violation output:

```
Summary of all failing tests
 FAIL  components/FilterPanel/FilterPanelAccordionItem/__tests__/FilterPanelAccordionItem.spec.js
  ● FilterPanelAccordionItem › should have no Axe or DAP violations

    Rule violation: #list [serious]
    > Ensures that lists are structured correctly
    <ul> and <ol> must only directly contain <li>, <script> or <template> elements
    https://dequeuniversity.com/rules/axe/3.4/list?application=axeAPI
    ================================================================================
    Node:
    <ul class="security--filter-panel-accordion-item__list">


    Fix all of the following:
      List element has direct children that are not allowed inside <li> elements
    Node:
    <ul class="security--filter-panel-accordion-item__list">


    Fix all of the following:
      List element has direct children that are not allowed inside <li> elements
```

### Steps for reproducing

See #375 for more context

jendowns:
## Affected issues

- Resolves #376 

## Proposed changes

- reorder elements so that `ul`s are only direct parents of `li` elements (nothing else was reordered)

### DIFF
--- a/src/components/FilterPanel/FilterPanelAccordionItem/FilterPanelAccordiontItem.js
+++ b/src/components/FilterPanel/FilterPanelAccordionItem/FilterPanelAccordiontItem.js
@@ -85,37 +85,39 @@ const FilterPanelAccordionItem = ({
       open={open}
       {...other}
     >
-      <ul className={`${namespace}__list`}>
-        <ScrollGradient
-          scrollElementClassName={`${namespace}__scroller`}
-          getScrollElementRef={setListContainer}
-          color={scrollGradientColor}
+      <ScrollGradient
+        scrollElementClassName={`${namespace}__scroller`}
+        getScrollElementRef={setListContainer}
+        color={scrollGradientColor}
+      >
+        <div
+          role="presentation"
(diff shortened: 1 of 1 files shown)


jendowns:
@diego-codes I took a stab at resolving #376 -- so I was wondering if you could take a look at this when you get a chance?

In short, I ONLY moved the `ul` element that was wrapping outer `ScrollGradient`, and I wrapped both occurrances of `li` elements with it.

It looks okay when I tested it, but do you think I'm missing anything here? Also, I cannot manage to test the `ScrollGradient` properly 🤔 Do you have tips/advice for that?

PR #379, opened 2020-02-14
fix(ica): update tests for ica and skeleton variation

jendowns:
## Proposed changes

- update test format of `ICA` tests & add a11y test
- add a11y test for `ICASkeleton`
- remove old snapshots


src/components/ICA/ICASkeleton/__tests__/ICASkeleton.spec.js
```diff
@@ -4,10 +4,18 @@
...
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render } from '@testing-library/react';
+
 import ICASkeleton from '../ICASkeleton';
 
-describe('ICASkeleton', () => {
-  it('matches default snapshot', () =>
-    expect(shallow(<ICASkeleton />)).toMatchSnapshot());
+describe('ICA', () => {
```
SimonFinney:
```suggestion
describe('ICASkeleton', () => {
```

src/components/ICA/ICASkeleton/__tests__/ICASkeleton.spec.js
```diff
@@ -4,10 +4,18 @@
...
+  test('should have no Axe or DAP violations', async () => {
```
SimonFinney: 👏

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,82 @@
...
+      // DAP requires a landmark '<main>' in the DOM:
+      container: document.body.appendChild(main),
+    });
+    await expect(document.body).toHaveNoAxeViolations();
+    await expect(document.body).toHaveNoDAPViolations('ICA');
   });
 
-  it('renders', () => {
-    expect(ica).toMatchSnapshot();
+  it('should render en dash when `value` is `null`', () => {
```
SimonFinney: Reminder to update the `it` declarations to `test`

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,82 @@
...
+    expect(queryByText('1.0m')).toBeVisible();
   });
 
-  it('renders a large value', () => {
-    ica.setProps({ value: largeValue });
-
-    expect(ica).toMatchSnapshot();
+  it('should render a subset of values when `value` and `total` are provided', () => {
+    const { getByText } = render(<ICA label="test ICA" value={5} total={10} />);
+    expect(getByText('5').nextSibling).toBe(getByText(/10/i));
```
SimonFinney: I had a quick question on some of the following unit tests - is there an alternative way to query / verify these without relying heavily on the internal DOM structure (`nextSibling`, `parentNode.children`, etc.)? I worry that if this structure was to change but the logic remains the same, these tests would no longer pass.

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,82 @@
...
+      <ICA label="test ICA" value={10} percentage />
+    );
+    expect(queryByText('10').firstElementChild).toBe(queryByText('%'));
   });
 
-  it('renders a total when forced', () => {
-    ica.setProps({ forceShowTotal: true, total: value });
-
-    expect(ica).toMatchSnapshot();
+  it('should render a percentage', () => {
```
SimonFinney: - [ ] This appears to be a duplicate of the above

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,82 @@
...
   Locales.forEach(locale =>
-    it(`accepts '${locale}' locale`, () => {
-      ica.setProps({ locale });
-
-      expect(() => ica).not.toThrow();
+    test(`should accept '${locale}' locale`, () => {
+      const { container } = render(
+        <ICA label="test ICA" value={10} locale={locale} />
+      );
+      expect(() => container).not.toThrow();
```
SimonFinney: 👍

src/components/ICA/__tests__/ICA.spec.js
jendowns: 🤔 I understand your concern, but wanted to point out two things --

1. In this way, it's not so different from the snapshot tests, where if DOM structured changed, they had to be revisited + regenerated (and those tests sufficed for a time 🤷‍♀ )

2. Moreso to your point though -- I think the ICA is different in this respect. We are dealing with a component that displays fractions or percentages. So for example, if the next sibling of the numerator is NOT the denominator, we have a problem -- and that problem could mean that the component is unreadable for screen readers.

jendowns:
Thanks @SimonFinney I just applied your feedback 👍

src/components/ICA/__tests__/ICA.spec.js
SimonFinney: Sorry, I'm also trying to reinforce my understanding of React Testing Library here at the same time!

That makes sense to me - thanks for clarifying!

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,75 @@
...
 
-  it('renders a subset of values', () => {
-    ica.setProps({ total });
-
-    expect(ica).toMatchSnapshot();
+  test('should not render subset of values when `total` is close to `value`', () => {
+    const { queryByText } = render(
+      <ICA label="test ICA" value={1000000} total={999995} />
+    );
+    expect(queryByText('999995')).not.toBeInTheDocument();
```
SimonFinney: What's the difference between `toBeInTheDocument` and `toBeVisible` in this case?

If I understand correctly, `toBeVisible` is based off of style properties (https://github.com/testing-library/jest-dom#tobevisible), which is something the test runner wouldn't pick up here - let me know if I'm misunderstanding this!

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,75 @@
...
 
-  it('renders a total when forced', () => {
-    ica.setProps({ forceShowTotal: true, total: value });
-
-    expect(ica).toMatchSnapshot();
+  test('should render a percentage', () => {
+    const { queryByText } = render(
+      <ICA label="test ICA" value={10} percentage />
+    );
+    expect(queryByText('10').firstElementChild).toBe(queryByText('%'));
```
SimonFinney: Can we be consistent in our usage of `getByText` and `queryByText`? I don't mind either way as long as it's the same 🚀

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,75 @@
...
-  it('renders a percentage', () => {
-    ica.setProps({ percentage: true, total });
-
-    expect(ica).toMatchSnapshot();
+  test('should render `total` when forced', () => {
+    const { getByText } = render(
+      <ICA label="test ICA" value={10} total={10} forceShowTotal />
+    );
+    // Expect 3 ICA nodes: the label, value, and the total.
+    expect(getByText(/test ICA/i).parentNode.children.length).toBe(3);
```
SimonFinney: Is it sufficient to use solely `queryByText(' / 10')` here?

src/components/ICA/__tests__/ICA.spec.js
```diff
@@ -3,76 +3,75 @@
...
 
-  it('does not render a subset when the total is close to the provided value', () => {
-    ica.setProps({ total: largeValue - value, value: largeValue });
-    expect(ica).toMatchSnapshot();
+  test('should not render `total` if it is the same as `value`', () => {
+    const { getByText } = render(
+      <ICA label="test ICA" value={10} total={10} />
+    );
+    // Expect 2 ICA nodes: the label and the value.
+    expect(getByText(/test ICA/i).parentNode.children.length).toBe(2);
```
SimonFinney: Is it sufficient to use `queryByText(' / 10').not.toBeInTheDocument()` here?

src/components/ICA/__tests__/ICA.spec.js
jendowns: Hey there, I am consistent in my usage (`queryBy*` is to check for the existence of, because it may not return something). This appears to be an error and should be `getBy*`

src/components/ICA/__tests__/ICA.spec.js
jendowns: This would return `null` & not pass the test. There's weirdness with the new line so maybe a matching function could solve it. Just seems easier to check for the correct nodes.

src/components/ICA/__tests__/ICA.spec.js
jendowns: This would not produce the result you want, possibly because of the new line before/after the slash.

src/components/ICA/__tests__/ICA.spec.js
jendowns: Probably better in this case to run a `queryByAll` type of query and check for the number of results: https://testing-library.com/docs/dom-testing-library/api-queries#queryallby

src/components/ICA/__tests__/ICA.spec.js
jendowns: `toBeVisible` is an extra step taken -- something can be in the document but may not be visible.

That's why I used `not.toBeInTheDocument()` here. Just checking for visibility doesn't seem strict enough to me.

I agree with a comment @diego-codes made (I think in Slack) where he recommended using `toBeVisibile()` to check for the affirmative, and use `not.toBeInTheDocument()` to check for the negative. There can be exceptions but that makes sense to me.

src/components/ICA/__tests__/ICA.spec.js
jendowns: It's probably overkill though. 🤔  I have an update to remove a lot of this.

src/components/ICA/__tests__/ICA.spec.js
jendowns: Well correction -- with edits, it will now be (so `queryBy*` is appropriate):

```
expect(queryByText('%')).toBeVisible();
```

jendowns:
Thanks @SimonFinney I made another round of edits per your comments. 🎉  I don't believe there are any more tests that rely on nodes now -- I've replaced those with queries (specifically a couple `queryAllByText` lines)

src/components/ICA/__tests__/ICA.spec.js
diego-codes: As @jendowns said, `getBy*` selectors will fail the test if nothing is found. In tests where you are 100% sure that the element should be there, `getBy*` is preferred. On the other hand, when you need to test if something not rendered, `queryBy*` is ideal because the tests will not fail when the selector returns `null`.

jendowns: ☝️ MUCH better explanation! 😆 Thank you for those details!

PR #380, opened 2020-02-14
fix(portal): replace old snapshots with a11y & event bubbling tests

jendowns:
## Proposed changes

- remove snapshots
- add a11y tests for variation with and without overlay
- add `unmount` test 
- add tests for event bubbling

NOTE: I did have an `initialFocus` test but it wasn't working as expected. Plus that's moreso a feature of the `focus-trap` library? Anyway, I took that one out.


jendowns:
@diego-codes I added you as a reviewer here because I recall we had talked about how best to conduct some of this tests before I opened this PR. So I'd love your review if you have time today, or whenever. Thank you! 🙏

src/components/Portal/__tests__/Portal.spec.js
```diff
@@ -3,31 +3,152 @@
...
 describe('Portal', () => {
-  let portal;
-
-  beforeEach(() => {
-    portal = (
-      <Portal className={className} title={label}>
-        <div>
-          <button>TEST</button>
-        </div>
+  beforeEach(cleanup);
```
diego-codes: I'm pretty sure that `cleanup` is usually [not necessary for React](https://testing-library.com/docs/react-testing-library/api#cleanup). Is it required here?

jendowns: Thank you, that makes sense! 👍 I have remove that

PR #382, opened 2020-02-17
fix(interactive-tag): update tests and remove old snapshots

jendowns:
## Proposed changes

- update tests to check for a11y violations, custom classes, extra props, and onRemove
- remove old snapshosts


jendowns:
Thanks @SimonFinney! I just added tests for `children`, `isSelected`, and the `removable` props 👍

src/components/Tag/InteractiveTag/__tests__/InteractiveTag.spec.js
```diff
@@ -20,13 +21,34 @@ describe('InteractiveTag', () => {
     await expect(document.body).toHaveNoDAPViolations('InteractiveTag');
   });
 
+  test('should add close button with `aria-label` when `removable` is `true`', () => {
+    const { queryByLabelText } = render(
+      <InteractiveTag removeBtnLabel="test label" removable>
+        test tag
+      </InteractiveTag>
+    );
+    expect(queryByLabelText(/test label/i)).toBeVisible();
```
SimonFinney: I think I asked this before but I'm not sure we fully landed on a stance here - should we be using `toBeVisible` or `toBeInTheDocument` for situations like this, or can these be interchangeable?

src/components/Tag/InteractiveTag/__tests__/InteractiveTag.spec.js
```diff
@@ -20,13 +21,34 @@ describe('InteractiveTag', () => {
     await expect(document.body).toHaveNoDAPViolations('InteractiveTag');
   });
 
+  test('should add close button with `aria-label` when `removable` is `true`', () => {
```
SimonFinney: 👏

src/components/Tag/InteractiveTag/__tests__/InteractiveTag.spec.js
jendowns: Hey @SimonFinney, here's that conversation: https://github.com/carbon-design-system/ibm-security/pull/379#discussion_r382197316

To put it another way: 

- when checking for the _affirmative_ (i.e., in XYZ present?) it is better to check for `toBeVisible` because you are checking whether it is in the DOM *and* whether it is visible to the user.

- when checking for the *negative*, it is better to check `toBeInTheDocument` because that check is all that matters. Something wouldn't be visible but not the DOM.

jendowns:
@SimonFinney Is this PR good to go?

src/components/Tag/InteractiveTag/__tests__/InteractiveTag.spec.js
SimonFinney: Thanks for clarifying - that wasn't immediately clear to me from the documentation https://github.com/testing-library/jest-dom#tobevisible

PR #383, opened 2020-02-17
fix(type-layout): update tests and remove old snapshots

jendowns:
## Proposed changes

- update tests to check for a11y violations, custom classes, and extra props for all subcomponents
- remove old snapshots


src/components/TypeLayout/__tests__/TypeLayout.spec.js
```diff
@@ -3,53 +3,80 @@
...
-                  <ul>
-                    <li>{description}</li>
-                    <li>{description}</li>
-                    <li>{description}</li>
-                  </ul>
-                </TypeLayoutCell>
-              </TypeLayoutRow>
-            );
-          })}
+  test('should add a custom class to each component', () => {
```
SimonFinney: 👏 Awesome!

src/components/TypeLayout/__tests__/TypeLayout.spec.js
```diff
@@ -3,53 +3,80 @@
...
-
-    it("renders the HTML of the node's subtree", () => {
-      expect(typeLayout.render()).toMatchSnapshot();
-    });
-
-    it('renders correct classes if `size` or `bordered` are set', () => {
-      typeLayout.setProps({ size: 'sm', border: true });
-      expect(typeLayout.render()).toMatchSnapshot();
-    });
+  test('should pass through extra props via spread attribute', () => {
```
SimonFinney: This is a reminder to also include tests for `border`, `children`, and `size`

jendowns: `border` is a Carbon `StructuredList` prop.
All of those props, particularly `children`, are demonstrated in the Storybook 🤔

SimonFinney: On initial view `border` and `size` look like a classes we add ourselves based on the prop, so my assumption is we'd want to cover that.

Also, so I fully understand, do you suggest we should avoid unit testing these props because they're covered by the examples in Storybook?

jendowns: Well so yes and no. 🤔 The snapshots were not helpful in this regard (they aren't checked visually for any meaning; they are just updated when they need to be updated). I can check for the presence of class names but that doesn't seem in the right spirit & the test itself seems less valuable than seeing what is rendered in storybook.

jendowns:
@SimonFinney I just added tests for the `children`, `border`, and `size` props 👍

PR #387, opened 2020-02-19
fix(status-icon): add a11y tests, update test structure, remove old snapshots

jendowns:
## Proposed changes

- add a11y tests per `status` (including when `status={undefined}`)
- add tests for sizes, message, and custom class name
- remove old snapshots


jendowns:
This component has been updated 👍

src/components/StatusIcon/__tests__/StatusIcon.spec.js
```diff
@@ -3,35 +3,70 @@
...
-describe('StatusIcon', () => {
-  let statusIcon;
+import { carbonPrefix } from '../../../globals/namespace';
 
-  beforeEach(() => {
-    statusIcon = shallow(<StatusIcon />);
-  });
+describe('StatusIcon', () => {
+  // Should check for `undefined` status as well:
+  STATUS.push(undefined);
```
diego-codes: Instead of pushing to the status, which could permanently change the array for other tests, I would create a new array that combines the `STATUS` array and `undefined`.

diego-codes: And actually, now that I'm seeing the rest of the tests. I think it's easier to just have special single tests for the undefined status.

jendowns: Thanks @diego-codes that's a good point! I just updated the tests to avoid pushing the array & have separate tests for `status={undefined}` 👍

PR #388, opened 2020-02-20
fix(status-indicator): add a11y tests, update testing structure, apply minor fixes

jendowns:
### `StatusStep`
- add a11y tests
- update testing structure
- ensure that no empty `<p>` elem is applied when `description` isn't provided
- update prop description for `label` to better reflect what it does

### `StatusIndicator`
- add a11y tests
- update testing structure
- deprecate the `description` prop (which isn't used anywhere? 🤔 )

### NOTE ⚠️ 
I think there needs to be some work done for `currentIndex` prop, which doesn't appear to work as expected. It needs to be fixed/addressed separately.


src/components/StatusIndicator/StatusStep/StatusStep.js
```diff
@@ -80,7 +82,7 @@ const getStatusIcon = (status, description) => {
  */
 class StatusStep extends Component {
   static propTypes = {
-    /** @type {string} A label for a step. */
+    /** @type {string} Used to create a custom key for a step. */
     label: string.isRequired,
```
jendowns: This prop is used to apply extra characters to a `key` 🤔 I think it needs to be re-evaluated... but for now I can update the prop desc.

src/components/StatusIndicator/StatusStep/StatusStep.js
```diff
@@ -54,7 +54,9 @@ const getStatusIcon = (status, description) => {
               title={description}
             />
           </div>
-          <p className={`${namespace}__label`}>{description}</p>
+          {description && (
+            <p className={`${namespace}__label`}>{description}</p>
+          )}
```
jendowns: `description` doesn't appear to be used anywhere?? 😆

src/components/StatusIndicator/StatusStep/__tests__/StatusStep.spec.js
```diff
@@ -0,0 +1,54 @@
...
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { StatusStep } from '../../../..';
+
+import { STATUS } from '../StatusStep';
+
+describe('StatusStep', () => {
+  // eslint-disable-next-line compat/compat
+  Object.values(STATUS).forEach(status =>
```
jendowns: This is because of Opera Mini support 🤔

src/components/StatusIndicator/StatusIndicator.js
```diff
@@ -80,15 +83,17 @@ class StatusIndicator extends Component {
...
-            <Button
-              className={`${namespace}__button--retry`}
-              kind="ghost"
-              onClick={this.state.retry.action}
-              onKeyPress={this.state.retry.action}
-              renderIcon={Restart20}
-            >
-              {this.state.retry.label}
-            </Button>
+            <li>
```
jendowns: Needs to be wrapped in an `li` because it is a child of a `ul`

src/components/__tests__/__snapshots__/publicAPI.spec.js.snap
```diff
@@ -7723,9 +7723,7 @@ Map {
             "action": Object {
               "type": "func",
             },
-            "description": Object {
-              "type": "string",
-            },
+            "description": [Function],
```
jendowns: FYI this is happening because this prop is now wrapped in a `deprecate` function. I think it is unavoidable.

SimonFinney:
Thanks @jendowns ! I'll aim to get a code review finalised today - one thing I've noticed is that the type size appears to have shrunk for the completed steps, but I believe this problem also exists in `dev` so I'll create a stub for that:

StatusIndicator

src/components/StatusIndicator/StatusIndicator.js
```diff
@@ -47,7 +47,10 @@ class StatusIndicator extends Component {
       label: string,
 
       /** @type {string} The description for the retry button. */
-      description: string,
+      description: deprecate(
```
SimonFinney: 👏

src/components/StatusIndicator/StatusStep/__tests__/StatusStep.spec.js
SimonFinney: As a fallback, I believe we can do the following - 

```suggestion
  Object.keys(STATUS).map(status => STATUS[status]).forEach(status =>
```

jendowns: Sure, but the opera mini issue comes up elsewhere as well:
Screen Shot 2020-03-23 at 10 20 37 AM

I suggest we update the compat rules instead. 🤔

jendowns:
@SimonFinney I responded to your comment above about the linting rule. We are using `Object.values` elsewhere. Also about `currentIndex`, I mentioned in the desc that it isn't working/testing as expected & likely a bug needs to be fixed separately.

Issue #389, opened 2020-02-20
Upgrade to Carbon 10.10.0 release


jendowns:
## Requirements

- [ ] Upgrade to [Carbon 10.10.0](https://github.com/carbon-design-system/carbon/pull/5236), which is scheduled to be released on [March 3, 2020](https://github.com/carbon-design-system/carbon/wiki/Release-radar#schedule)
- [ ] Confirm that v10.10.0 changes are non-breaking for our codebase

PR #390, opened 2020-02-20
fix(step-indicator): update tests, remove old snapshots

jendowns:
## Proposed changes

- add a11y test
- test for custom class name
- remove old snapshots

Also:
- remove unnecessary code that sets unused props to `null`

~**⚠️ NOTE: BLOCKED BY https://github.com/carbon-design-system/ibm-security/issues/389**~
~That release includes a critical a11y fix for the `ProgressIndicator`, which is used in this `StepIndicator`~


src/components/StepIndicator/StepIndicator.js
```diff
@@ -19,12 +19,12 @@ const Step = props => {
...
-    description: props.description || null,
-    disabled: props.disabled || null,
-    invalid: props.invalid || null,
-    label: props.label, // required
+    current: props.current,
+    complete: props.complete,
+    description: props.description,
+    disabled: props.disabled,
+    invalid: props.invalid,
+    label: props.label,
```
jendowns: setting these to `null` is not necessary because there are default props already

src/components/StepIndicator/__tests__/StepIndicator.spec.js
```diff
@@ -4,40 +4,45 @@
...
+    );
+    expect(getByText(/test label/i).closest('li')).toHaveClass(
+      'custom-step-class'
+    );
+  });
 
-    it('renders if `currentIndex` is null', () => {
-      list.setProps({ currentIndex: null });
-      expect(list.render()).toMatchSnapshot();
-    });
```
jendowns: I don't think this test is necessary, as it is built-in to the underlying `ProgressIndicator`

src/components/StepIndicator/__tests__/StepIndicator.spec.js
```diff
@@ -4,40 +4,54 @@
  */
 
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render } from '@testing-library/react';
 
-import { Step, StepIndicator } from '../';
+import { Step, StepIndicator } from '../../..';
```
SimonFinney: It looks like we ultimately spread the `Step` props into `ProgressStep`, but I believe we'll need to account for that logic, as well as potentially the `renderLabel` prop

jendowns: Thank you -- just added a new file for `Step` tests in particular 👍

jendowns:
@SimonFinney This PR has been updated & is ready to review again 👍

PR #395, opened 2020-02-21
fix(button): ensure loading icon does not make button taller

### LINKED ISSUE #391: Button height changes when toggling loading prop
## Bug

When the loading prop is used, the height of the button changes from 48px to 60px. This can cause items on a page to move around.

**Expected behavior -**

The `loading` prop, whether set to true or false should not change the height of the button.

**Actual behavior -**

When loading it set to `true`, the height of the button changes.

### Steps for reproducing

1.  Vist [Storybook](https://ibm-security.carbondesignsystem.com/?path=/story/components-button--default)
2.  Inspect button height (48px)
3.  Set `loading` knob to `true`
4. Inspect button height (60px)

### Screenshots

image

image

### Affected browsers

All browsers

jendowns:
## Affected issues

- Resolves #391 

## Proposed changes

- add negative top/button margins to the loading animation icon (I think this is the simplest way to handle the issue where the loading icon makes the button taller -- because the icon itself is so tall 🤔 I experimented with absolute positioning, setting strict heights, etc., but this works cleanly in the latest Firefox, Safari, and Chrome. Definitely let me know what you think though!)

## Testing instructions

- toggle the `Loading` knob in storybook and then check the button heights for default, small, field, and large. The `Loading` variation should be the same height as the non-loading variations.


SimonFinney:
I've tested this at all sizes and it LGTM! 👏 Adding @carbon-design-system/ibm-security-design for verification

PR #396, opened 2020-02-21
feat(details-page-1): add grid template with nested column styles

jendowns:
## ⚠️ NOTES! ⚠️ 
* This component and all subcomponents have *terrible* names. I realize they are "template 1, template 2, etc" in the originating issues... buuuuut any naming suggestions? 😆 Something shorter and without a number would be heavily preferred.
* ~This is **not ready to merge until Carbon officially releases `10.10.0`** around March 3 (right now it is just a release candidate)~

## Proposed changes

- ~adds Carbon `10.10.0` release candidate to use `Grid` components~
- add `DetailsPage1` ("Details page template 1) with story
- add styles for "nested" columns that applied as subcomponents

To break down the template...

```jsx
<DetailsPage1>

    // START ROW!
    // 6 out of 16 columns
    <DetailsPage1Section>
        // 2 out of 6 columns
        <DetailsPage1SectionTerm />
        // 4 out of 6 columns
        <DetailsPage1SectionDefinition />
    </DetailsPage1Section>

    // 6 out of 16 columns
    <DetailsPage1Section>
        // 2 out of 6 columns
        <DetailsPage1SectionTerm />
        // 4 out of 6 columns
        <DetailsPage1SectionDefinition />
    </DetailsPage1Section>

    // 4 out of 16 columns
    <DetailsPage1Aside />

    // NEW ROW!
    // 16 out of 16 columns
    <DetailsPage1Main />

</DetailsPage1>
```

## Testing instructions

- see `DetailsPage1` story in the branch deploy


src/components/GridTemplates/DetailsPage1/DetailsPage1.js
```diff
@@ -0,0 +1,147 @@
+/**
+ * @file Details page template #1.
```
jendowns: Quick note -- I threw allllll of the subcomponents in the same file for this initial review. But let me know if it is too much to look at, and I can split them up. I do plan to split them out eventually, but I didn't want to do that right now while the naming structure is still up in the air.

package.json
```diff
@@ -44,16 +44,16 @@
...
-    "@carbon/themes": "10.9.3",
-    "@carbon/type": "10.8.3",
-    "carbon-components": "10.9.3",
-    "carbon-components-react": "7.9.3",
+    "@carbon/layout": "10.8.0-rc.0",
+    "@carbon/motion": "10.6.0-rc.0",
+    "@carbon/themes": "10.10.0-rc.0",
+    "@carbon/type": "10.9.0-rc.0",
+    "carbon-components": "10.10.0-rc.0",
+    "carbon-components-react": "7.10.0-rc.0",
```
jendowns: Because of these updates, a number of snapshots need to be regenerated. 🙃 
But to keep the subset of changes here low, I'd like to avoid updating snapshots in this branch right now. Let me know if you have concerns about this though. 👍

src/components/GridTemplates/DetailsPage1/DetailsPage1.js
```diff
@@ -0,0 +1,147 @@
...
+  children: undefined,
+  className: undefined,
+};
+
+DetailsPage1Section.defaultProps = {
+  children: node,
+  className: string,
+};
+
+function DetailsPage1SectionTerm({ children, className, ...other }) {
```
jendowns: This name is... so much 😆 
I was thinking of "term" and "definition" for this plus `DetailsPage1SectionDefinition`, since they are side-by-side nested columns. But again I am VERY open to changing these names.

src/components/GridTemplates/DetailsPage1/_index.scss
```diff
@@ -0,0 +1,35 @@
...
+      }
+    }
+
+    // Imitating styles for a column that spans 4 (out of 6)
+    // & also has typical padding:
+    .#{$namespace}details-page-1__section__definition {
+      flex: 0 0 100%;
+      max-width: calc(66.66% - 1px);
+      padding: 0 $carbon--spacing-05;
+    }
```
jendowns: I took column styles & repurposed them for this situation. Unfortunately I don't think there are methods we can use for this type of thing? 🤔 (everything I saw was set to `@private` scope)

src/components/GridTemplates/DetailsPage1/DetailsPage1.js
SimonFinney: That's exactly what I did for the `Shell` spike, so I completely understand!

SimonFinney:
@jendowns Thanks for putting this together so quickly - I looked at the deploy briefly, but so far, so good!

@cameroncalder should be able to provide insight into a more suitable naming convention. I completely agree with your concern here! Numbering these won't scale them very well, and it provides very little guidance into what the intent of each of these are.

jendowns:
Thanks @SimonFinney! So just double-checking -- aside from naming (and the need to split out into diff files), it looks good?

src/index.js
```diff
@@ -227,3 +227,12 @@ export UnorderedList from 'carbon-components-react/lib/components/UnorderedList'
 export UNSTABLE__Pagination, {
   PageSelector,
 } from './components/UNSTABLE__Pagination';
+
+export {
```
SimonFinney: To be safe, how do you feel about marking these as unstable until we work out the correct naming?

package.json
SimonFinney: Good idea! Let's update these towards the end to mitigate noise

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
```diff
@@ -0,0 +1,73 @@
...
+        minHeight: '240px',
+        width: '100%',
+      }}
+    >
+      lorem ipsum
+    </div>
+  );
+}
+
+disableCentered(storiesOf(patterns('DetailsPage1'), module)).add(
```
SimonFinney: Would you mind using [Component Story Format](https://storybook.js.org/docs/formats/component-story-format) here? I believe the `storiesOf` format will be deprecated in the future.

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @file Details page template #1 story.
+ * @copyright IBM Security 2020
+ */
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import { disableCentered, patterns } from '../../../../.storybook';
```
SimonFinney: For organisational purposes, it might be appropriate to create a new Storybook category specific for 'Layout' - what are your thoughts?

src/components/GridTemplates/index.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @file Grid templates entry point.
```
SimonFinney: This goes back to naming, but I wonder if the term `LayoutTemplates` would be more suitable than `GridTemplates`, as the grid is the underlying technology, but the layout is the intended solution.

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
```diff
@@ -0,0 +1,73 @@
...
+function DemoContent() {
+  return (
+    <div
+      style={{
+        backgroundColor: theme.interactive04,
+        minHeight: '240px',
+        width: '100%',
+      }}
+    >
+      lorem ipsum
```
SimonFinney: @cameroncalder should be able to provide some content that would be more descriptive for this example.

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
```diff
@@ -0,0 +1,73 @@
...
+  DetailsPage1SectionDefinition,
+  DetailsPage1Aside,
+} from '../../..';
+
+function DemoContent() {
+  return (
+    <div
+      style={{
+        backgroundColor: theme.interactive04,
+        minHeight: '240px',
```
SimonFinney: Let's use `@carbon/layout` for any of these pixel values where possible

src/components/GridTemplates/DetailsPage1/_index.scss
```diff
@@ -0,0 +1,35 @@
+////
+/// Details page template 1 styles.
+/// @group accordion
```
SimonFinney:
```suggestion
/// @group details-page-1
```

src/components/GridTemplates/DetailsPage1/DetailsPage1.js
```diff
@@ -0,0 +1,147 @@
...
+function DetailsPage1({ children, className, ...other }) {
+  const classes = classnames(namespace, className);
+  return (
+    <Grid fullWidth condensed className={classes} {...other}>
+      <Row>{children}</Row>
+    </Grid>
+  );
+}
+
+DetailsPage1.propTypes = {
```
SimonFinney: - [ ] It looks like the `defaultProps` and `propTypes` definitions are reversed throughout these components

src/components/GridTemplates/DetailsPage1/DetailsPage1.js
SimonFinney: Good point! This is definitely dependent on the kind of content @cameroncalder expects will most likely be contained in these columns

src/components/GridTemplates/DetailsPage1/_index.scss
```diff
@@ -0,0 +1,35 @@
...
+/// @copyright IBM Security 2018
+////
+
+@import '../../../globals/namespace/index';
+@import '../../../globals/layout/index';
+
+@include export-namespace($name: 'details-page-1') {
```
SimonFinney: What's the rationale behind using `export-namespace($name: details-page-1)` over `@include security--component($name: details-page-1)` here?

src/components/GridTemplates/DetailsPage1/_index.scss
```diff
@@ -0,0 +1,35 @@
...
+  // Note that this column spans 6:
+  .#{$namespace}details-page-1__section {
+    display: flex;
+
+    // Imitating styles for a column that spans 2 (out of 6)
+    // & also has typical padding:
+    .#{$namespace}details-page-1__section__term {
+      flex: 0 0 100%;
```
SimonFinney: 👏 I remember you mentioning we could pull this out into a helper or `mixin` to determine and provided these values - have you given this further thought?

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: We don't really have descriptive content anywhere in these components (just like "example" and "label"), but I can at least change these to just "content".

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: For demo content? We have a lot of story-specific style that do this, and Carbon does as well. Wouldn't that be overkill? It's kinda nice for a story to use the pixel value so that an end user can view this in the story source tab and immediately understand what the value is, without specialized knowledge of a Carbon-specific variable.

src/components/GridTemplates/index.js
jendowns: This has been updated 👍

src/components/GridTemplates/DetailsPage1/_index.scss
jendowns: Updated 👍

src/components/GridTemplates/DetailsPage1/_index.scss
jendowns: If I remember correctly, I just copied this scss file from the `Accordion` to use as a template, so that's why it has `export-namespace` & the wrong namespace in the top-of-file comments. 

Fixed now 👍

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
SimonFinney: You're right, we don't have much to work with right now! I'm envisioning a later story for these examples would be to leverage some of the components available to provide examples with additional context

jendowns:
Blocked by #417

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: Honestly I'm not a fan of another category. Already, the line between "patterns" and "components" is pretty blurry. So then how confusing would "patterns" versus "layouts" be? And these are all components, technically. 😬 I'm almost of the mind that we should remove the categories entirely. They don't serve much of a purpose for someone using the sidebar (just splitting content, taking things out of order, creating categories that could arguably mean multiple things, and apply to the same components in different ways, etc.)

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: Just revisiting this -- I still don't think it makes sense to use variables for mocked up / demo content.

src/components/GridTemplates/DetailsPage1/_index.scss
jendowns: I think that effort should be separate. It would require some testing across different layouts.

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: Where is this being used elsewhere in the codebase?

jendowns:
@SimonFinney this has been updated & is now ready to be reviewed again 👍 
With these changes, I believe this is now merge-able, once it passes review (as opposed to before when it was blocked and waiting on other updates).

src/components/LayoutTemplates/UNSTABLE__DetailsPage1/UNSTABLE__DetailsPage1.js
```diff
@@ -0,0 +1,35 @@
...
+  const classes = classnames(namespace, className);
+  return (
+    <Grid fullWidth condensed className={classes} {...other}>
+      <Row>{children}</Row>
+    </Grid>
+  );
+}
+
+UNSTABLE__DetailsPage1.propTypes = {
+  children: node,
```
SimonFinney: Would you mind documenting these props for them to display in the generated Storybook 'Show Info' modal?

src/components/LayoutTemplates/UNSTABLE__DetailsPage1/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file Details page template 1 entry point.
+ * @copyright IBM Security 2020
+ */
+
+import UNSTABLE__DetailsPage1 from './UNSTABLE__DetailsPage1';
```
SimonFinney: For consistency, would you mind following the same import syntax used in `LayoutTemplates/index` ?

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
SimonFinney: Fair enough! Let's reassess the information architecture of these at a later stage

src/components/GridTemplates/DetailsPage1/_index.scss
SimonFinney: Fair point!

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
SimonFinney: This is currently in place for - 

- [`TimeIndicator`](https://github.com/carbon-design-system/ibm-security/blob/master/src/components/TimeIndicator/TimeIndicator.stories.js)
- [`TrendingCard`](https://github.com/carbon-design-system/ibm-security/blob/master/src/components/TrendingCard/TrendingCard.stories.js)
- [`UNSTABLE__Shell`](https://github.com/carbon-design-system/ibm-security/blob/next/src/components/UNSTABLE__Shell/Shell.stories.js)

As per the [Storybook team](https://storybook.js.org/docs/formats/component-story-format/), this is now the recommended way of writing stories - in my opinion, we should be writing new stories in this format moving forward, to avoid a larger migration effort in future.

src/components/LayoutTemplates/UNSTABLE__DetailsPage1/index.js
jendowns: So speaking of this, was thinking it would be better to start doing `export { default as ComponentName } from 'path/to/component'`. Would clear syntax errors, follow Carbon's format, and also address this comment: https://github.com/carbon-design-system/ibm-security/pull/315#discussion_r367624887

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: Yeah I'm open to two categories but the differences currently seem super subtle 👍 Happy to reevaluate in the future of course

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: Hey @SimonFinney I'm seeing the reaction above so wanted to ask about it. Not sure what variable you want to see that comes out to 240px: https://www.carbondesignsystem.com/guidelines/spacing

And there are lots of other places in the storybook where we have demo pixel values step, to mock something up for the user. 🤔 Quick search in codebase's stories will get you these examples: `Accordion`, `CodeSnippet`, `Dropdown`, `FileUploader`, `FilterPanel`, `IconButtonBar`, `InlineLoading`, `MultiSelect`, `Nav`, `Notification`, `OverflowMenu`, `Pagination`, `PanelV2`, `Pill`, `ScrollGradient`, `Search`, `Select`, `Shell`, `SkeletonText`, `StructuredList`, `Tabs`, `Tile`, `UNSTABLE__Pagination`, and `Wizard`. I might have even missed one when copying the list, there are so many.

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
jendowns: @SimonFinney how should `disableCentered` be applied in this branch?

SimonFinney: Good question! I believe we have support for this in `next` - we may need to bring the change over

jendowns: So in that case, I wouldn't be able to use the component story format without an interesting workaround unfortunately. This component is full-width and not centered.

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
SimonFinney: No worries! If there isn't something readily-available from Carbon that makes sense, I'm happy to leave it as a pixel value - to your point, it seems like overkill unless there's something we can leverage for consistency

src/components/GridTemplates/DetailsPage1/DetailsPage1.stories.js
SimonFinney: Let's leave it for now... I'll create a stub to revisit and bring in the configuration from `next`

jendowns:
Per conversations with Cameron & Jimmy earlier, I'm closing this out. Instead, the direction will be to add layout patterns as storybook demos featuring the `Grid`, `Row`, and `Column` components as-is, not abstracted.

SimonFinney:
😮 Is this correct? My understanding is we'd still need the abstraction for the templates, but the components that fall into it would be part of the demo

jendowns:
Yes! They want a level of flexibility that cannot be achieved by creating abstracted layout components. Their main concern with regards to uniformity was more about padding around rows and headings, which I showed could be achieved with targeted classes.

So we agreed the best thing to do was to provide patterns *demos* with targeting padding classes, and then put all of that in storybook. That way someone can copy + paste and get started, and if necessary that can move a column or row around to suit their situation.

PR #398, opened 2020-02-25
fix(skip-to-content): use Carbon's SkipToContent component, update focus styles

### LINKED ISSUE #397: SkipToContent link does not have proper focus state
## Bug - SkipToContent link does not have proper focus state

**Expected behavior -**

When the `SkipToContent` link is focused, it should have a proper focus state. (3px solid `$focus` border)

**Actual behavior -**

The `border` is set to `0` for this link, so it does not have the right focus state.

jendowns:
## Affected issues

- Resolves #397 

Originally our `SkipToContent` link had `border: 0` due to the use of the `hidden` mixin. The component is not a typical button, but it should "appear" as a focused button (not a un-focused button) when it is tabbed to. 

This PR fixes that by matching Carbon's styles. 

## Proposed changes

- ~update styles to match Carbon's skip to content link: https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/ui-shell/_header.scss#L304 (the main difference though is that ours is a "primary" button that is also centered horizontally)~
- import and use Carbon's `SkipToContent` instead to recreating our own
- update focus styles


SimonFinney:
Sorry @jendowns , I'm only seeing this now - I added the `focus-outline` mixin in https://github.com/carbon-design-system/ibm-security/pull/285/files#diff-3384391611d359dc312ba198cda6bcdaR111 , and it doesn't appear to have any unintended side effects - https://deploy-preview-285--ibm-security.netlify.com/?path=/story/patterns-unstable-shell--skip-to-content

Are there any issues with that approach versus what is proposed here?

jendowns:
@SimonFinney In both cases I believe the `hidden` mixin is being used, which was the root of the original issue I think. 

The solution you have in the Shell refactor is cool but IMO it would be best to mimic what Carbon is doing as much as possible. In this case I think everything is copied from the Carbon version except for positioning, because Cameron had originally requested it be centered, I think.

jendowns:
@SimonFinney did you have any thoughts about the comment above?

Deriohig:
I don't quite understand all the context here, but I think that hardcoding something that can potentially be applied as a mixin is good practise even if its not exactly what carbon is doing. It provides consistency and the ability to change things globally.

jendowns:
@Deriohig @SimonFinney Looking at this again, there are several problems with the approach that is in the `UNSTABLE__Shell` (which was merged into the `next` branch).

## `UNSTABLE__Shell` style variation: 

You can see the `UNSTABLE__Shell` version that @SimonFinney wanted to emulate here: https://next--ibm-security.netlify.com/?path=/story/patterns-unstable-shell--skip-to-content

Most notably the way it's written, its not completely "hidden" when inactive. It sits on top of content & has a set width/height, z-index, visibility, etc.
Screen Shot 2020-03-16 at 10 41 46 AM

And then when focussed, the way it is being shown is with `clip: auto` which is not ideal:
Screen Shot 2020-03-16 at 10 41 26 AM

## My variation

The Carbon styles I copied here keep the component completely hidden -- TRULY hidden -- until it is focused.

Screen Shot 2020-03-16 at 10 41 57 AM

Screen Shot 2020-03-16 at 10 40 51 AM

The issue this PR is specifically addressing is the use of a `hidden` mixin that had unintended reset side effects.

So again I really recommend that we emulate Carbon here and ensure that the `UNSTABLE__Shell` also keeps those correct skip-to-link styles. As it stands, the `UNSTABLE__Shell`'s skip-to-link is not correctly hidden or correctly shown. There isn't an easy mixin we can use for this; we need these styles.

jendowns:
Ah one more thing, to specifically address the proposed use of `focus-outline` --

We are targeting a 4px width border here, to make the button clearly visible. 4px is not available via the `focus-outline` mixin: https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/globals/scss/_helper-mixins.scss#L65 (which is likely why Carbon isn't using that mixin, either)

So to summarize -- we shouldn't use the `hidden` mixin (which this PR is trying to fix) and we shouldn't use the `focus-outline` mixin (it's not sufficient).

jendowns:
@SimonFinney just a reminder that this is ready to review & I've added more explanation about the change in comments above. 👍

SimonFinney:
@jendowns From what I can see, the fix for what you're describing above is achievable by moving the margin values that are currently set in `&__link` to the `focus` declaration - the other values are the same output of the `hidden` mixin and feel redundant to reproduce in this case.

jendowns:
@SimonFinney unfortunately, the `hidden` mixin has unintended consequences & led me to make this issue. Carbon is not currently using it for their Skip link, either.

SimonFinney:
So I understand this better, what are the differences between the following screenshot (taken from `next` with the margin removed), and the screenshot you posted above?

Screenshot 2020-03-26 at 19 05 22

jendowns:
@SimonFinney are you taking issue with `margin: -1px`? What is it you are trying to show me here? 

In `next`, the margins exist and therefore the un-focused button takes up space.

In this PR, the margins are `-1px` without focus, therefore they don't take up space.

jendowns:
@SimonFinney perhaps it would help to link this: https://webaim.org/techniques/css/invisiblecontent/

snippet from the article:

```
// CSS clip

{
clip: rect(1px, 1px, 1px, 1px);
clip-path: inset(50%);
height: 1px;
width: 1px;
margin: -1px;
overflow: hidden;
padding: 0;
position: absolute;
}
```

> A fairly modern technique of using CSS to hide or clip content that does not fit into a 1 pixel visible area will essentially hide the content visibly, but still allow it to be read by modern screen readers.

src/components/Shell/_mixins.scss
```diff
@@ -103,22 +103,34 @@
   &__skip-to-content {
     display: flex;
     position: relative;
-    height: 0;
```
jendowns: Please note that I just removed this height rule, based on guidance about ensuring that the "hidden" content doesn't actually have a zero height: https://webaim.org/techniques/css/invisiblecontent/

We want the child of this element, the actual link, to be readable by a screenreader.

SimonFinney:
@jendowns I created a quick PR to illustrate my suggestions above, please take a look and let me know if you have any questions! https://github.com/carbon-design-system/ibm-security/pull/459

src/components/Shell/_mixins.scss
```diff
@@ -101,24 +101,16 @@
...
-      margin-left: -$toolbar__sizing__width;
-      transition: unset;
-      z-index: z($layer: floating);
-
-      &:focus {
-        width: auto;
-        padding: $button-padding;
-        clip: auto;
-      }
+      font-weight: 600;
```
jendowns: had to add this weight because as I was checking the component with DAP, the latest ruleset now flags it for now having enough contrast + weight:

Screen Shot 2020-03-27 at 3 14 00 PM

src/components/Shell/_mixins.scss
```diff
@@ -101,24 +101,16 @@
...
-      @include hidden;
-
+    &__link.#{$prefix}--skip-to-content:focus {
+      position: relative;
+      border-color: $focus;
+      background-color: $interactive-01;
       margin-top: $header__spacing__padding;
-      margin-left: -$toolbar__sizing__width;
-      transition: unset;
-      z-index: z($layer: floating);
```
jendowns: Something worth noting here -- `floating` is 10,000, and Carbon's `SkipToContent` is set to 9,999 by default.

SimonFinney:
Interesting! My only thought here is if we're going to significantly adjust the appearance of the work, we should pull in design ( @cameroncalder ) for a visual review

jendowns:
@SimonFinney this has been reviewed by @cameroncalder in Slack (instead of bolded text, needed to change gray10 to white). Please review again. 🙏

PR #400, opened 2020-02-25
feat(unstable-pagination): add page input child component, split out tests

### LINKED ISSUE #295: Unstable Pagination - add `NumberInput` subcomponent
### Proposal

Similar to how the composable `UNSTABLE__Pagination` component accepts a `PageSelector` (i.e., a `Select`) child, we should also add a new `NumberInput` ( called`PageInput`?) that can be used instead of the `Select`.

jendowns:
## Affected issues

- Resolves #295 

NOTE: tied to https://github.com/carbon-design-system/carbon/pull/5485

## Proposed changes

- add `PageInput` component with tests & a story with docs
- split out `PageSelector` tests into separate file
- add a11y test to `UNSTABLE__Pagination` test file that checks for the new `PageInput` child
- add a couple styles but these are largely based on assumptions 😅  Definitely willing to change this with some design direction 👍 
- temporarily set invalid text to "visually hidden" -- again, need some design guidance on what should happen when that text. 🤔 

## Testing instructions

- check out the storybook demo 👀

### DIFF
--- a/src/components/UNSTABLE__Pagination/PageInput.js
+++ b/src/components/UNSTABLE__Pagination/PageInput.js
@@ -18,6 +18,8 @@ function PageSelector({
   id,
   invalidText,
   label,
+  max,
+  min,
   totalPages,
   ...other
 }) {
@@ -29,8 +31,8 @@ function PageSelector({
       invalidText={invalidText}
       label={label}
       value={currentPage}
-      min={1}
-      max={totalPages}
+      max={max || totalPages}
+      min={min}
       {...other}
     />
   );
@@ -52,6 +54,12 @@ PageSelector.propTypes = {
   /** Translatable string to label the page input element. */
   label: PropTypes.string,
 
+  /** The maximum value accepted. */
+  max: PropTypes.number,
+
+  /** The minimum value accepted. */
+  min: PropTypes.number,
+
   /**
    * Total number of pages.
    * This value is calculated using a valid `totalItems` prop passed to the parent `UNSTABLE__Pagination`.
@@ -64,6 +72,8 @@ PageSelector.defaultProps = {
   className: null,
   id: 1,
   label: 'Current page number',
+  max: undefined,
+  min: 1,
   invalidText: 'Not a valid page number.',
 };
 
--- a/src/components/UNSTABLE__Pagination/PageInput.js
+++ b/src/components/UNSTABLE__Pagination/PageInput.js
@@ -1,50 +1,36 @@
 /**
- * @file Page selector.
- * @copyright IBM Security 2019
+ * @file Page input.
+ * @copyright IBM Security 2020
  */
 
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Select, SelectItem } from '../Select';
+import NumberInput from '../NumberInput';
 import { appendComponentNamespace } from '../../globals/namespace/index';
 import { namespace as paginationNamespace } from './Pagination';
 
-const namespace = appendComponentNamespace(
-  paginationNamespace,
-  'page-selector'
-);
+const namespace = appendComponentNamespace(paginationNamespace, 'page-input');
 
 function PageSelector({
   className,
   currentPage,
   id,
-  labelText,
+  label,
   totalPages,
   ...other
 }) {
-  const renderPages = total => {
-    const pages = [];
-    for (let counter = 1; counter <= total; counter += 1) {
-      pages.push(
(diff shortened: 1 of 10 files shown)


src/components/UNSTABLE__Pagination/PageInput.js
```diff
@@ -0,0 +1,84 @@
...
+  totalPages,
+  ...other
+}) {
+  const instanceId = `${namespace}__input-${getInstanceId()}`;
+
+  return (
+    <NumberInput
+      className={classnames(namespace, className)}
+      hideLabel
+      id={instanceId || id}
```
jendowns: I opted to do this for both the `PageInput` and `PageSelector` because they just have one place where `id` is used & it was easy enough to generate and use an instance ID that could be replaced.

However, I opened not to do this for the parent pagination component because that component ends up using an id throughout for setting a11y attributes etc. I didn't get to make the default of that parent component's `id` equal to `getInstanceId()` or anything like that. 🤔 Seemed... strange to do so? 

But let me know if you disagree / have concerns 👍

src/components/UNSTABLE__Pagination/PageInput.js
```diff
@@ -0,0 +1,84 @@
...
+      invalidText={invalidText}
+      label={label}
+      value={currentPage}
+      max={max || totalPages}
+      min={min}
```
jendowns: Even though ideally someone wouldn't be setting these values, seems worthwhile to let them override defaults. 🤷‍♀

src/components/UNSTABLE__Pagination/_index.scss
```diff
@@ -151,4 +151,20 @@
...
+
+    &.#{$prefix}--number input[type='number'] {
+      border: none;
+      background-color: $ui-03;
+    }
+
+    // Visually hide error text for now:
+    .#{$prefix}--form-requirement {
+      @include hidden;
+    }
```
jendowns: The "error message" text is visually hidden for now but I'm unsure what to do with it? 😅 Thoughts?

src/components/UNSTABLE__Pagination/Pagination.stories.js
```diff
@@ -92,7 +92,69 @@ storiesOf(components('UNSTABLE Pagination'), module)
...
+        {...props()}
+        totalItems={350}
+        pageSizes={array('Choices of `pageSize` (pageSizes)', [10, 20, 30])}
+      >
+        {({ currentPage, onSetPage, totalPages }) => (
+          <PageInput
+            currentPage={currentPage}
+            totalPages={totalPages}
+            id="number-input-1"
+            onChange={event => onSetPage(event.imaginaryTarget.value)}
```
jendowns: It's super important that people use `event.imaginaryTarget.value` instead of typical `event.target.value` because this component is controlled. 🤔 Do you think it is clear enough here & in the `info` for the story?

src/components/UNSTABLE__Pagination/PageSelector.js
```diff
@@ -37,7 +41,7 @@ function PageSelector({
     <Select
       className={classnames(namespace, className)}
       hideLabel
-      id={`${namespace}__input-${id}`}
+      id={instanceId || id}
```
jendowns: Similar to `PageInput` -- it seems ideal to generate this ID ourselves since it's just used in one spot, and people can override if they want.

src/components/UNSTABLE__Pagination/Pagination.stories.js
```diff
@@ -34,7 +34,7 @@ const props = () => ({
 storiesOf(components('UNSTABLE Pagination'), module)
   .addDecorator(story => <div style={{ width: '800px' }}>{story()}</div>)
   .add(
-    'default',
```
jendowns: No longer seems appropriate to call something "default" here 😅

jendowns:
I'm going to close out this PR for now.

I had proposed a similar PR to Carbon core, to add all parts of this new `Pagination` component: https://github.com/carbon-design-system/carbon/pull/5485

But there needs to be a design exploration of the `PageInput` (i.e., number input) specifically. That work is being tracked by the Carbon design team here: https://github.com/carbon-design-system/carbon/issues/5546

So we shouldn't add a number input at this time, not until we hear back about how it should look/behave. I'm going to close this out for now & reopen/revisit it when the Carbon design team finishes their exploration.

(Side note: https://github.com/carbon-design-system/carbon/pull/5485 will still remain open but without the `PageInput` subcomponent)

PR #405, opened 2020-02-26
feat(card-skeleton): allow custom classname and extra props via spread attr

### LINKED ISSUE #295: Unstable Pagination - add `NumberInput` subcomponent

### LINKED ISSUE #404: <CardSkeleton> does not allow for className definition
## Bug - 
CardSkeleton does not allow for additional props such as className. Perhaps we could spread the props that arent used as well as append a classname in case aria labels may be required for long load times. 

**Expected behavior -**

CardSkeleton takes a className or aria attributes without issue

**Actual behavior -**

CardSkeleton does not allow for className or any aria attributes

### Steps for reproducing

add a class to the component or view [source code](https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/Card/CardSkeleton/CardSkeleton.js).

### Optional information

**Version -**  all of them I imagine since first support

jendowns:
## Affected issues

- Resolves #404

## Proposed changes

- use function declaration instead
- move `WIDTH` globals into component
- allow `className` prop and spread attribute for extra props


src/components/Card/CardSkeleton/CardSkeleton.js
```diff
@@ -3,39 +3,54 @@
...
  */
 
+import classnames from 'classnames';
+import { PropTypes } from 'prop-types';
 import React from 'react';
 
 import { getComponentNamespace } from '../../../globals/namespace/index';
 
-import Card from '..';
+import Card from '../Card';
```
jendowns: Note that this was throwing a linting warning previously, because it was importing a default from the `index.js` instead of just directly from `Card/index.js`

src/components/Card/CardSkeleton/CardSkeleton.js
```diff
@@ -3,39 +3,54 @@
...
+function CardSkeleton({ className, ...other }) {
+  const classes = classnames(namespace, className);
+
+  const widths = {
+    sm: '25%',
+    md: '50%',
+    lg: '75%',
+  };
+
+  const { sm, md, lg } = widths;
```
jendowns: Didn't make sense to keep these as "globals" in this file -- they aren't being used anywhere else.

src/components/Card/CardSkeleton/CardSkeleton.js
```diff
@@ -3,39 +3,54 @@
  * @copyright IBM Security 2019
  */
 
+import classnames from 'classnames';
+import { PropTypes } from 'prop-types';
```
SimonFinney: This doesn't throw any errors so it doesn't seem like a big issue, but elsewhere we use either `import PropTypes from 'prop-types';` or `import { string } from 'prop-types';` - what's your preference?

jendowns: Oh sure thing, I'll remove the brackets.

PR #417, opened 2020-03-03
feat: update to Carbon 10.10.0 release

### LINKED ISSUE #389: Upgrade to Carbon 10.10.0 release

jendowns:
## Affected issues

- Resolves #389 

## Proposed changes

- upgrade packages to [Carbon's 10.10.0 release](https://github.com/carbon-design-system/carbon/releases/tag/v10.10.0).
- update snapshots for updated inner Carbon core components
- update `FilterPanelSearch` tests in particular -- previously using `getByLabelText` because they only had one occurrence of the label text, but with this update there are now two occurrences. ~So I updated the affected tests to use `getAllByLabelText` & grab the first occurrence.~ I updated the tests to use placeholders + `getByPlaceholderText` queries instead.

## Testing instructions

- Review release notes to check for breaking changes: https://github.com/carbon-design-system/carbon/releases/tag/v10.10.0
- Review each component in the storybook deploy & compare to our current release.


src/components/FilterPanel/FilterPanelSearch/__tests__/FilterSearch.spec.js
```diff
@@ -19,24 +19,24 @@ describe('FilterPanelSearch', () => {
...
+    const { queryByTestId, getAllByLabelText } = render(
       <FilterPanelSearch labelText="search label">
         <div data-testid="result-content" />
       </FilterPanelSearch>
     );
 
-    const searchInput = getByLabelText('search label');
+    const searchInput = getAllByLabelText('search label')[0];
     searchInput.focus();
     searchInput.blur();
```
jendowns: @diego-codes Just bringing this to your attention -- anything amiss in this change?

src/components/FilterPanel/FilterPanelSearch/__tests__/FilterSearch.spec.js
```diff
@@ -19,24 +19,24 @@ describe('FilterPanelSearch', () => {
   });
 
   it('renders results content when keyboard focus is added', () => {
-    const { getByTestId, getByLabelText } = render(
+    const { getByTestId, getAllByLabelText } = render(
```
jendowns: @diego-codes Just bringing the updates in this file to your attention -- anything amiss?

jendowns:
Sorry for the CircleCI build spam 😭 I needed to clear my yarn cache.

src/components/FilterPanel/FilterPanelSearch/__tests__/FilterSearch.spec.js
diego-codes: I suspect that the labels changed so `getByLabelText` errors out because there is more than one element with this label?

It might be worth switching the queries to `getByPlaceholderText` so you're not dependent on the order of elements that are labeled `"search label"`

jendowns: Thanks @diego-codes! Unfortunately `getByPlaceholderText` won't work here. 🤔 (I suspect because we are dealing with `label` elements, not placeholder attributes)

diego-codes: Sorry, I meant that you can provide a placeholder text to the search component and query it with that placeholder text.

jendowns: Just committed the update 👍

jendowns:
@SimonFinney Would you be able to take a look at this today? This PR is blocking a couple testing PRs (specifically there are some a11y fixes that we need, before some of our own a11y tests can pass)

SimonFinney:
@jendowns As I understand it, we previously assigned this work item to be completed next sprint - https://github.com/carbon-design-system/ibm-security/milestone/11

I thought it was sufficient to leverage the release candidate for https://github.com/carbon-design-system/ibm-security/pull/396 ?

jendowns:
That's fine, I was just hoping it was possible & was trying to suggest a PR that could help close out others 👍 
I just tagged #396 above, but there's also #390

SimonFinney:
Thanks @jendowns ! I did some manual verification against `dev`, and the only issue I found is related to `ProgressIndicator` alignment:

ProgressIndicator

jendowns:
@SimonFinney I'm cool deferring to #439 but in addition to the snap tests + `ProgressIndicator` alignment, that PR would also need to update tests for `FilterPanelSearch` (just need to use `getAllByLabelText` etc., like here: https://github.com/carbon-design-system/ibm-security/pull/417/files/deb35f0c65767726e71a9b977de5bb7a8c221f3c#diff-153d2e60e874551550ccb8491ff5b713)

jendowns:
Hey @SimonFinney I got this sorted out and it's ready for review again 👍  Should be the correct number of offline files being updated as well now. I also went ahead and updated the story for the `ProgressIndicator` to match Carbon's, as they now include and demonstrate a prop to make the indicator vertical (which will ultimately mean we can deprecate our `StepIndicator` btw 👀 )

As for the `ProgressIndicator` alignment you mentioned in https://github.com/carbon-design-system/ibm-security/pull/417#issuecomment-600653220, I believe that is a Carbon issue that I need to investigate.

Compare our storybook for this branch -- https://deploy-preview-417--ibm-security.netlify.com/?path=/story/components-progressindicator--default
To theirs -- https://react.carbondesignsystem.com/?path=/story/progressindicator--default

And inspect the "Second step..." label in Chrome or Safari (it looks fine in Firefox). Our text is smaller, because the `Tooltip` label is getting precedence. Whereas theirs is larger, because the progress indicator's components proper label is taking precedence.

They appear to have an ordering issue here, and we are affected by it because our order arbitrarily different. I need to isolate this issue more & fix on their end for sure.

SimonFinney:
That's awesome that we can supersede the `StepIndicator`! I double-checked the component and there are now some hover issues introduced 😭:

StepIndicator

I'll approve to keep this moving, but I'll leave it up to you as to whether we should fix this now or defer to a defect, or follow-up for deprecating the `StepIndicator`.

jendowns:
Yes I think I will follow up with three things here —

1. Quick PR here to fix cosmetic Issues in the ProgressInidcator (I can open this today)
2. v2.x issue to deprecate StepIndicator (will need to involve updating the Wizard as well)
3. See about an upstream fix in Carbon for the cosmetic issues

PR #418, opened 2020-03-03
feat(progress-animation): add component with demo

### LINKED ISSUE #342: Progress indicator visualization
### Details

- **Application / team -** Design team
- **Needed by -** ASAP

### Proposal

image

We'd like to be able to pass in a percentage and have that reflect in the visualization, as well as specific colors. It would also be great if it could also animate on load (ability to turn that functionality on/off).

SVG code:
```
<?xml version="1.0" encoding="UTF-8"?>
<svg width="32px" height="32px" viewBox="0 0 32 32" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <!-- Generator: Sketch 62 (91390) - https://sketch.com -->
    <title>Oval 8 Copy 8</title>
    <desc>Created with Sketch.</desc>
    <defs>
        <circle id="path-1" cx="16" cy="16" r="16"></circle>
        <mask id="mask-2" maskContentUnits="userSpaceOnUse" maskUnits="objectBoundingBox" x="0" y="0" width="32" height="32" fill="white">
            <use xlink:href="#path-1"></use>
        </mask>
    </defs>
    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
        <g id="Oval-8-Copy-8">
            <circle id="Oval-8" stroke="#D5D9E0" stroke-width="2" cx="16" cy="16" r="15"></circle>
            <use id="Oval-8" stroke="#8A3FFC" mask="url(#mask-2)" stroke-width="4" stroke-dasharray="80,48" transform="translate(16.000000, 16.000000) scale(1, -1) translate(-16.000000, -16.000000) " xlink:href="#path-1"></use>
        </g>
    </g>
</svg>

```

jendowns:
## Affected issues

- Resolves #342 

## Proposed changes

- add `ProgressAnimation` component with light styles & demo

## Testing instructions

- check out the `ProgressAnimation` story in the deploy (https://deploy-preview-418--ibm-security.netlify.com/?path=/story/components-progressanimation--default) & play around with the knobs.


src/components/ProgressAnimation/index.js
```diff
@@ -0,0 +1,148 @@
...
+       * Otherwise, a new style tag with a unique keyframe needs to be
+       * created per component instance, because the unique progress
+       * percetange is used in the keyframe animation.
+       */}
+      {animationTimer ? (
+        <style
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={{
+            __html: `
+              @keyframes ${progressAnimationName} {
```
jendowns: For an animated instance of this component, it is required that it is animated to the percentage value provided. Which to me, means that the svg path needs to move from a "0" value to the provided percentage complete value.

To achieve this with keyframes (which seemed like the best solution? let me know if you disagree), we need to insert a style tag in this manner so that *each* instance of the component has each own keyframes, based on its own unique percentage complete value.

Worth noting that I did explore using `emotion` and `styled-components`, but I don't believe those solution are ideal for us. We need these keyframes to be in the global scope AND also have the prop `percentage` value. Hence, the solution above. 🤔 But please let me know what you think!

src/components/ProgressAnimation/index.js
```diff
@@ -0,0 +1,148 @@
...
+                  stroke-dashoffset: 0;
+                }
+                to {
+                  stroke-dashoffset: ${percentage};
+                }
+            `,
+          }}
+        />
+      ) : (
+        ''
```
jendowns: I changed this to a ternary instead of the this pattern...

```jsx
animationTimer && (
  <Style />
)
```

because weirdly, when I would set `animationTimer` to `0` in the storybook knobs, a `0` character was being rendered ON the component? 🤔 Bizarre...

SimonFinney: That makes sense as the `false` value will convert to `0` when coerced as a string

jendowns: Yes, I understand that. 😅😉 That’s why I used the ternary.

src/components/ProgressAnimation/index.js
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Optional custom timer for this component instance's animation, in seconds.
+   * If value is nonexistant (i.e., set to its default `null`, `0`, or `undefined`),
+   * then the component will not be animated at all.
+   */
+  animationTimer: PropTypes.number,
+
+  /** Unique id for this component instance. */
+  id: PropTypes.number.isRequired,
```
SimonFinney: - [ ] I recommend enforcing this as a string here as it will provide better flexibility for `id`s that most likely would be generated with both alphabetical and numerical characters

jendowns: I thought about that, however this ID is appended to a class name so i worried that allowing a string would open it up to non-classname-friendly characters. 🤔

jendowns: I said class name above but i meant the name of the keyframe animation 😅

SimonFinney:
@jendowns Per our conversation on Slack yesterday, I created a quick proof-of-concept of achieving this using hooks and transitions - https://github.com/carbon-design-system/ibm-security/pull/424

Let me know your thoughts!

jendowns:
Closing out this PR since we did not intend to merge it into the library.

PR #421, opened 2020-03-04
docs(checkbox-skeleton): add required props to story demo

jendowns:
The `CheckboxSkeleton` has two required props (`id` and `labelText` that are no included in the storybook demo). In development, this component generates console errors due to the missing required props & creates noise:

Screen Shot 2020-03-04 at 3 26 16 PM

## Proposed changes

- add required `CheckboxSkeleton` props to storybook demo


jendowns:
Closing this PR because it is the wrong way to fix these errors. The props aren't required; they are being erroneously pulled from the non-skeleton version of the component. 👍

PR #424, opened 2020-03-05
feat(progress-animation): use hooks

SimonFinney:
## Proposed changes

- Combine CSS and hooks to achieve animation
- Use string identifier

### DIFF
--- a/src/components/ProgressAnimation/_index.scss
+++ b/src/components/ProgressAnimation/_index.scss
@@ -7,12 +7,21 @@
 @import '../../globals/namespace/index';
 @import '../../globals/layout/index';
 
-@include export-namespace($name: 'progress-animation') {
-  .#{$namespace}progress-animation {
-    display: flex;
+@import '../Component/mixins';
+
+@include security--component($name: progress-animation) {
+  display: flex;
+
+  &__stroke {
+    stroke-dasharray: 100;
+    stroke-dashoffset: 0;
+
+    &--animated {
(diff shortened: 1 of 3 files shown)


src/components/ProgressAnimation/_index.scss
```diff
@@ -7,12 +7,21 @@
...
+    stroke-dasharray: 100;
+    stroke-dashoffset: 0;
+
+    &--animated {
+      transition-property: stroke-dashoffset;
+    }
   }
 
-  .#{$namespace}progress-animation__message {
+  &__message {
```
jendowns: Would you mind restoring the previous style structure?
These changes are meant to be copied by a product team, and they wouldn't be using the `security--component` mixin in this manner. It would be more helpful for them to have a flat style structure so that whatever their scss/css situation is, they can quickly absorb these styles. 👍

SimonFinney: Good point, I probably got a bit trigger happy here! 😆I'm good with not merging all of these changes in, but rather `ProgressAnimation/index.js` can be leveraged as inspiration for the originating PR as you see fit

jendowns: Okay, are you saying you want to close this PR, not merge it with any update. And see changes reflected in the history #418 instead?

jendowns:
Closing this PR as discussed, and as #418 has also been closed. This component was never meant be merged, only a demo from the previous sprint.

PR #429, opened 2020-03-09
test(api): add snapshot test for public API interface

### LINKED ISSUE #426: Add snapshot test for public component api interface
### Proposal

Let's add a snapshot for our public component API, similar to what Carbon has already done & also what the IoT PAL is trying to do right now as well.

Related: https://github.com/IBM/carbon-addons-iot-react/pull/980
Related: https://github.com/carbon-design-system/carbon/pull/5495

jendowns:
## Affected issues

- Resolves #426 

## Proposed changes

- add snapshot test for public API interface that is the same as https://github.com/carbon-design-system/carbon/pull/5495 & https://github.com/IBM/carbon-addons-iot-react/pull/980
- fix the `DataTable` subcomponent exports (they were making the public API interface test crash with a `Maximum call stack size exceeded` error 😬 )
- remove the individual `DataTable` subcomponent exports test (no longer necessary with the public API snapshot anyway, since those are all included in that test)


jendowns:
Sorry for the CircleCI build fail spam -- once again, my yarn cache needed to be cleared. 😆

jendowns:
@SimonFinney I recommend we introduce this test prior to merging https://github.com/carbon-design-system/ibm-security/pull/438 (or any similar PR) to ensure that we have a full grasp of public API updates.

src/components/__tests__/publicAPI.spec.js
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        api[key] = Component[key];
+      });
+
+      return api;
+    }
+
+    // eslint-disable-next-line global-require
+    const CarbonAddonsIoTReact = require('../../index');
```
SimonFinney: I'm wondering if we can move this import to the top of this file, similar to how we handle other imports? With that, we should also be able to avoid using a global require - For example:

```js
import * as Components from '../../index';
```

jendowns: For sure, I understand this — as mentioned in the PR desc, the test is copied exactly as-is.

joshblack:
@SimonFinney I'd be down! We have some similar ideas for Public API matching that we could also expose for:

- Verifying `package.json` changes
  - Major use-case is specifying adding a `peerDependency` as a breaking change
- Verifying stable sass paths
  - For paths that the package intends to expose, make sure the paths don't change and verify that they can compile
- Verifying sass public exports
  - Check that sass public exports are available and match the expected type (if possible)

Issue #434, opened 2020-03-11
`Shell` - Add accessibility tests and resolve violations


SimonFinney:
## Description

Add accessibility tests and resolve violations. (based on the original unedited version of this review: https://github.com/carbon-design-system/ibm-security/pull/285#pullrequestreview-370603953)

### Skip to content link

On the `a` element:
- [ ] **DAP Violation: Each form control must have an associated label.** (Form control element `a` is missing associated label.)  [Checkpoint 1.3.1 - Info and Relationships](http://www.ibm.com/able/guidelines/ci162/info_and_relationships.html); [Checkpoint 4.1.2 Name, Role, Value ](http://www.ibm.com/able/guidelines/ci162/name_role_value.html)
- [ ] **DAP Violation: All content must reside within a HTML5 Sectioning element, WAI-ARIA landmark or labelled region role.** (Ophan content is not in a WAI-ARIA landmark for labelled region role.) [Checkpoint 2.4.1 - Bypass Blocks](https://www.ibm.com/able/guidelines/ci162/bypass_blocks.html)

Styles:
- [ ] The link is not properly hidden when not-focused -- it still has a z-index, height/width, and visibility just not content due to the `clip` style. And `clip: auto` is the only style being used to "show" the button when focused. (See this comment for more context: https://github.com/carbon-design-system/ibm-security/pull/398#issuecomment-599610372)

### "Return to..." banner

On the `a` element:
- [ ] **DAP Violation: All content must reside within a HTML5 Sectioning element, WAI-ARIA landmark or labelled region role.** (Ophan content is not in a WAI-ARIA landmark for labelled region role.) [Checkpoint 2.4.1 - Bypass Blocks](https://www.ibm.com/able/guidelines/ci162/bypass_blocks.html)

### Toolbar

On the `a#security--nav__list__item` element:
- [ ] **DAP Violation: Element ids must be unique within the document.** (The ids `security--nav__list__item` are used repeatedly, and they should not be.) [Checkpoint 4.1.1 - Parsing](http://www.ibm.com/able/guidelines/ci162/parsing.html) **NOTE:** Was starting to address this in #385 , so may be more appropriate to defer to that PR & fixing the inner `NavListItem` etc. components)

On the `Toolbar`'s `aside` element:
[...]

PR #436, opened 2020-03-11
fix(header-action): add a11y tests and update role

jendowns:
## Affected issues

- Part of #434 
(doing these incrementally where it makes sense)

## Proposed changes

-  add a11y tests for when `HeaderAction` has an account list, notifications, and an ordinary button
- changed `role="complementary"` to `role="dialog"` which seems to make more sense given its usage: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role


jendowns:
Yes @Deriohig! I believe the unit tests in general were something @SimonFinney wanted to work on, whereas adding some a11y tests are part of a followup for a review I did for #285.

PR #437, opened 2020-03-11
fix(header-notification): add visible hover state to clear button

### LINKED ISSUE #435: HeaderNotification - close button's hover state not distinguishable from surrounding div
**Expected behavior -**

When you hover over the "close" button for an individual notification in a header popover, that hover state should be distinguishable from the hover state of the surrounding div.

**Actual behavior -**

When you hover over a notification (i.e., the whole notification "row" or wrapping div), then the notification row's background color changes and the close button appears.

However, when you then move your mouse over the close button for that notification, the button itself should have its own hover state that is distinguishable from the hover state of the surrounding div. The button's hover state has no additional change, and the cursor is also a default/auto cursor, not a pointer.

In other words: there's no visual indication to users that they can click on the close button, via a background color + cursor change.

### Steps for reproducing

https://ibm-security.carbondesignsystem.com/?path=/story/patterns-header--with-notifications

jendowns:
## Affected issues

- Resolves #435 
- Also related to a review for #285 

## Proposed changes

- ~add specific class name for "clear" notification button~
- add visible hover state for clear button (also use background change + pointer cursor for `active` and `focus` states)

## Testing instructions

- I used `$active-tertiary` as the background color 😬 It looks great for the gray100 theme but not sure about the coolGray10 theme. Would definitely need a designer to take a look. Tried my best based on available theme colors here: https://carbon-elements.netlify.com/themes/examples/preview/


src/components/Header/HeaderNotification/HeaderNotification.js
```diff
@@ -39,7 +40,10 @@ const HeaderNotification = ({
     </a>
     <div className={`${namespace}__wrapper`}>
       <button
-        className={`${namespace}__button`}
+        className={classnames(
+          `${namespace}__button`,
+          `${namespace}__button--clear`
```
SimonFinney: I'm curious to hear more about the rationale behind applying a new class here, versus adding the styles to `${namespace}__button`

jendowns: There are other buttons that share the same class but don’t need these particular styles. The base class provides a number of general styles for all header buttons, and this extra class provides some specific overrides.

SimonFinney: As far as I recall, this class will output as `security--header__notification__button`, which should be scoped to this component

jendowns:
Thanks @SimonFinney this PR has been updated 👍

Deriohig:
@jendowns could you add a screenshot of the component that this updated? thanks!

jendowns:
@Deriohig it's for this story here: https://deploy-preview-437--ibm-security.netlify.com/?path=/story/patterns-header--with-notifications (the hover state of the "clear" button for an individual notification)

Before:
Screen Shot 2020-03-16 at 9 51 06 AM

After:
Screen Shot 2020-03-16 at 9 51 59 AM

I'm not 100% sure what color token to use though. (tokens: https://carbon-elements.netlify.com/themes/examples/preview/) The one I picked looks decent but let me know if you have a better one. I'm trying to pick an appropriately-named token role 🤔

Deriohig:
No, this looks good to me. I just couldnt figure out what button it was !

PR #441, opened 2020-03-13
feat(decorator): deprecate inert prop and fix color contrast

### LINKED ISSUE #440: DataDecorator value is below minimum WCAG AA text contrast ratio on cg100 theme
The text contrast on the blue value in the `DataDecorator` component is flagging as a violation on the DAP tool (see screenshots).

> Contrast of 3.67 with its background is not sufficient to meet WCAG AA minimum requirements for text of size 14px and weight of 400.

Just for context – in our case on Falcon, we have a data table of QRadar offenses with each row containing at least one data decorator component to represent IP addresses relating to each offense. As a result, we're seeing this violation duplicated many times while running the DAP tool.

NOTE: I'm sure design might want to take a quick look but I had no luck when briefly playing around with the blue text theme tokens from Carbon to pass contrast on both light and dark theme. It seems the background color of the decorator ($hover-ui) doesn't play well with the blue text tokens. Changing the background to $field-01 resolved the contrast issue though so this could be a potential solution.

**Expected behavior -**
`DataDecorator` value should meet WCAG AA text contrast guidelines (minimum 4.5:1 ratio). It should not flag a DAP violation.

**Actual behavior -**
`DataDecorator` value does not meet WCAG AA text contrast guidelines. It flags a DAP violation.

### Steps for reproducing

1.  Go to the `DataDecorator` component's [storybook](https://ibm-security.carbondesignsystem.com/?path=/story/patterns-datadecorator--default)
2. Ensure theme is `Cool Gray 100`
3. Run the DAP tool in dev tools
4. View results

### Screenshots

Screenshot 2020-03-13 at 10 45 07
Screenshot 2020-03-13 at 10 45 46

jendowns:
## Affected issues

- Resolves #440 

## Proposed changes

scss:
- update interactive color to be accessible against background
- update active color to also be accessible
- ensure that inert decorator's value text color does not look like a link (add more specificity)

js:
- deprecate `inert` prop and make the component "inert" by default (it will be interactive only if `href` or `onClick` prop is provided)

## Testing instructions

- run the DAP tool on the component


jendowns:
@ShayMurnin could you take a look when you get a chance & let me know if this works for you? https://deploy-preview-441--ibm-security.netlify.com

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -64,13 +64,13 @@ class Decorator extends Component {
...
-      [`${namespace}--link`]: this.props.href,
-      [`${namespace}--active`]: this.props.active,
+      [`${namespace}--link`]: this.props.href && !inert,
+      [`${namespace}--active`]: this.props.active && !inert,
       [`${namespace}--inert`]: this.props.inert,
       [`${namespace}--inline`]: this.props.inline,
     });
 
-    if (this.props.href) {
+    if (this.props.href && !inert) {
```
jendowns: @diego-codes When you get a chance, I was curious if you could take a look at the above & see if this change would be all right with you? I feel like this makes sense since `inert` is meant to mean that a component is not interactive. So it shouldn't get interactive classes (like `--link` or `--active`) and it also shouldn't render as an anchor.

Currently, an `inert` decorator could still render as an `<a>` if an `href` is provided 😬

diego-codes: Yeah, that makes sense to me. `inert` would remove any interactivity, which makes sense. I think a potential API update is to think of props being additive than subtractive, which `inert` is IMO. 

Perhaps a decorator is inert when no `onClick` or `href` is provided or something like that? Inertion (is that a word?) could be implied based on other props. That way, you don't end up with different props as we have now where `inert` and `href` could contradict each other.

ShayMurnin:
Thanks @jendowns! All good — DAP tool passing on both themes 🙌

src/components/DataDecorator/Decorator/Decorator.js
Deriohig: hey @diego-codes, just to clarify, you are suggesting to update and remove the `inert` prop and make this component's interactivity based on `href || onClick` ?

diego-codes: Yep! If the consumer does not provide a prop to make the decorator interactive, then it should be implied that it is inert.

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -71,29 +74,28 @@ class Decorator extends Component {
       scoreThresholds
     );
     const decoratorClasses = classnames(namespace, classes, className, {
-      [`${namespace}--link`]: this.props.href,
-      [`${namespace}--active`]: this.props.active,
-      [`${namespace}--inert`]: this.props.inert,
-      [`${namespace}--inline`]: this.props.inline,
+      [`${namespace}--link`]: href,
+      [`${namespace}--button`]: onClick,
```
jendowns: Need this class so that I can target it in stylesheet to add a pointer cursor on hover.

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -71,29 +74,28 @@ class Decorator extends Component {
...
-    if (inert) {
-      return <span className={decoratorClasses}>{decorator}</span>;
-    }
-
-    return (
-      <button className={decoratorClasses} onClick={this.handleClick}>
-        {decorator}
-      </button>
-    );
+    return <span className={decoratorClasses}>{decorator}</span>;
```
jendowns: "inert" span is now the default. If `href` is provided, we render a link. If `onClick` is provided, we render a button.

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -153,9 +159,8 @@ Decorator.defaultProps = {
   active: false,
   className: '',
   href: undefined,
-  inert: false,
   inline: false,
-  onClick: () => {},
+  onClick: undefined,
```
jendowns: Need to be `undefined` so that the `onClick` prop can be properly assessed as "not provided" when determining what type of element to render the Decorator as.

src/components/DataDecorator/Decorator/Decorator.stories.js
```diff
@@ -27,18 +27,38 @@ const storyProps = () => ({
...
+      </p>
+    </>
+  ))
+  .add('with onClick', () => (
+    <>
+      <p className="bx--type-body-long-01">This Decorator is a button.</p>
+      <p>
+        <Decorator {...storyProps()} onClick={action('onClick')} />
+      </p>
+    </>
```
jendowns: Because the visual distinctions here can sometimes be unclear unless you inspect or hover over a given element, I think it is best to say which variation is being displayed.

src/components/DataDecorator/Decorator/Decorator.js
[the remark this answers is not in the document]
jendowns: I saw that the button was the only one lacking the pointer so just went for that. The `--link` class exists currently because there are some specific link-only styles. Seems better to add a class for this case rather than restructure + take something away.

jendowns:
Thank you @diego-codes I made the changes you suggested above & also updated the tests to reflect the deprecation of the `inert` prop. Could you take another look when you get a chance? Thanks!

jendowns:
@SimonFinney @diego-codes just a reminder that this is ready to review again, if you have time 🙏

src/components/DataDecorator/Decorator/Decorator.stories.js
```diff
@@ -27,18 +27,38 @@ const storyProps = () => ({
...
+    <>
+      <p className="bx--type-body-long-01">This Decorator is inert.</p>
+      <p>
+        <Decorator {...storyProps()} />
+      </p>
+    </>
+  ))
   .add('Inline', () => (
     <p className="bx--type-body-long-01">
       This is an inline decorator <Decorator {...storyProps()} inline /> that
```
SimonFinney, line 45: - [ ] Can we ensure casing is consistent across the helper text for each of these stories? For example, either 'Decorator' or 'decorator'

jendowns: Sure thing 👍 But other than that, is this good to go?

SimonFinney:
@jendowns The code LGTM! The only request I have is that we loop in design in to visually sign off on the changes

jendowns:
@SimonFinney Just checking is the code approved? I will reach out for a design review.

PR #448, opened 2020-03-20
test(public-api): sort components to ignore order

SimonFinney:
## Proposed changes

- Sort components before creating snapshot to ignore reorganisation of exports
- Add `jest` script to allow individual file testing


jendowns:
Btw @SimonFinney the way I've been testing individual files: `yarn test -u -t="Component"`

PR #450, opened 2020-03-23
feat(carbon): mark carbon core pkgs as peer dependencies

### LINKED ISSUE #402: Mark Carbon packages as `peerDependencies`
## Description

If Carbon packages are not marked as `peerDependencies`, any team consuming Carbon for IBM Security and requiring anything from Carbon's packages are most likely importing separate versions of Carbon.

This will be considered a breaking change, and should happen in v2.

### Acceptance criteria

- [ ] All Carbon packages are marked as `peerDependencies`
- [ ] The change is documented in migration documentation

jendowns:
## Affected issues

- Resolves #402 

## Proposed changes

- mark `carbon-components`, `carbon-components-react`, and `@carbon/*` elements packages as peer dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,6 +159,16 @@
     "yarn": "^1.16.0"
   },
   "peerDependencies": {
+    "@carbon/colors": "10.7.2",
+    "@carbon/grid": "10.8.2",
+    "@carbon/icons-react": "10.8.2",
+    "@carbon/import-once": "10.3.0",
+    "@carbon/layout": "10.7.3",
+    "@carbon/motion": "10.5.2",
+    "@carbon/themes": "10.9.3",
+    "@carbon/type": "10.8.3",
+    "carbon-components": "10.9.3",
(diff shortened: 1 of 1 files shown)


jendowns:
@SimonFinney I meant to open this as a draft for now, as the work will be dependent on #445 (or rather, should certainly follow that) due to import statements needing to change. 🤔

package.json
```diff
@@ -51,8 +51,6 @@
     "@carbon/motion": "10.5.2",
     "@carbon/themes": "10.9.3",
     "@carbon/type": "10.8.3",
-    "carbon-components": "10.9.3",
```
SimonFinney: - [ ] I believe these still need to remain as `dependencies` in order for us to leverage them in the codebase
- [ ] For full compatibility, we should also include the Carbon elements packages here

jendowns:
Ooops! Some confusion on my end. This is good to go now @SimonFinney 🎉

PR #454, opened 2020-03-24
fix(scroll-gradient): apply scroll state styles to direct children only

diego-codes:
## Proposed changes

- Fix bug that a parent scroll gradient applies gradients on child scroll gradients.

## Testing instructions

- Check out the new story added for the ScrollGradient component
- Notice that the internal scroll gradient does not get its gradient activated by its parent

### DIFF
--- a/src/components/ScrollGradient/ScrollGradient.stories.js
+++ b/src/components/ScrollGradient/ScrollGradient.stories.js
@@ -87,8 +87,9 @@ storiesOf(components('ScrollGradient'), module)
         style={style}
         className={className}
       >
-        Content fits completely within content area, so neither scrollbar nor
+        <p>Content fits completely within content area, so neither scrollbar nor
         scroll gradients are necessary.
+        </p>
       </ScrollGradient>
     ),
     {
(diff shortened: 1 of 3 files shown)


src/components/ScrollGradient/ScrollGradient.stories.js
```diff
@@ -98,4 +99,32 @@ storiesOf(components('ScrollGradient'), module)
...
+      className={className}
+    >
+      <p>Content before the scrollable element.</p>
+
+      <ScrollGradient
+        color={color('child color (color)', gradientColor)}
+        style={{ height: '6rem', margin: '1rem 0'}}
+      >
+        <p>Does not need to scroll so does not have a gradient.</p>
+      </ScrollGradient>
```
jendowns: For this example I tried to see what happens in this case if the `>` child combinater you added is removed, to see if a scroll gradient is incorrectly applied here. But I didn't see any extra scroll gradients 🤔 

Screen Shot 2020-03-25 at 12 11 32 PM


Am I testing this correctly? Could you show me how to recreate the bug? Thank you!

diego-codes: Hi there! Thanks for the prompt review 😄. Here is what I did to do test the as-is:
1. I turned off the CSS changes I made
2. I used the knobs to change both scroll gradient colors to some obnoxious ones that stand out more
3. Scrolled to notice that when the gradients on the parent appear, so does the child's ones.
Kapture 2020-03-25 at 13 35 01

To test the PR (to-be), added the CSS changes back in and made sure to keep the obnoxious gradient colors:
Kapture 2020-03-25 at 13 47 20

Notice how the child scroll gradient is not shown so they are independent of each other.

jendowns: Thank you!

PR #469, opened 2020-04-02
feat(truncated-list): create component

### LINKED ISSUE #468: Request - Filter panel checkbox truncation component
### Details

- **Application / team -** Data explorer

### Proposal

The purpose of having a customizable filter panel is to give consumers the freedom to create their own filter panel structure. I've uncovered that we need the ability to truncate filters in components other than the `FilterPanelAccordionItem`. 

What I'm requesting/proposing is the creation of a filter truncator component that we can use in other components as shown in the data sources section here: 

### Goals / use cases

- [ ] Allow consumers to truncate filters in components other than the `FilterPanelAccordionItem`
- [ ] Could this be a general-purpose component?

diego-codes:
## Affected issues

- Resolves #468

## Proposed changes

- Create a `TruncatedList` component
- Clean up some content truncation logic
- Replace the list truncation logic in the `FilterPanelAccordionItem` component with the `TruncatedList` component

## Testing instructions

- Review the new `TruncatedList` component in Storybook.
- Review the documentation for the `TruncatedList` component.
- Review changes in the `FilterPanel` related to the new component


diego-codes:
Thanks to any reviewers out there! In addition to reviewing the code, I wanted to ask how you all feel about:

1. Removing the truncation functionality from the `FilterPanelAccordionItem` so consumers can choose if they want to truncate or not. I realize that this might cause disruptions to consumers of this component. Is there a way that you would recommend making this change happen? Maybe add a prop like `truncateItems` that is set to `true` by default and is deprecated? This way consumers are alerted that they will need to use the `TruncatedList` to keep the same functionality?

2. Do the three props `truncateThreshold`, `collapsedItemLimit`, `expandedItemLimit` make sense? Does the documentation provide enough context?

3. Does the simplification of the expand button styles make sense? It deviates from the original design, but I think it's important to not modify it too much if we expect this component to be usable in other contexts.

src/components/TruncatedList/TruncatedList.stories.js
```diff
@@ -0,0 +1,59 @@
...
+import { storiesOf } from '@storybook/react';
+import { number, select, color, text } from '@storybook/addon-knobs';
+import { components } from '../../../.storybook';
+import TruncatedList from '.';
+import { createChildrenArray } from './_mocks_';
+
+storiesOf(components(TruncatedList.name), module).add(
+  'default',
+  () => (
+    <TruncatedList
```
jendowns, line 16: Could you wrap this in a demo container with a set width to avoid the jumpiness in the demo?

src/components/TruncatedList/TruncatedList.stories.js
```diff
@@ -0,0 +1,59 @@
...
+    </TruncatedList>
+  ),
+  {
+    info: {
+      text: `## Truncated list
+This component is a list that accepts any number of list items (li). It allows consumers to control
+how many items are revealed to the user while giving the user the ability to expand the list and see
+the entire list. There are three props that control how many items a user can see at once:
+
+  * \`truncateThreshold\`: This prop will give consumers control of how many items are rendered
```
jendowns, line 48: I love the formatting you did here but unfortunately the rendered `ul`, `li`, etc elements are not styled in storybook: 

Screen Shot 2020-04-07 at 9 18 00 AM

There are a couple ways to get around this, I think 🤔 One option is to just add some storybook/demo-specific styling, so that the lists are formatted? Or you could not use markdown here and try to apply a `List` component.

src/components/TruncatedList/__tests__/TruncatedList.spec.js
```diff
@@ -0,0 +1,159 @@
...
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import TruncatedList from '..';
+import OrderedList from '../../OrderedList';
+import { createChildrenArray } from '../_mocks_';
+
+describe(TruncatedList.name, () => {
```
jendowns, line 14: Nice! 🤩 I feel like I learn so much from your tests

diego-codes:
Thank you @petervachon and @jendowns for the feedback! I'm going to look at the comments today and address the changes requested.

src/components/TruncatedList/__tests__/TruncatedList.spec.js
diego-codes: 😅

src/components/TruncatedList/TruncatedList.stories.js
diego-codes: So um I don't know what happened, but the list styles got all different since I last looked at this. Suddenly all lists now have bullets and padding, which resolved this comment. It broke some of the other components related to the `TruncatedList`, but I fixed them.

diego-codes:
Hi, @jendowns and @petervachon! Sorry for the delay. I've addressed most of your comments. I think you'll find that the new style of the button looks much cleaner in the context of both the stand-alone `TruncatedList` and filter panel. Please let me know what you think.

The only piece of feedback I didn't fully address was whether we include an icon or not. I agree that removing them cleans up the UI. I also see Jen's point to keep the icon consistent in the filter panel. However, I think it would be better to provide some constraints and a point-of-view for all consumers. Basically, we should agree to have or remove the icons.

petervachon:
Looking good! 

In regards to the icon, I believe in the last revision you did the icon was on the right. In the current implementation the, in the filters component, the + icon is on the left and aligned with the checkboxes, is there any specific reason you change it to be on the right?

diego-codes:
Thanks for the prompt response, @petervachon! Let me try to recount my thought process if you will.

The design for the filter panel predated the decision position icons to the right of the text in a button. We did not have any precedence on what made sense for the system. At the time, aligning the icon to the checkbox and their text was preferred. 

Fast forward to now, we have a more standardized pattern to have icons in buttons to the right of the text. Additionally, the idea of the `TruncatedList` component has come out of necessity to have this functionality to apply to other filter layouts as outlined in #468 and in other non-filter panel contexts. I remember seeing designs that could have benefitted from this component. 

Ignoring the standard button with icon and text pattern does not seem ideal for consistency. Also getting the icon and text to align perfectly with the filter checkbox and text is so specific that the button would not work as well in other contexts where the list is not a checkbox. 

This is what it looks like when we align the icons and text:


This is what it looks like when these alignment adjustments apply to non-checkbox lists:


I am proposing that we either remove the icon altogether, which would look like this:
 

Or go in with icons, but keep the same order as we do with the other buttons:
 

What are your thoughts on the options I've outlined?

petervachon:
Had a quick review with Cameron and Jimmy, they are in favor of taking out the icon altogether.

diego-codes:
Thank you guys for the quick feedback! I've removed the icon. This PR is ready for a design and code review.

PR #480, opened 2020-04-08
Allow link to open in new tab on error page

### LINKED ISSUE #477: Allow error page link to open in new tab
Allow error page link to open in new tab

clairek20:
## Affected issues

- Resolves #477

## Proposed changes

Add prop 'target' to allow link to open in new tab


src/components/ErrorPage/ErrorPage.js
```diff
@@ -87,6 +87,7 @@ const ErrorPage = ({
                 id={id}
                 key={id}
                 href={href}
+                target="_blank"
```
jendowns, line 90:
```suggestion
                target="_blank"
                rel="noopener noreferrer"
```

SimonFinney:
@jendowns I suspect so as my assumption is there would be internal links included as part of this.

@clairek20 What's the specific use case this addresses? Would you have an example of the kinds of links that would be applied here?

clairek20:
Hi

> @jendowns I suspect so as my assumption is there would be internal links included as part of this.
> 
> @clairek20 What's the specific use case this addresses? Would you have an example of the kinds of links that would be applied here?

Hi @SimonFinney , so most of the error pages just want to direct the user to IBM support, so the link would be https://www.ibm.com/mysupport/s/topic/0TO0z000000RcPxGAK/cloud-pak-for-security

SimonFinney:
Ah, if so then there's no need for additional configuration! I wasn't entirely certain if that was the case

clairek20:
Great stuff thanks @SimonFinney , i don't have the ability to merge it, is this on ye?

SimonFinney:
@clairek20 I noticed after checking on the cluster that there's also an internal link for linking back to the homepage - will this continue to be the case?



If not, it would be ideal to allow consumers to specify whether a link should be external to maintain usability and avoid breaking existing functionality

clairek20:
> @clairek20 I noticed after checking on the cluster that there's also an internal link for linking back to the homepage - will this continue to be the case?
> 
> 
> 
> If so, it would be ideal to allow consumers to specify whether a link should be external to maintain usability and avoid breaking existing functionality

Ah ya that would make more sense, thanks for spotting, i didn't realise we had a link that did that! What would be the best way to implement that? Have you an example of making something optional? @SimonFinney

SimonFinney:
You could add an `external` prop to the `link` propTypes declaration, defaulting to `false`, and apply the external link attributes you've already proposed conditionally depending on the value of `external`

clairek20:
Hi @SimonFinney 

I have the latest up there now in regards to making the link external. Let me know if this is okay!

Thanks, Claire.

clairek20:
> LGTM - thanks! 💯

Great thanks! Would you mind merging? I don't have the permissions

PR #486, opened 2020-04-13
feat(filter-panel-checkbox-with-overflow-menu): create component

diego-codes:
## Proposed changes

- Implement the `FilterPanelCheckboxWithOverflowMenu` component that will be used for not-filters

## Testing instructions

- Review the story for the `FilterPanelCheckboxWithOverflowMenu`


diego-codes:
Thank you all for reviewing this PR! First, I have to say that I am not married at all to any of the details of the code, especially the name and the API. I've been debating whether or not this functionality should be its own component or just added functionality to the existing checkbox component. Additionally, if you think of a better way to give consumers better control over the overflow menu, please let me know.

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
```diff
@@ -0,0 +1,119 @@
...
+  overflowMenuClassName,
+  overflowMenuOptionsClassName,
+  children,
+  ...other
+}) => {
+  const [showOverflowButton, setShowOverflowButton] = React.useState(false);
+  const [onBlurTimeout, setOnBlurTimeout] = React.useState(undefined);
+
+  // Clear any timeouts before unmounting.
+  React.useEffect(() => () => clearTimeout(onBlurTimeout), [onBlurTimeout]);
```
areddon, line 33: To do a cleanup with hooks, simulating unmounting, your `useEffect` would return your cleanup function. 
[React docs](https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup) for effects with cleanup.

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
```diff
@@ -0,0 +1,119 @@
...
+  const menuOptionsClassName = `${namespace}__overflow-options`;
+
+  // Set a short delay to detect if the blur event resulted in a complete loss of keyboard focus or
+  // the focus just shifted to another element within this same component.
+  const handleBlur = event => {
```
ShayMurnin, line 38: Hey @diego-codes! I've implemented a very similar component on Falcon and run into this same issue. I used something like this snippet:

```
  let timeout = null;

  const handleFocus = () => {
    clearTimeout(timeout);
    if (!setShowOverflowButton) {
      setShowOverflowButton(true);
    }
  };

  const handleBlur = () => {
    timeout = setTimeout(() => {
      if (showOverflowButton) {
        setShowOverflowButton(false);
      }
    }, 10);
  };
```

It basically relies on the `onFocus` event from the next descendant receiving focus to bubble up straight after `onBlur` and clear the `handleBlur` timeout before it can execute. It's a bit less code  and no use of `document` but also relies on the descendants allowing it's `onFocus` event to propagate up so perhaps your method is more robust!

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/__tests__/FilterPanelCheckboxWithOverflowMenu.spec.js
```diff
@@ -0,0 +1,260 @@
...
+      }
+    );
+
+    // Open overflow menu and waitFor for options to appear on the DOM.
+    fireEvent.mouseEnter(screen.getByLabelText(/checkbox label/i));
+    userEvent.click(screen.getByLabelText(/filter selection options/i));
+    await screen.findByText(/option 1/i);
+
+    await expect(document.body).toHaveNoAxeViolations();
+    await expect(document.body).toHaveNoDAPViolations(
```
diego-codes, line 60: @jendowns, I wonder if you have any insights on this. This test is failing and I _think_ it has to do with the `OverflowMenu` implementation from Carbon that is maybe the cause?

areddon:
@diego-codes 
I thought the design was to have the overflow menu items be the same width as the component itself?

This probably doesn't work great inside the search, what if the component grew in height to show the options instead of using the overflow menu?

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
diego-codes: Hi @ShayMurnin! Thanks for the code snippet. I like it! What elements did you apply the `handleFocus` and `handleBlur` handlers to? I'm wondering because the issue I ran into was that the overflow menu options container does not share the same DOM hierarchy as the overflow menu button.

diego-codes: Okay, I got your snippet to work!! And it makes a good candidate for a hook that can also be used for the `FilterPanelSearch`. Thank you so much, @ShayMurnin! Super clean approach ✨

diego-codes:
@SimonFinney I added a custom hook in this PR. Please let me know if you think there is a better file or structure to place a custom hook!

diego-codes:
This PR is ready for another review. I will try to address the menu options width to match the checkboxes tomorrow.

SimonFinney:
Thanks for your patience @diego-codes ! I have an outstanding task to submit a review for this on Friday

package.json
```diff
@@ -101,9 +101,9 @@
     "@storybook/addons": "^5.2.4",
     "@storybook/react": "^5.2.4",
     "@storybook/theming": "^5.2.4",
-    "@testing-library/jest-dom": "4.2.4",
-    "@testing-library/react": "9.3.2",
-    "@testing-library/user-event": "8.1.0",
+    "@testing-library/jest-dom": "^5.5.0",
```
SimonFinney, line 104: @jendowns I remember you initially locked the versions of React Testing Library here - could you provide insight as to whether updating would cause any impact on the rest of the testing suite?

package.json
```diff
@@ -101,9 +101,9 @@
     "@storybook/addons": "^5.2.4",
```
SimonFinney, line 101: Are all of the offline dependency changes a result of the testing package updates?

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/_index.scss
```diff
@@ -0,0 +1,43 @@
+////
+/// Filter panel checkbox with overflow menu styles.
+/// @group filter-panel
+/// @copyright IBM Security 2020
+////
+
+@import '../../Component/mixins';
```
SimonFinney, line 7: Would you also mind including spacing and tokens imports here? This will silently fail because they may have been included elsewhere in the Sass import order.

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
```diff
@@ -0,0 +1,152 @@
...
+};
+
+FilterPanelCheckboxWithOverflowMenu.propTypes = {
+  ...FilterPanelCheckbox.propTypes,
+
+  /** Optional class name. */
+  className: PropTypes.string,
+
+  /** Optional checkbox class name. */
+  checkboxClassName: PropTypes.string,
```
SimonFinney, line 123: I'm curious about `checkboxClassName`, `checkboxWrapperClassName`, and `overflowMenuClassName` - these are definitely nice to have for additional customization, but do you foresee use cases in which this might be desired?

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Sets the width of the overflow menu to match the width of this component's width and adjusts
+   * its position so it is right alingned with the overflow menu button.
+   *
+   * @param {HTMLElement} menuBody Element containing the overflow menu options.
+   * @param {string} direction The menu position relative to the menu button. Not used.
+   * @param {HTMLElement} menuButton Overflow menu button element.
+   * @returns {{top: number, left: number}} The menu offset measurements.
+   */
+  const updateMenuWidthAndSetOffset = (menuBody, direction, menuButton) => {
```
SimonFinney, line 52: 👏

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanel.stories.js
```diff
@@ -0,0 +1,66 @@
...
+import { action } from '@storybook/addon-actions';
+import { text, number } from '@storybook/addon-knobs';
+
+import React from 'react';
+
+import { patterns } from '../../../../.storybook';
+
+import FilterPanelCheckboxWithOverflowMenu from '.';
+import FilterPanelSearch from '../FilterPanelSearch';
+import OverflowMenuItem from '../../OverflowMenuItem';
```
SimonFinney, line 16: As an extra measure for verifying the public API, we've been importing components for stories directly from `src/index.js` - For example, in this case it might be `import { FilterPanelCheckboxWithOverflowMenu, FilterPanelSearch, OverflowMenuItem } from '../../..'`

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
```diff
@@ -0,0 +1,152 @@
...
+        >
+          {children}
+        </OverflowMenu>
+      )}
+    </div>
+  );
+};
+
+FilterPanelCheckboxWithOverflowMenu.propTypes = {
+  ...FilterPanelCheckbox.propTypes,
```
SimonFinney, line 117: 👏 I like this because we can maintain parity with the `FilterPanelCheckbox` in case that API changes!

The only thing to note is that Storybook's documentation generation doesn't like this so much. 😭 We could potentially fill in the gaps by linking out to the `FilterPanelCheckbox`stories from the `FilterPanelCheckboxWithOverflowMenu`, similarly to https://ibm-security.carbondesignsystem.com/?path=/story/components-accordion--default 's 'Show Info'.

src/globals/utils/focus.js
```diff
@@ -30,3 +32,47 @@ export const trapTabFocus = (element, event) => {
...
+ * hook is useful when you want to allow users to shift keyboard focus inside component and only
+ * apply blur effects when the keyboard focus has completely left this component.
+ *
+ * @param {number} [timeoutDelay=10] Millisecond delay for the blur effect to take place. This
+ * should not be less than 10 to give the browser time to register the subsequent focus event first.
+ *
+ * @returns {{createFocusHandler: function, createBlurHandler: function }} Utility functions to
+ * create the focus and blur handlers that apply the necessary effects or state changes.
+ */
+export const useComponentFocus = (timeoutDelay = 10) => {
```
SimonFinney, line 47: 👏

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
diego-codes: That's a good question. I don't foresee that I'll be using them. Should I remove them?

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
diego-codes: Ah yes, I can do that.

package.json
jendowns: @SimonFinney @diego-codes Digging into the git history here, I found this comment from January: https://github.com/carbon-design-system/ibm-security/pull/337#discussion_r373161223

It looks like I was having issues where CircleCI was using a slightly older version of `user-event` that was causing my tests to fail in the pipeline, but locally everything passed because I had the updated version. 🤷‍♀️

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
jendowns: The biggest use case here is if someone wants to target a specific element, either for testing or their own customizations. Consider that they won't necessarily add an id to any of these inner elements -- they would need to reply on these class name props.

I vote that we keep them. IMO they definitely fill a need that we can't necessarily address otherwise.

package.json
jendowns: It's possible since many of these package updates are major versions, including the `jest` core library.

diego-codes:
I don't think the issue is the memory heap allocation. I think that the second a11y test is failing and I believe it is because the overflow menu button does not meet some a11y criteria? 

This is the first violation that is listed on the page:
>  Message: Accessible name does not match or contain the visible label text.

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.js
diego-codes: I think it's fair to remove them if we don't have an immediate need for it. It is easier to add customization later than remove it. I find that smaller component APIs are easier to read and make use of. For testing, consumers should not be relying on class names and there are ways to customize styles if necessary that don't require us to provide a prop for every single possible classname.

package.json
diego-codes: Yeah, I don't remember why I did this... I will rever this because I didn't need it.

diego-codes:
FINALLY! Got the build to pass. @jendowns and @SimonFinney this PR is ready to review again. Thank you.

diego-codes:
Sounds good! Thanks, @jendowns! Also, does anyone have a better name for this component? Just thought I'd ask one more time.

.storybook/index.js
```diff
@@ -59,14 +59,14 @@ const disableCentered = stories =>
...
  * @returns {Object<string, string>} The configuration containing information to apply.
  */
-const info = (description, carbon = null) => ({
-  info: carbon
+const info = (description, component = { library: 'react' }) => ({
+  info: component.story
     ? `${description}
 
-Also refer to http://react.carbondesignsystem.com/?path=/story/${carbon.story}--${carbon.id}`
+Also refer to http://${component.library}.carbondesignsystem.com/?path=/story/${component.story}--${component.id}`
```
SimonFinney, line 69: I'm good with this change, but it's no longer working as expected in existing implementations of `Accordion` and `UIShell`:

Accordion

diego-codes:
Thanks for catching that, @SimonFinney! This PR is ready for review.

ShayMurnin:
LGTM @SimonFinney — nice work @diego-codes 🙌

PR #488, opened 2020-04-15
docs(storybook): update storybook sidebar with package info and version

### LINKED ISSUE #458: Request: Include the package version in the Storybook title
### Proposal

It appears that `carbon-components-react` now includes the version of the package the Storybook is showcasing. It would be great if this package did the same!

image

### Goals / use cases

- [ ] Easier for consumers to know what version documentation we are looking at
- [ ] Better transparency

jendowns:
## Affected issues

- Resolves #458

## Proposed changes

(look at top left -- where there's the lockup image etc.)

- move lockup image to `brandName` and also include package info + name
- remove `brandImage` (because it visually overrides `brandName`)

## Testing instructions

Keep in mind that this is how Carbon's storybook looks (see top left): https://react.carbondesignsystem.com

How similar do we want to be?
Should we keep the lockup image?
Should the package name + version be written differently?

### DIFF
--- a/.storybook/theme/index.js
+++ b/.storybook/theme/index.js
@@ -9,7 +9,7 @@ import { fontFamilies } from '@carbon/type/lib';
 
 import { create } from '@storybook/theming';
 
-import { name, homepage } from '../../package.json';
+import { name, homepage, version } from '../../package.json';
 
 const { mono, sans } = fontFamilies;
 const { field02, inverse01, activeUI, text01, ui01, ui02 } = g100;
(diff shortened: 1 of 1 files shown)


diego-codes:
I like it. I wonder if wrapping the text in `<code>` to make it monospace would be sense?

jendowns:
Good idea @diego-codes -- I just wrapped the package info in `code` tags and IMO it does look better with the monospace type 👍

SimonFinney:
@jendowns A note on this one - Netlify currently only triggers the Storybook deploy from commits to `dev` and `master`, not tags. When a commit triggers a tag, publishes a package, and version bumps back to Git, a new Storybook deploy will not be spun up resulting in the previous version number always being displayed in the latest Storybook deploy.

We won't observe this in https://dev--ibm-security.netlify.app just yet as e98a9d2073329184e6dd83efe76f0c3d19b03318 is documentation-related and won't bump the version, but it's worth noting that this will be an issue going forward.

PR #492, opened 2020-04-17
fix(themes): add token ignore list for CSS custom properties

SimonFinney:
## Proposed changes

Generate an ignore list for CSS custom properties theme tokens to prevent spacing and typography tokens from being converted


SimonFinney:
@jendowns For Carbon's spacing values, they don't expose the token names as part of their exports so initially, the blacklist seemed like the best option:

Spacing

 However, I prefer the idea of filtering out by color and have made revisions to accommodate that.

src/globals/theme/_index.scss
```diff
@@ -10,10 +10,12 @@
...
 @import '../feature-flags/index';
 
-@import 'functions';
 @import 'variables';
 
-// TODO: `2.x` - To be removed.
+// Theme blacklist.
+/// @type List<String>
+/// @access private
+$security--theme__blacklist: ();
```
jendowns, line 18: Could you please rename this? I suggest a "blocklist", "exclusionary list", "list of excluded variables", etc.

PR #495, opened 2020-04-20
feat(button): permit disable toggle for `loading`

### LINKED ISSUE #392: Button loading prop should not disable button
## Request

The `loading` prop on the `Button` component disables the button. I don't believe that this should be the default behavior. Let the consumer decide whether to disable the button or not. If they already pass in a prop to set `loading`, they can use the same one to disable if they want.

Not sure if it is a bug or desired behavior, but when setting `loading` to true, the button changes to a ghost button, regardless of initial type.

SimonFinney:
## Affected issues

Resolves #392

## Proposed changes

- Accommodate toggling both `disabled` and `loading` for buttons simultaneously
- Enforce disabling button for `SummaryCardAction` to maintain existing design


src/components/Button/Button.js
```diff
@@ -66,9 +63,6 @@ Button.defaultProps = {
 Button.propTypes = {
   ...propTypes,
 
-  /** @type {boolean} Whether or not the button is disabled. */
-  disabled: PropTypes.bool,
-
   /** @type {boolean} Whether or not the button is in a loading state. While loading, the button is disabled & icons provided via the `renderIcon` prop will not be shown. */
```
jendowns, line 66: This prop description will need to be updated 👍

SimonFinney:
@jendowns Good catch! Because we're now allowing consumers to determine whether a loading button should be disabled or not, we likely now need a visual indicator to differentiate between the two variations.

Reverting to Carbon's default behavior for these styles seemed like a reasonable approach, but I'd like to @cameroncalder or @petervachon to chime in if we require additional consideration here.

jendowns:
@SimonFinney The origin designs had the `SummaryCard`'s loading button with the blue60 color text, I believe. For the normal button, I don't necessarily have an opinion about that. But for that `SummaryCard`, it's a design regression. 👍

src/components/Button/Button.js
SimonFinney: What do you mean here?

PR #504, opened 2020-04-22
feat(modular-layouts): add demo with generated spacing & debugging classes

jendowns:
## Proposed changes

- add a large storybook demo for the "details page layout", with the ability to toggle "debugging" on/off with knobs and also show/hide various sections of the layout
- generate classes with styles and also debugging styles for each spacing token from Carbon

⚠️ **NOTE: This is for top and bottom padding only**. I have not generated any classes or debugging styles for left or right padding.

## Testing instructions

- Naming?
  - What should the demo be called?
  - What should the mixins/functions to generating spacing classes & debugging styles be called?

- Show/hide?
  - Is it worthwhile to show/hide pieces of this UI if in almost all cases, a whole column + contents are just taken out?


src/components/Layouts/Layouts.stories.js
```diff
@@ -0,0 +1,248 @@
...
+} from '../..';
+
+import { disableCentered, patterns } from '../../../.storybook';
+
+const toggleDebugging = () => boolean('Toggle debugging 🕵️‍♀️', false);
+
+const showAccordion = () => boolean('Show accordion', true);
+const showButtonGroup = () => boolean('Show button group', true);
+const showCardGroup = () => boolean('Show card group', true);
+const showTabGroup = () => boolean('Show tab group', true);
```
jendowns, line 32: Some of these don't feel particular valuable to me. With the except of `showTabGroup`, which ends up changing the column span and therefore modifies the layout, toggling on/off these UI groups just removes a column. 🤔

src/components/Layouts/Layouts.stories.js
```diff
@@ -0,0 +1,248 @@
...
+
+const showFirstRow = showCardGroup() || showButtonGroup();
+const showSecondRow = showTabGroup() || showAccordion();
+
+disableCentered(storiesOf(patterns('Modular Layouts'), module)).add(
+  'details page layout',
+  () => (
+    <main className={toggleDebugging() && 'debug'}>
```
jendowns, line 40: I realize this story is VERY verbose but it is nice to have the demo like this so that people can just copy + paste. If this code is abstracted away, then storybook won't display the inner components in their entirety.

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
+////
+/// Modular layouts styles.
```
jendowns, line 2: For any/all mixins, classes, and functions in this file, I'm very open to renaming 😆

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+  div[class^='bx--col'] {
+    outline: 1px solid $carbon--orange-40;
+  }
+
+  @include debug-spacing-top;
+  @include debug-spacing-bottom;
+}
+
+@include spacing-top;
+@include spacing-bottom;
```
jendowns, line 104: Note that I'm only generating styles for top and bottom padding, not left/right.

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+    .spacing--#{$token}--bottom::after {
+      @include debug;
+      content: 'spacing--#{$token}--bottom';
+      height: $value;
+      bottom: 0;
+    }
+  }
+}
+
+.debug {
```
jendowns, line 94: Probably should be renamed 😆 
`security--spacing-debugging`? 🤷‍♀️

src/components/Layouts/Layouts.stories.js
```diff
@@ -0,0 +1,248 @@
...
+const showAccordion = () => boolean('Show accordion', true);
+const showButtonGroup = () => boolean('Show button group', true);
+const showCardGroup = () => boolean('Show card group', true);
+const showTabGroup = () => boolean('Show tab group', true);
+
+const showFirstRow = showCardGroup() || showButtonGroup();
+const showSecondRow = showTabGroup() || showAccordion();
+
+disableCentered(storiesOf(patterns('Modular Layouts'), module)).add(
+  'details page layout',
```
jendowns, line 38: The idea here is that each layout would be their own variation 🤔

src/components/_index.scss
```diff
@@ -91,3 +91,4 @@
 
 /// Unstable experimental components.
 @import 'UNSTABLE__Pagination/index';
+@import 'Layouts/index';
```
SimonFinney, line 94: Should we follow the same `UNSTABLE__*` naming convention here?

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+
+@import '@carbon/colors/scss/colors';
+@import '@carbon/layout/scss/spacing';
+@import '@carbon/themes/scss/theme-maps';
+
+/// Generate a map of spacing tokens from the provided Carbon theme map.
+/// @access private
+/// @param {Map} $theme-map - Map of theme variables.
+/// @returns {Map} - The filtered map of only spacing variables.
+@function get-spacing-tokens-map($theme-map) {
```
SimonFinney, line 15: Because only [Dart Sass supports `@use`](https://sass-lang.com/documentation/at-rules/use), we should practice namespacing any publicly accessible functions, mixins, and variables (as an extra precaution) to avoid potential conflicts with consumers' own Sass builds - A previous example of this was with the `component` mixin, in which our implementation ended up overwriting the consumers and caused errors in their compilation.

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+
+  @return $-map;
+}
+
+/// Map of spacing tokens.
+/// @type Map<String, Number>
+/// @access private
+$spacing-tokens-map: get-spacing-tokens-map($carbon--theme--g100);
```
SimonFinney, line 35: Unless we foresee `get-spacing-tokens-map` being used elsewhere, I recommend calling it inline to promote the practice of only creating functions in situations where code is duplicated, with the exception of if the piece is being tested.

I remember you mentioning you'd like to introduce some tests for `get-spacing-tokens-map` to ensure that the correct values are being returned - is this still the case?

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+
+// Generate styles for the debugging message.
+@mixin debug {
+  @include carbon--type-style($name: caption-01);
+  position: absolute;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding: $spacing-02;
```
SimonFinney, line 44:
```suggestion
  padding: $carbon--spacing-02;
```

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+  @return $-map;
+}
+
+/// Map of spacing tokens.
+/// @type Map<String, Number>
+/// @access private
+$spacing-tokens-map: get-spacing-tokens-map($carbon--theme--g100);
+
+// Generate styles for the debugging message.
+@mixin debug {
```
SimonFinney, line 38: 👏

src/components/Layouts/_index.scss
SimonFinney: For anything that outputs CSS as part of the Sass compilation, it's important we wrap it in either the `export-namespace` or `security--component` mixins to safeguard the styles from being compiled multiple times consumers' output.

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+  color: $carbon--black-100;
+  background-color: $carbon--magenta-40;
+}
+
+// Generate "top" padding classes and styles for each spacing token.
+@mixin spacing-top {
+  @each $token, $value in $spacing-tokens-map {
+    .spacing--#{$token}--top {
+      padding-top: $value;
+      position: relative;
```
SimonFinney, line 55: Would you mind clarifying the purpose of the `position` property here? If it's to assist with debugging, I recommend including it as part of the `debug-spacing-*` mixins to ensure it doesn't introduce any unintended side effects.

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,104 @@
...
+@mixin debug-spacing-bottom {
+  @each $token, $value in $spacing-tokens-map {
+    .spacing--#{$token}--bottom::after {
+      @include debug;
+      content: 'spacing--#{$token}--bottom';
+      height: $value;
+      bottom: 0;
+    }
+  }
+}
```
SimonFinney, line 92: Given there are multiple `@each` directives here on the same map with similar properties, it feels like an opportunity for a mild refactor.

SimonFinney: I'm also curious whether you foresee these mixins being leveraged for other use cases?

src/components/_index.scss
jendowns: I’m not sure why this would be considered unstable though? It isn’t a component, and the classes are generated from existing spacing variables (which are unchanged).

src/components/Layouts/_index.scss
jendowns: Yes, I realize this. I made a couple comments throughout asking for guidance on naming. 🙂 I’m hoping for naming suggestions here and elsewhere. Thank you.

src/components/Layouts/_index.scss
jendowns: Relative positioning needs to be applied so that the absolutely-positioned pseudo elements can appear in the correct spot.

This is related to debugging though so I can move it there. 👍

src/components/Layouts/_index.scss
jendowns: Yes these can likely be combined. 👍

But no I don’t see how these could be used outside of this context.

src/components/Layouts/_index.scss
jendowns: Sure thing. Do you have any suggestions on naming?

src/components/Layouts/Layouts.stories.js
```diff
@@ -0,0 +1,269 @@
...
+  () => (
+    <main className={toggleDebugging() && 'security--debug'}>
+      <h1 className="bx--type-productive-heading-04">productive-heading-04</h1>
+      <Grid condensed>
+        {showFirstRow && (
+          <Row>
+            {showCardGroup() && (
+              <Column md={3} lg={6}>
+                <h2
+                  style={{ margin: 0 }}
```
jendowns, line 48: @SimonFinney related to your review comment: https://github.com/carbon-design-system/ibm-security/pull/504#pullrequestreview-402513118

Not sure the best say to go about resetting these `h2` margin defaults... 

Should we start exporting some heading reset styles (just resetting margins and padding, before applying specific padding via these generated classes)? Seems like the direction to go... if the end goal is to have folks apply these layouts + classes as-is, then we've want to ensure that the headings are styled correctly. 🤔

src/components/_index.scss
SimonFinney: Good question! Even though these aren't being distributed as formal 'components', all of the CSS classes outputted will still have to be considered part of our public API - I suggest marking it as unstable until we have the majority of examples tested and available

src/components/Layouts/_index.scss
SimonFinney: Personally, `security--layout` makes sense to me as a top-level namespace - my only fear is that this may cause some lengthier class names further down the chain and when we introduce modules like the card and ICA.

src/components/Layouts/Layouts.stories.js
SimonFinney: There are pros and cons to each approach - reset utility classes would provide additional control but an extra cognitive step for consumers and potentially more room for user error.

An alternative could be to use margins instead of padding for the layout classes, but I don't fully understand what the implications of that would have on the larger system you're proposing.

I suspect @cameroncalder could provide insight and work through some of these details that are arising as we understand it better.

jendowns:
Hey @SimonFinney this is ready to review again. I moved the debugging styles to the storybook stylesheet so that way they don't get exported with the library's styles, for now. I added margin classes as well (though worth noting that the debugging styles currently only apply to padding classes).

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,67 @@
+////
+/// Modular layouts styles.
+/// @group modular-layouts
+/// @copyright IBM Security 2020
+////
+
+@import '@carbon/colors/scss/colors';
```
SimonFinney, line 7:
```suggestion
```

src/components/Layouts/Layouts.stories.js
```diff
@@ -0,0 +1,208 @@
...
+  'details page layout',
+  () => (
+    <main className={toggleDebugging() && 'security--debug--padding'}>
+      <h1 className="bx--type-productive-heading-04">productive-heading-04</h1>
+      <Grid condensed>
+        {showFirstRow && (
+          <Row>
+            {showCardGroup() && (
+              <Column md={3} lg={6}>
+                <h2 className="bx--type-productive-heading-01 security--padding-top--layout-01 security--padding-bottom--layout-03 security--margin-top--layout-00 security--margin-bottom--layout-00">
```
SimonFinney, line 47: This makes sense as we discussed, but seeing it in written out in code, it starts to become a fairly lengthy step for consumers? 🙀

If the 'spacing' for titles is always going to be this combination, it might be beneficial to roll these up into a single class for consumers to apply (with the exception of `bx--type-productive-heading-01`)? This is more a question to consider for the stories we're picking up this sprint, and I don't think we'll have the right answer until we work through them!

src/components/_index.scss
SimonFinney: @jendowns I'm not sure if you saw this, but I feel it's worth marking these as experimental while this code has been merged into `dev` and we work through the stories for this sprint

src/components/Layouts/Layouts.stories.js
jendowns: @SimonFinney Yes they can be lengthy, but I would caution against "rolling them up" because it would get us back to an earlier iteration where we were over-abstracting too much. We can't, and shouldn't, always assume a certain combination of top + bottom padding here. There could be a number of iterations -- and how would the naming convention work for those?

This is lengthy but it is readable, understandable, customizable. 👍 

Now there IS an opportunity to create top + bottom and even "block" padding classes for *the same padding value* vertically or all around but that wouldn't change what you see here.

src/components/_index.scss
jendowns: @SimonFinney I still stand by my original comment, especially now that the generated CSS has been greatly reduced to only output single padding and margin style rules. I don't see anything experimental about those classes. Once the naming convention for the styles is finished when this PR is merged, it should be considered a stable feature like any other feature.

To that end, I was curious: what do you expect to see change about the layout tokens, or padding rules, that would warrant marking them unstable?

src/components/Layouts/_index.scss
```diff
@@ -0,0 +1,68 @@
...
+/// Generate a map of 'layout' spacing tokens from the provided Carbon theme map.
+/// @access private
+/// @param {Map} $theme-map - Map of theme variables.
+/// @returns {Map} - The filtered map of only spacing variables.
+@function security--get-spacing-tokens-map($theme-map) {
+  $-map: ();
+
+  @each $token, $value in $theme-map {
+    @if type-of($value) == 'number' {
+      @if str-index($token, 'layout-') {
```
jendowns, line 21: @SimonFinney I added an extra change here because I noticed that we were getting a number of class names generated that we probably would never realistically used, like for icon spacing for larger "container" spacing. If we want to expand the scope, we could update this check in the future. But we now, I think we should keep the scope small to just `layout-` tokens.

src/components/Layouts/Layouts.stories.js
SimonFinney: We need to continue to be mindful of the pros and cons of abstraction - the previous iteration you're mentioning was one avenue of exploration and learning, and the outcomes of that, good and bad, should not necessarily be disregarded outright as it helped better tease out and inform the design concept with @cameroncalder . A combination of techniques should still be considered to inform ease-of-implementation by our consumers.

As you've mentioned, we've learned that it might not make sense to abstract out grid columns, because that's inflexible, but I don't believe that means we should potentially irritate all of our consumers to apply a string of classes à la `security--padding-top--layout-01 security--padding-bottom--layout-03 security--margin-top--layout-00 security--margin-bottom--layout-00 security--etc` when perhaps all they want is one thing that does it for them.

As I've said in the past, abstraction is a delicate balance and used everywhere - I encourage you to look at examples of where this is useful (see Carbon's `Grid` component), and also to wear the shoes of our consumers to understand how it can be beneficial, especially as we review the specifications for the layout modules and discover that recurring patterns do exist and there are implementation details that they should not care about.

PR #508, opened 2020-04-24
fix(stacked-notification): improve proptype info and clear proptype errors

jendowns:
A proptype error for `statusIconDescription` was generating noise in tests. I went ahead and added that prop here, as well as made some improvements to proptype documentation. Most descriptions are pulled directly from Carbon's `ToastNotification` proptypes: https://react.carbondesignsystem.com/?path=/story/notifications--toast

## Proposed changes

- include `statusIconDescription` in tests
- add `statusIconDescription` to component WITH a default value (NOTE: Carbon does not have a default value for this prop, but they also don't require it. So I created a default based on what made sense. Definitely willing to adjust 👍 )
- add proptype descriptions


jendowns:
@SimonFinney or @diego-codes could one of you take a look at this when you get a chance? The change is a11y related: I've added a new `statusIconDescription` prop (which is required in the underlying Carbon component) & updated prop documentation. Please let me know if you have any questions -- thank you!

PR #512, opened 2020-04-27
fix(tagwall): accommodate truncated text in tags and filter list items

### LINKED ISSUE #511: TagWallFilter: long tags are not truncated & overflow their container
**Expected behavior -**

A long tag, either in the top filter area or in the bottom list area, should be truncated with an ellipsis.

**Actual behavior -**

Very long tags, in both the top filter area and the bottom list area, are not truncated and overflow their containers.

### Steps for reproducing

Check out the storybook & add very long tags:

jendowns:
## Affected issues

- Resolves #511

## Proposed changes

- use Carbon's filterable `Tag` instead of custom `InteractiveTag`
- apply a few color overrides to the `Tag`
- use Carbon's truncation class to truncate items in the filter options list

## Testing instructions

- check out the `TagWall` + `TagWallFilter` storybook demos, inspect a tag or filter option, and change it to a very long entry that needs to be truncated.

### DIFF
--- a/src/components/TagWall/TagWall.js
+++ b/src/components/TagWall/TagWall.js
@@ -13,7 +13,7 @@ import * as defaultLabels from '../../globals/nls';
 import defaultItemToString from '../__tools__/defaultItemToString';
 
 import Button from '../Button';
-import InteractiveTag from '../Tag/InteractiveTag';
+import Tag from '../Tag';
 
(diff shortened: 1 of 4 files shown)


jendowns:
Closing this one. While it would be nice to use Carbon's `Tag` with the `filter` prop enabled, that would require more of a refactor that is out of scope for this defect.

PR #513, opened 2020-04-27
fix(tagwall): accommodate truncation in tags and list items

### LINKED ISSUE #511: TagWallFilter: long tags are not truncated & overflow their container

jendowns:
## Affected issues

- Resolves #511 

## Proposed changes

- truncate tags in `TagWall` to 40ex per input from design
- truncate list items in `TagWall` to width of container
- update storybook demo to show what happens with long strings

## Testing instructions

- The `TagWall` demo has a long string now, so you can see how it appears when truncated. You can then remove the tag, returning it the list below, and see how it appears truncated in the list of options.

### DIFF
--- a/src/components/TagWall/TagWall.js
+++ b/src/components/TagWall/TagWall.js
@@ -7,7 +7,10 @@ import classnames from 'classnames';
 import { arrayOf, bool, func, shape, string } from 'prop-types';
 import React from 'react';
 
-import { getComponentNamespace } from '../../globals/namespace/index';
+import {
+  carbonPrefix,
+  getComponentNamespace,
+} from '../../globals/namespace/index';
 import * as defaultLabels from '../../globals/nls';
 
 import defaultItemToString from '../__tools__/defaultItemToString';
@@ -71,7 +74,9 @@ const TagWall = ({
             removeBtnLabel={TAG_WALL_REMOVE_BUTTON}
             type="gray"
           >
-            {itemToString(item)}
+            <span className={`${carbonPrefix}text-truncate--end`}>
+              {itemToString(item)}
+            </span>
           </InteractiveTag>
         );
       })}
--- a/src/components/TagWall/TagWall.js
+++ b/src/components/TagWall/TagWall.js
@@ -74,7 +74,10 @@ const TagWall = ({
             removeBtnLabel={TAG_WALL_REMOVE_BUTTON}
             type="gray"
           >
-            <span className={`${carbonPrefix}text-truncate--end`}>
+            <span
+              className={`${carbonPrefix}text-truncate--end`}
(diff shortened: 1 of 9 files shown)


src/components/TagWall/TagWall.js
```diff
@@ -71,7 +74,9 @@ const TagWall = ({
             removeBtnLabel={TAG_WALL_REMOVE_BUTTON}
             type="gray"
           >
-            {itemToString(item)}
+            <span className={`${carbonPrefix}text-truncate--end`}>
```
SimonFinney, line 77: Awesome! If we're applying the Carbon truncation helper class here, we should also import their partial (I believe it's called `globals/scss/helper-classes`) in `TagWall/_index.scss` to ensure that the styles will be compiled for individual component imports.

src/components/TagWall/_mixins.scss
```diff
@@ -14,5 +14,10 @@
 
   > #{$tag--interactive__namespace} {
     margin-right: $carbon--spacing-03;
+    max-width: 40ex;
```
SimonFinney, line 17: 👏

src/components/TagWallFilter/Filter/Filter.js
```diff
@@ -135,7 +135,9 @@ class Filter extends React.Component {
             className={`${namespace}__list-item__entry`}
             aria-labelledby={itemProps.id}
           >
-            {itemText}
+            <span className={`${carbonPrefix}text-truncate--end`}>
```
SimonFinney, line 138: Similar comment to https://github.com/carbon-design-system/ibm-security/pull/513/files#r416481413

SimonFinney:
Thanks @jendowns - Everything behaves great! I have one suggestion related to style imports

src/components/TagWall/TagWall.js
jendowns: Thanks! I think I'll keep using their helper class here (with the correct import of course). Otherwise I would need to still keep the `span` & create our own class to apply the `text-overflow` mixin to. Just seems easier this way.

jendowns:
@SimonFinney are you all right with me applying a `title` attribute to the truncated text wrapper? (the attribute would be present for all, regardless of whether the text is actually truncated)

SimonFinney:
@jendowns That makes sense to me!

jendowns:
Thanks @SimonFinney this is ready to review again 👍

PR #518, opened 2020-04-29
fix(checkbox): override border color for focus state

### LINKED ISSUE #517: Checkbox focus outline color is white and not accessible
## Description

Related to https://github.com/carbon-design-system/carbon/issues/5984

The `Checkbox`'s `:focus` state uses the `$focus` color token, which is **white** for g90 and g100 themes.

I already propose a fix upstream, but I'd like to also get a quick fix here. Proposed solution is to override the focus outline color & use `$interactive-04` instead (which is [blue40 for dark themes](https://carbon-elements.netlify.app/themes/examples/preview/)).

### Steps for reproducing

Check out the `Checkbox` story & tab to focus on one of the boxes: https://ibm-security.carbondesignsystem.com/?path=/story/components-checkbox--checked

jendowns:
## Affected issues

- Resolves #517 

## Proposed changes

- Remove old style override that is no longer needed
- Overrides the `border-color` from Carbon to use `$interactive-04` (which is blue40 for dark themes) instead of `$focus` (which is white for dark themes & causing issues). Note that I'm also open to using `$interactive-01` here (blue60 for all themes), but blue40 seemed appropriate because it also matches the link color 🤔 

NOTE: rules here are necessary to override this code block https://github.com/carbon-design-system/carbon/blob/master/packages/components/src/components/checkbox/_checkbox.scss#L148

### DIFF
--- a/src/components/Checkbox/_index.scss
+++ b/src/components/Checkbox/_index.scss
@@ -4,7 +4,7 @@
 /// @copyright IBM Security 2019
 ////
 
-@import '@carbon/colors/scss/colors';
+@import '@carbon/themes/scss/tokens';
 @import 'carbon-components/scss/components/checkbox/checkbox';
 @import 'carbon-components/scss/globals/scss/vars';
 
--- a/src/components/Checkbox/_index.scss
+++ b/src/components/Checkbox/_index.scss
@@ -4,17 +4,24 @@
 /// @copyright IBM Security 2019
 ////
 
+@import '@carbon/colors/scss/colors';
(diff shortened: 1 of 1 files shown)


src/components/Checkbox/_index.scss
```diff
@@ -4,17 +4,24 @@
 /// @copyright IBM Security 2019
 ////
 
+@import '@carbon/colors/scss/colors';
```
SimonFinney, line 7:
```suggestion
@import '@carbon/themes/scss/tokens';
```

PR #526, opened 2020-05-07
feat(datatable): remove ternary render, require render prop, update stories

### LINKED ISSUE #442: DataTable - Add size prop
## Request

The Carbon React component library supports a `size` prop to append different spacing classes such as `compact`, `short`, and `tall`. The `DataTable` component should accept and pass through the `size` prop.

[Storybook](https://www.carbondesignsystem.com/components/data-table/code)

### Acceptance criteria

- [ ] `DataTable` supports [Carbon](https://react.carbondesignsystem.com/?path=/story/datatable--default)'s `size` prop
- [ ] `TableOverflowCell` supports each of the specified variations - `compact`, `short`, `tall`, etc.

### LINKED ISSUE #449: Simplify `DataTable`
## Description

The `DataTable` is currently extended and modified, which incurs a manual process for implementing updates from Carbon - this work item aims to capture if this can be reduced to a pure proxy or if it can be simplified to better absorb upstream changes.

## Acceptance criteria

- [ ]

### LINKED ISSUE #525: Not passing sortHeaderKey, sortDirection through
## Bug - 

The `DataTable` + `DataTablePagination`  components do not pass through `sortHeaderKey` and `sortDirection` through to the carbon components react DataTable.

**Expected behavior -**
We should be able to provide an initial sort for the table to render, and have our backend handle the sorting.

**Actual behavior -**
Here is a log from `getDerivedStateFromProps` in carbon-components react `DataTable/state/getDerivedStateFromProps.js`

image

https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/DataTable/DataTable.js

DataTable does not pass in these props

jendowns:
## Affected issues

- Resolves #525 
- Resolves #449 
- Resolves #442 

## Proposed changes

- remove the ternary `render`, where we were rendering our own `DataTable` + subcomponents if a consumer wasn't providing a `render` prop themselves. This has proven problematic, as with #525, where we don't pass through all subcomponent props & end up providing a version of the `DataTable` that is not fully featured. We should instead require consumers include their own `render` prop, and construct their own `DataTable` with desired subcomponents + subcomponent props, just like Carbon does. **This will be a breaking change for anyone not providing a `render` prop to `DataTable`.**
- remove storybook demos that are no longer needed -- now that we are no longer providing our own version of `render`, we can import Carbon's `default` story. And for other stories, we can import Carbon's version of those as well. The only stories I did not remove were `with-skeleton` and `with-overflow` because both of those are somewhat unique to our codebase. I think we can definitely explore removing those or even just modifying them in the future, but for now I wanted to keep the scope of this PR small.


jendowns:
FYI this will also involve some changes to `DataTablePagination` -- I think our approach of providing a very opinionated `DataTable` inside `DataTablePagination` may need to come to an end -- instead, we should consider providing a demo so that folks still have a solid example.

Working out the best way to do this 👍

src/components/DataTable/DataTable.stories.js
```diff
@@ -14,27 +14,37 @@ const props = () => ({
...
-      info: {
-        text: `
+          /* eslint-enable no-useless-escape */
+        },
+      }
+    )
+    .add(
+      'with batch actions',
+      () =>
+        require('carbon-components-react/lib/components/DataTable/stories/with-batch-actions').default(
```
SimonFinney, line 42: 👏

SimonFinney:
Should also resolve #442

src/components/DataTable/DataTablePagination/DataTablePagination.stories.js
```diff
@@ -91,7 +90,7 @@ const render = ({
             <TableSelectRow {...getSelectionProps({ row })} />
             {row.cells.map(cell => (
               <TableCell key={cell.id}>
-                {cell.value ? cell.value : missingDataCharacter}
+                {cell.value ? cell.value : '-'}
```
SimonFinney, line 93: Perhaps we can add the '-' as a static `missingDataCharacter` property to the `DataTable` to still provide some consistency here - That way, here they could say `{cell.value ? cell.value : DataTable.missingDataCharacter}`

src/components/DataTable/DataTable.js
```diff
@@ -43,86 +42,30 @@ const DataTable = ({
...
-                  </TableRow>
-                ))}
-              </TableBody>
-            </Table>
-          </TableContainer>
-        )}
-      />
-    )}
+    <CarbonDataTable
+      filterRows={filterRows}
```
SimonFinney, line 54: Nice! I wonder here if we can treat this as a true pass-through now and remove all of these props bar `other`, pointing to Carbon's Storybook for additional guidance

src/components/DataTable/DataTablePagination/DataTablePagination.stories.js
jendowns: Thanks @SimonFinney -- what kind of consistency are you seeking here? I just added it to the stories, since it's like the preferred missing data character for us. But other developers in other contexts may want a different character to represent missing data. The point of this line is to provide developers an example (along with the example demo as a whole) so that they know what is possible. I'm not currently seeing the value of adding a status property for a single character that they'd likely change anyway?

SimonFinney: It's about offloading the work for those that want to remain compliant - For example:

- https://github.ibm.com/xforce-exchange/xfe-isc/blob/master/services/reports/src/common/XFEDataTable.js#L113
- https://github.ibm.com/investigate/investigate/blob/master/src/components/tableWrapper/TableWrapper.js#L152

I assume these consumers aren't interested in what character to use for the `missingDataCharacter`, only that it's in line with what we recommend - if for example, this was to change, they won't need to keep an eye on Storybook or revisit it. It's a low-hanging helper for something that provides a lot of maintenance value.

jendowns: Thanks for your thoughts!

I said "preferred by us" above not to reference some official recommendation (which I am not aware of), but instead to reference that this is merely the way I'm presenting the story demo of the component.

The stakes are low for them to just copy & paste the `-` character in a ternary inside this demo, or go a different direction if they prefer. I don't think we should save off a single character like this as a static property (and then add that to our growing list of exports). 👍

jendowns:
@SimonFinney I believe I addressed your comments above & I also updated the description to include the other related issue you found. This is ready to review again when you get a chance -- thanks!

src/components/DataTable/DataTablePagination/DataTablePagination.stories.js
SimonFinney: I'm not sure I entirely agree - I'm not hugely concerned about what's exported as long as it provides added value in the form of consistency and efficiency to consumers. That's the whole point of why we're doing this 🙃

We can defer this for now, but I'll reach out to a selection of consumers currently leveraging this to understand their needs and we can create a follow-up appropriately.

PR #531, opened 2020-05-11
feat(tagwall): apply id to each tag, allow extra attrs for tags

### LINKED ISSUE #530: `TagWall` improvements
### Proposal

This came up in Slack --
A couple improvements could be made to the `TagWall` to improve developer experience.

1. Minimum: **Ability to target specific tags**

2. Bonus: **Ability to add attributes to specific tags**

In both cases, developers require more control over the child tags inside the parent `TagWall`.

The "minimum" capability in this case would be enabling developers to easily target a specific tag (with a CSS selector). A step-up from the minimum would be to allow developers to add additional attributes to specific tags.

jendowns:
## Affected issues

- Resolves #530

## Proposed changes

- add the `id` from the `items` prop to each `Tag` inside `TagWall`
- add a new key to the `items` object to allow for extra attributes on specific tags
- update snapshots to reflect prop updates

## Testing instructions

- I updated the mocks to use the new `props` object so you should be able to see `example-attribute` rendered in the storybook demo.


src/components/TagWall/TagWall.js
```diff
@@ -116,6 +118,7 @@ TagWall.propTypes = {
       id: string.isRequired,
       isSelected: bool,
       label: string.isRequired,
+      props: object,
```
jendowns, line 121: - Is `props` an ideal name for this? Or maybe `extra-props`? 🤔 
- We don't know the shape of `props` in this case, but it still may help to provide more info?

src/components/TagWall/_mixins.scss
```diff
@@ -16,6 +16,7 @@
   > #{$tag--interactive__namespace} {
     margin-right: $carbon--spacing-03;
     max-width: 40ex;
+    justify-content: space-between;
```
jendowns, line 19: If a tag is made wider than its content (like if a width is set to 100% or just a larger pixel value), then the tag content + the close button are both centered inside the tag, which can look strange. This change will just ensure that the text is on the left, and the close button is on the right. You can see what I mean if you look at the `dev` branch deploy and set a specific width on a tag, like 300px or something larger than the natural width.

src/components/TagWall/TagWall.js
```diff
@@ -63,6 +63,7 @@ const TagWall = ({
 
         return (
           <InteractiveTag
+            id={item.id}
```
jendowns, line 66: Note that `id` for each item is already required.

src/components/TagWall/TagWall.js
SimonFinney: 🎉 Awesome stuff! A quick search of Carbon's repo wasn't able to produce any results, although I saw `attrs` used in places - Perhaps @joshblack could provide some guidance here? Otherwise, I'm good with props for now!

joshblack: Looking at this real quick, is this a forwarding-props use-case? 👀 In the code it seems like it's a part of the `items` array that is mapped over to return an `InteractiveTag`.

If so, I think this is a pattern that we've found ourselves doing for APIs like Downshift or DatePicker. Downshift might be the most comparable in that it has a similar need for items.

In general, I think we've realized that forwarding props in this way have either limited the types of changes that we can make, or puts us in a situation where it is hard to test whether a change breaks existing usage.

In the first case, the issue may manifest itself in props like `downshiftProps` where we forward things down to the Downshift library itself. This couples us to the version of downshift at a given point in time and has made it so that we can't upgrade without a major change. Similarly, if forwarding props to a component (especially in a component package future) we could run into the same situation.

In the latter case, having unbounded inputs has prevented us from identifying what we needed to test for while making updates. Our preference in this case would be to strictly type/define the interface for each prop so that we can enumerate the situations we have to test for and (hopefully) catch bugs in the future.

Similarly, in these cases we might try to invert the Component pattern:

```js
// From (pull syntax, receive items and decorate them)
<List items={items} />

// To (a) (implict method, add props with cloneElement or Context)
<List>
  {items.map(...)}
</List>

// To (b) (push method, give props and delegate to team to render)
<List renderItem={itemProps => ...} />

// To (c) (push method, hooks variant, return props based on options)
const [items] = useList(options);

<List>
  {items.map(item => ...)}
</List>
```

SimonFinney: @joshblack Thanks, that's very helpful! I think that's ideally where we want to go but it would constitute a breaking change

PR #532, opened 2020-05-12
fix(decorator): update icons, add critical severity

### LINKED ISSUE #478: DataDecorator shape update
Switch the shapes used to indicate low & high 

The "low" triangle shape looks similar to the warning icon which unintentionally could signal that low appears to be more of a warning than the medium & high shapes

For future reference
image

jendowns:
## Affected issues

- Mostly resolves #478 (minus the "informational" icon)

## Proposed changes

- update "low" severity icon
- update "high" severity icon
- update "unknown" icon
- add "critical" severity icon that is applied only when highest threshold is met
- remove unused `DataDecorator` constants file

## Considerations

- "Critical" is a new severity level, so I made an assumption that it would only be applied is a score was equal to (or somehow, above) the highest threshold. ***Is this correct?***

- **We do not have an "informational" state as mentioned in #478** so I'm not aware of what conditions would need to be met for that state to apply. I'm thinking that means there is no score, but I believe that the "unknown" state would apply in that case. How should we distinguish between "information" and "unknown"?


src/components/DataDecorator/constants.js
```diff
@@ -1,36 +0,0 @@
-/**
```
SimonFinney, line 1: I'd be interested in understanding if this will introduce a breaking change for any consumers leveraging these icons

SimonFinney:
Visual design is not my area of expertise, but I _think_ the 'high' and 'critical' icons are misaligned here - @cameroncalder could you confirm? 

Critical

High

src/components/DataDecorator/constants.js
jendowns: @SimonFinney that's a good point -- I'll return the file for now and just make sure it imports icons from `Decorator/constants` (not redefines them).

SimonFinney:
@jendowns Thanks! The only thing is I realised the constants need to be reverted after I hit approve 🤦

jendowns:
No worries @SimonFinney! I restored the constants file so now I'll go ahead and merge this 👍

PR #537, opened 2020-05-12
fix(decorator): change interactive decorator text color

### LINKED ISSUE #535: Bug: Interactive decorators do not look different from inert ones
**Expected behavior -**
Interactive decorators should look different from inert ones to provide better guidance to users. 

**Actual behavior -**
Interactive and inert decorators look identical in non-hover states.

### Screenshots

**inert**
image

**as link**
image

**as button**
image

### Optional information

**Version -** 1.22.1

jendowns:
## Affected issues

- Resolves #535 

## Proposed changes

- change interactive decorator text color to blue30


src/components/DataDecorator/_mixins.scss
```diff
@@ -47,8 +47,11 @@
   padding: unset;
   border: unset;
   vertical-align: middle;
+  color: $text-02;
+
+  &#{$root}--interactive {
```
diego-codes, line 52: Is there a reason that you need to increase the specificity of this selector?

joshkimmell:
@jendowns The decorator type (i.e. `IP`) should maintain the default color and should NOT change to blue-30.

src/components/DataDecorator/_mixins.scss
jendowns: Because I don't want to rely on style order to set a text color 👍

diego-codes: Based on the changes I'm seeing, the specificity can go down and it should apply the styles as expected. Higher specificities are hard to work with as a consumer.

jendowns: To illustrate, this relies on order:

```
.security--decorator {
  color: $text-02;
}

.security--decorator--interactive {
  color: $hover-primary-text;
}
```

And this relies on specificity:

```
.security--decorator {
  color: $text-02;
}

.security--decorator.security--decorator--interactive {
  color: $hover-primary-text;
}
```

We should lean on adding specificity when necessary, instead of relying on style order, to get desired results.

In fact there is currently a situation where `.bx--link` color styles are overriding the `.security--decorator` color rule, due solely to style order, if you  use the browser inspector and check out which text color is being applied, BEFORE I override it with `.security--decorator.security--decorator--interactive`

jendowns:
@SimonFinney @diego-codes @joshkimmell I updated this PR with some more substantial changes than before. Namely it's not longer taking in the `bx--link` class which was injecting a bunch of styles that we ended up overriding anyway. In addition, I have increased the specificity of a number of rules because we need to override things by relying on specificity, not style order. If you see a style rule that you think is too specific, let me know which rule & we can take a look if indeed the specificity can be reduced without having to rely on style order. Thanks!

Also @petervachon based on the feedback above about link color, do you mind taking another look & re-reviewing? Thanks!

jendowns:
For sure, @diego-codes. I understand CSS specificity & the cascade.

As mentioned in https://github.com/carbon-design-system/ibm-security/pull/537#discussion_r424652724, changes to specificity here are required to avoid bugs with ordering (for example: without this change, a mere class name order swap could mean that the wrong color is applied).

So in that respect, I'm not sure how the painpoint you describe applies to this? But again, if you see a place where specificity is not required (and also, where we wouldn't have to rely on ordering, if specificity were reduced), then please let me know. I'm happy to update this PR with that feedback!

For illustrative purposes, here is a list of all selectors created by this stylesheet:

```scss
.security--decorator {
   //
}

.security--decorator__icon {
    //
}

.security--decorator__type,
.security--decorator__value {
   //
}

.security--decorator--interactive:not(.security--decorator--active) .security--decorator__type {
    //
}

.security--decorator--interactive:not(.security--decorator--active) .security--decorator__value {
    //
}

.security--decorator--interactive:not(.security--decorator--active):visited .security--decorator__value {
    //
}

.security--decorator.security--decorator--interactive:hover {
    //
}

.security--decorator.security--decorator--interactive:focus {
    //
}

.security--decorator.security--decorator--interactive:active,
.security--decorator.security--decorator--interactive:focus,
.security--decorator.security--decorator--interactive:hover,
.security--decorator.security--decorator--interactive:focus:active {
    //
}

.security--decorator.security--decorator--interactive:active .security--decorator__value, .security--decorator.security--decorator--interactive:focus .security--decorator__value, .security--decorator.security--decorator--interactive:hover .security--decorator__value, .security--decorator.security--decorator--interactive:focus:active .security--decorator__value {
    //
}

.security--decorator.security--decorator--active,
.security--decorator.security--decorator--active:hover,
.security--decorator.security--decorator:focus:active {
    //
}

.security--decorator.security--decorator--active__type,
.security--decorator.security--decorator--active:hover,
.security--decorator.security--decorator:focus:active {
    //
}

.security--decorator.security--decorator--active:focus {
    //
}

.security--decorator__icon,
.security--decorator__type,
.security--decorator__value {
    //
}

.security--decorator.security--decorator--inline {
    //
}

.security--decorator.security--decorator--inline > .security--decorator__icon {
    //
}

.security--decorator.security--decorator--inline > .security--decorator__type,
.security--decorator.security--decorator--inline > .security--decorator__value {
    //
}

.security--decorator.security--decorator--inline > .security--decorator__type,
.security--decorator.security--decorator--inline > .security--decorator__value {
    //
}

.security--decorator--unknown .security--decorator__icon {
    //
}

.security--decorator--low .security--decorator__icon {
    //
}

.security--decorator--medium .security--decorator__icon {
    //
}

.security--decorator--high .security--decorator__icon,
.security--decorator--critical .security--decorator__icon {
    //
}

.security--decorator--high--active .security--decorator__icon,
.security--decorator--high:focus:active .security--decorator__icon,
.security--decorator--critical--active .security--decorator__icon,
.security--decorator--critical:focus:active .security--decorator__icon {
    //
}

.security--decorator__type {
    //
}

.security--decorator__value {
    //
```

diego-codes:
I get the concern about the order of selectors. However, you, as the author of the styles, control the order. If the style breaks due to order, you can fix it by rearranging the selectors. My point is that we'd be trading off the ability to easily modify styles as a consumer (which happens more often than we'd think and for valid reasons) for convenience for the maintainers of this package. The trade-off I am proposing is requesting more discipline from maintainers of this package (1 team) for the ability to modify styles as a consumer without breaking CSS best practices.

A use case that comes to mind is one where we need to adjust the layout of the decorator to accommodate for a large payload value like so:
image

Increasing the specificity of the styles now will likely break the feature for us. Additionally, to resolve the problem, we would have to create even more specific CSS selectors that just encourage messy CSS. 

This is a problem we face as a single application team that is very much in lock-step with this package. Other application teams would not be as involved or aware of these changes and their implications.

jendowns:
@diego-codes What specific layout styles are you trying to override here in the example above?

And which specific selector is going to cause problems for you?

PR #550, opened 2020-05-20
fix(storybook): add keys to theme components

jendowns:
Right now in development, when viewing the storybook environment, you will see a large warning about a miss `key` value on the top level `Theme`:

Screen Shot 2020-05-20 at 12 51 15 PM

## Proposed changes

- the component wrapper `Theme` will have a `key={NAME-OF-THEME}` (this is not the specific missing `key` being flagged but still seems useful, anyway)
- the `Theme` addon panel will have a `key` that matches naming convention of other addon panels -- e.g., `key="storybook/theme/panel"`

### DIFF
--- a/.storybook/addons/addon-theme/index.js
+++ b/.storybook/addons/addon-theme/index.js
@@ -61,7 +61,11 @@ class Wrapper extends Component {
     this.setState({ theme });
   };
 
-  render = () => <Theme theme={this.state.theme}>{this.props.children}</Theme>;
+  render = () => (
+    <Theme key={this.state.theme} theme={this.state.theme}>
+      {this.props.children}
+    </Theme>
+  );
 }
 
 export default makeDecorator({
--- a/.storybook/addons/addon-theme/register.js
+++ b/.storybook/addons/addon-theme/register.js
@@ -58,7 +58,12 @@ register(namespace, api =>
   addPanel(`${namespace}/panel`, {
     title,
     render: ({ active }) => (
-      <Theme active={active} api={api} channel={getChannel()} />
+      <Theme
+        key="storybook/theme/panel"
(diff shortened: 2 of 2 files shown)


.storybook/addons/addon-theme/index.js
```diff
@@ -61,7 +61,11 @@ class Wrapper extends Component {
     this.setState({ theme });
   };
 
-  render = () => <Theme theme={this.state.theme}>{this.props.children}</Theme>;
+  render = () => (
```
SimonFinney, line 64: 👏

.storybook/addons/addon-theme/index.js
```diff
@@ -61,7 +61,11 @@ class Wrapper extends Component {
...
+    <Theme key={this.state.theme} theme={this.state.theme}>
```
SimonFinney, line 65: 👏

.storybook/addons/addon-theme/register.js
```diff
@@ -58,7 +58,12 @@ register(namespace, api =>
   addPanel(`${namespace}/panel`, {
     title,
     render: ({ active }) => (
-      <Theme active={active} api={api} channel={getChannel()} />
+      <Theme
+        key="storybook/theme/panel"
```
SimonFinney, line 62: Nice! I wasn't familiar with this convention, where can I find more information about it? Perhaps in a follow-up, we update the `namespace`

jendowns: You can see the keys of other storybook panels if you inspect the storybook environment using React DevTools:

Screen Shot 2020-05-21 at 8 20 11 AM

I don't recommend trying to edit those keys though.

PR #558, opened 2020-05-22
feat(filter-panel): use new heading prop, change title prop to string

### LINKED ISSUE #557: Tooltip for FilterPanelGroup & FilterPanelAccordionItem shows as [object Object] when using a node as a title
## Bug -
The definitions of `title` in FilterPanelGroup and FilterPanelAccordionItem indicate it's supposed to be `PropTypes.node`.

However, if you use a node as a title, the tooltip will become [object Object] when you hover over it.

**Expected behavior -**

Tooltip should not be [object Object]

**Actual behavior -**

Tooltip is [object Object]

### Steps for reproducing

Please copy the examples from https://ibm-security.carbondesignsystem.com/?path=/story/components-tooltip--default-bottom and https://ibm-security.carbondesignsystem.com/?path=/story/patterns-filterpanel--default

The example for Tooltip starts with `<[object Object]>` and ends with `</[object Object]>`, so please change that to `<Tooltip>` and `</Tooltip>` when copying the example.

In the `title` attributes for FilterPanelGroup and FilterPanelAccordionItem, pass in the example for Tooltip.

Now, using the mouse, hover over the FilterPanelGroup or FilterPanelAccordionItem.

Example: https://codesandbox.io/s/codesandbox-sundk?file=/src/index.js

### Screenshots

N/A

### Affected browsers

Google Chrome Version 83.0.4103.61 (Official Build) (64-bit), Windows 10

### Optional information

v1.23.0

jendowns:
## Affected issues

- Resolves #557 

A prop called `title` is being used to both populate the heading node of a component (which is fine) and is also being passed to the `title` attributes of those headings (which is not fine). The [`title` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) needs to be text, and it cannot be display HTML nodes. If the `title` HTML attribute is passed a node, it displays `[object Object]` on hover (as observed in #557)

## Proposed changes

- ~add a new prop called `hoverTitle` that should be a string. This prop exists to populate the `title` HTML attribute only.~

UPDATES:
- added a `heading` prop that can accept a `node` (this replaces the usage of the old `title` prop)
- changed the existing `title` prop to be a `string` type
- update tests
- update `heading` vs `title` logic -- if a `heading` is not provided, fallback to the `title` for contents of the `FilterPanelLabel` component 👍

### DIFF
--- a/src/components/FilterPanel/FilterPanel.stories.js
+++ b/src/components/FilterPanel/FilterPanel.stories.js
@@ -59,7 +59,7 @@ const LegacyFilterPanelWithState = compose(
 const checkboxProps = () => ({
   count: 10,
   onChange: action('FilterPanelCheckbox onChange'),
-  title: text('FilterPanelCheckbox title attribute (title)', title),
+  title: 'FilterPanelCheckbox title attribute (title)',
 });
 
 LegacyFilterPanelWithState.displayName = getDisplayName(FilterPanel);
--- a/src/components/FilterPanel/FilterPanel.stories.js
+++ b/src/components/FilterPanel/FilterPanel.stories.js
@@ -162,7 +162,7 @@ storiesOf(patterns('FilterPanel'), module)
       </FilterPanelAccordion>
       <FilterPanelAccordion title="Filter accordion 2" count={100}>
         <FilterPanelAccordionItem
-          title="Filter accordion item 1"
+          title={<span>Filter accordion item 1</span>}
           count={4}
           expandLabel="View more"
           collapseLabel="View less"
--- a/src/components/FilterPanel/FilterPanel.stories.js
+++ b/src/components/FilterPanel/FilterPanel.stories.js
@@ -56,6 +56,12 @@ const LegacyFilterPanelWithState = compose(
   })
 )(FilterPanel);
 
+const checkboxProps = () => ({
+  count: 10,
+  onChange: action('FilterPanelCheckbox onChange'),
+  title: text('FilterPanelCheckbox title attribute (title)', title),
+});
+
 LegacyFilterPanelWithState.displayName = getDisplayName(FilterPanel);
 LegacyFilterPanelWithState.__docgenInfo = FilterPanel.__docgenInfo;
 
@@ -83,21 +89,18 @@ storiesOf(patterns('FilterPanel'), module)
           <FilterPanelCheckbox
             labelText="Filter checkbox"
             id="result-filter-checkbox"
-            count={10}
-            onChange={action('FilterPanelCheckbox onChange')}
+            {...checkboxProps()}
           />
           <FilterPanelCheckbox
-            labelText="Long filter checkbox  label"
+            labelText="Long filter checkbox label"
             id="result-long-filter-checkbox"
-            count={10}
-            onChange={action('FilterPanelCheckbox onChange')}
+            {...checkboxProps()}
           />
           <FilterPanelCheckbox
             labelText="Checked"
             id="result-checked"
-            count={10}
             defaultChecked
-            onChange={action('FilterPanelCheckbox onChange')}
+            {...checkboxProps()}
           />
         </FilterPanelGroup>
         <FilterPanelGroup title="Truncated accordion item">
@@ -107,8 +110,7 @@ storiesOf(patterns('FilterPanel'), module)
               key={index}
               labelText={`Filter ${index + 1}`}
               id={`result-filter-checkbox-${index + 1}`}
-              count={10}
-              onChange={action('FilterPanelCheckbox onChange')}
+              {...checkboxProps()}
             />
           ))}
         </FilterPanelGroup>
@@ -125,21 +127,18 @@ storiesOf(patterns('FilterPanel'), module)
           <FilterPanelCheckbox
             labelText="Filter checkbox"
             id="filter-checkbox"
-            count={10}
-            onChange={action('FilterPanelCheckbox onChange')}
+            {...checkboxProps()}
           />
           <FilterPanelCheckbox
             labelText="Long filter checkbox  label"
             id="long-filter-checkbox"
-            count={10}
-            onChange={action('FilterPanelCheckbox onChange')}
+            {...checkboxProps()}
           />
           <FilterPanelCheckbox
             labelText="Checked"
             id="checked"
-            count={10}
             defaultChecked
-            onChange={action('FilterPanelCheckbox onChange')}
+            {...checkboxProps()}
           />
         </FilterPanelAccordionItem>
         <FilterPanelAccordionItem
@@ -154,8 +153,7 @@ storiesOf(patterns('FilterPanel'), module)
               key={index}
               labelText={`Filter ${index + 1}`}
               id={`filter-checkbox-${index + 1}`}
-              count={10}
-              onChange={action('FilterPanelCheckbox onChange')}
+              {...checkboxProps()}
             />
           ))}
         </FilterPanelAccordionItem>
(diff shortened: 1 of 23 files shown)


src/components/FilterPanel/FilterPanel.stories.js
```diff
@@ -162,7 +162,8 @@ storiesOf(patterns('FilterPanel'), module)
       </FilterPanelAccordion>
       <FilterPanelAccordion title="Filter accordion 2" count={100}>
         <FilterPanelAccordionItem
-          title="Filter accordion item 1"
+          title={<span>Filter accordion item 1</span>}
+          hoverTitle="Filter accordion item 1"
```
diego-codes, line 166: This was my fault for not having thought this through when I created these components. Especially because `title` means something in HTML 🤦. 

I'm wondering if a `heading` prop could be introduced instead to replace the current `title` prop and keep the `title` prop to mean the same as it means in HTML.

A transition strategy could be that title is applied when the `heading` is not provided:
```jsx
// Maintain current behavior for users (maybe add a warning when heading does 
// not exist to get people to transition)
<FilterPanelAccordion title="I appear as both the heading and title" />

```

```jsx
// Provide a separate heading and title (similar to title & hoverTitle, but 
// using title to match the HTML attribute name)
<FilterPanelAccordion
  heading={<span>Visible heading</span>}
  title="Hover title that is different"
/>
```

src/components/FilterPanel/FilterPanel.stories.js
```diff
@@ -162,7 +162,8 @@ storiesOf(patterns('FilterPanel'), module)
       </FilterPanelAccordion>
       <FilterPanelAccordion title="Filter accordion 2" count={100}>
```
diego-codes, line 163: Do the `FilterPanelAccordion` or `FilterPanelGroup` props also need updates to the prop types?

diego-codes:
Thanks for working on this, @jendowns! I have a suggestion and a question.

src/components/FilterPanel/FilterPanel.stories.js
jendowns: @diego-codes that's a great suggestion, but the primary reason I took this path is to avoid a breaking change for existing users.

But I suppose your team is the primary user in this case? So if you are okay with that change & updating your team's usage, this would work for me 👍

src/components/FilterPanel/FilterPanel.stories.js
jendowns: It looks like `FilterPanelAccordion` just passes the `title` prop value to `FilterPanelGroup`:
https://github.com/carbon-design-system/ibm-security/blob/9b621a92fc10b0651be3d4f5dee1db8bbc2b2f3c/src/components/FilterPanel/FilterPanelAccordion/FilterPanelAccordion.js#L24-L30

And `FilterPanelGroup` just sends it to `FilterPanelLabel`, which is addressed in this PR:
https://github.com/carbon-design-system/ibm-security/blob/9b621a92fc10b0651be3d4f5dee1db8bbc2b2f3c/src/components/FilterPanel/FilterPanelGroup/FilterPanelGroup.js#L27-L29

If you mean the propType documentation, sure that can be updated to be more specific. 👍

src/components/FilterPanel/FilterPanel.stories.js
diego-codes: I'd be fine with the change, especially if you fall back to `title` when a `heading` isn't provided then the change wouldn't be breaking.

src/components/FilterPanel/FilterPanel.stories.js
diego-codes: That and also the `propTypes` of the components probably need updates:

https://github.com/carbon-design-system/ibm-security/blob/cf0819c334010dd430319482d56ec9941d729d39/src/components/FilterPanel/FilterPanelAccordion/FilterPanelAccordion.js#L35-L49

src/components/FilterPanel/FilterPanel.stories.js
jendowns: I mean: it would be "breaking" in that if you are currently passing in a `node` type to `title`, you will start to get proptype errors because the prop has become more restrictive.

src/components/FilterPanel/FilterPanel.stories.js
jendowns: Yes, certainly -- if we are going with the changes you proposed in the other comment, then I will of course be updating the propType of the `title` prop to keep in it line with my original proposed fix: that is, not passing a `node` to the HTML `title` attribute 👍 

And I will also be adding propType information with a new `heading` prop 👍

src/components/FilterPanel/FilterPanel.stories.js
diego-codes: Oh yeah, that's a good point. 🤔

jendowns: Just realized this is a `feat` so it will be a non-trivial version bump 👍 I think it'll be fine

jendowns:
Thanks @diego-codes! This is ready to review again 🎉

src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
```diff
@@ -67,6 +73,7 @@ FilterPanelLabel.defaultProps = {
   countLabel: count => `${count} items`,
   className: undefined,
   countClassName: undefined,
+  title: 'title',
```
diego-codes, line 76: I would make this one `undefined` as well because a title is not always necessary and it looks strange having a title that says "title"

src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
```diff
@@ -59,6 +60,11 @@ FilterPanelLabel.propTypes = {
    * Optional class name for the count.
    */
   countClassName: PropTypes.string,
+
+  /**
+   * The `title` attribute that is applied to the label node.
+   */
+  title: PropTypes.string,
```
diego-codes, line 67: I don't know where else add this comment, but the `FilterPanelCheckbox` component also needs to accept the `title` prop. When I updated the storybook for it to test, the title did not get applied:

image

image

With this in mind, it might also be good to add a knob for the title prop for this component.

diego-codes:
I just had another thought, does it make sense to add some test to test the fallback logic to `title` and that we in-fact can provide a heading that is separate from the title? The best component to test this on would probably be the `FilterPanelLabel` component which is the hub for this functionality.

jendowns:
Hey @diego-codes I went ahead and updated a number of tests -- could you take a look when you get a chance?

In the case of the `heading` versus `title` rendering inside the `FilterPanelLabel`, I didn't add a test for that in that particular component's tests because that component appears to render whatever `children` it is passed. It seems like the more valuable test would be in the `FilterPanelGroup`, so I put it there. But let me know what you think!

With regards to a storybook knob for `title` etc., I definitely agree with your comment but the storybook demo right now has a LOT Of `title` props applied to the diff subcomponents etc., so I'm having trouble seeing where the most value would be, for the knob. 🤔 `FilterPanelGroup`, or...? (Doesn't seem worthwhile to have one knob control *all* `title` props used in that file, or have several knobs for each type of `title` prop usage...)

src/components/FilterPanel/FilterPanelLabel/FilterPanelLabel.js
jendowns: @diego-codes Just noticed this comment... it looks like `title: 'title'` by default both here and in `FilterPanelAccordionItem`, so I will make it `undefined` in both places 👍

diego-codes: That's a good question, I don't remember to be honest. Maybe because the `FilterPanelAccordionItem` must always have a label while `FilterPanelAccordion` doesn't? But I think that making it consistent with the rest might be preferred.

PR #561, opened 2020-05-25
chore: upgrade Carbon packages

### LINKED ISSUE #539: Upgrade Carbon packages - `v10.12.0`
## Description

To resolve #534, and as this package is lagging a bit behind in Carbon's packages versions, this work item should facilitate that. See also:

- https://github.com/carbon-design-system/carbon/releases/tag/v10.10.2
- https://github.com/carbon-design-system/carbon/releases/tag/v10.10.3
- https://github.com/carbon-design-system/carbon/releases/tag/v10.11.0
- https://github.com/carbon-design-system/carbon/releases/tag/v10.11.1
- https://github.com/carbon-design-system/carbon/releases/tag/v10.11.2

## Acceptance criteria

- [ ] Upgrade to [Carbon `v10.12.0`](https://github.com/carbon-design-system/carbon/releases/tag/v10.12.0)
- [ ] Confirm that changes are non-breaking

SimonFinney:
## Affected issue

Resolves #539 

## Proposed changes

- Upgrade Carbon packages
- Enforce `carbon-icons` dependency as it's a deep dependency of `Icon` in `carbon-components-react` and only defined as a `devDependency` - https://github.com/carbon-design-system/carbon/blob/master/packages/react/package.json#L84
- Fix `ComboButton` regression
- Fix updated import naming
- Prettier formatting


SimonFinney:
@jendowns I believe this is finally ready for review!

SimonFinney:
@jendowns Good point! It looks like this change has come in from Carbon explicitly setting the height to `2.5rem` now. I'm all for inheriting it, but I agree it looks strange for the `FilterPanel`:

Accordion

I suspect we need to discuss with @petervachon and @diego-codes (possibly others?) on how we'd like to approach this change - any thoughts?

petervachon:
Agree, it's a bit weird with the spacing in the filter panel. My initial question would be, what's the cost of ownership if we wanted to make updates?

jendowns:
@petervachon @SimonFinney I don't think it would be unreasonable for us to just explicitly set the high of the `FilterPanelAccordionItem` (which we "own") to 32px and call it good. That way we don't break any existing designs etc. 👍

SimonFinney:
Ah, thank you - I should've checked! I had assumed the styles were shared 😆

PR #572, opened 2020-05-28
fix(wizard): add className prop and spread attribute

jendowns:
## Affected issues

- Resolves #565

## Proposed changes

- add `className` prop and spread attribute to `Tearsheet`
- add `className` prop and spread attribute to `Wizard`

PR #576, opened 2020-05-28
fix(tearsheet): pass through isLoading, provide story examples of making tearsheet content non-interactive when loading

### LINKED ISSUE #563: Tearsheet is accessible and editable using the Tab key with loading
## Bug
Tearsheet is accessible and editable using the Tab key from the keyboard when it is in loading state

**Expected behavior -**
Fields inside the Tearsheet should not be accessible using the Tab key on loading state

**Actual behavior -**
Fields are accessible and editable on loading state

### Steps for reproducing
1. Run storybook
2. Go to Tearsheet(with loading)
3. Toggle `focusTrap`
4. Push the `Tab` button from the keyboard until it focusess one of the fields inside the Tearsheet

### Screenshots
screenshot

### Affected browsers
- Google Chrome 83.0.4103.61 (Official Build) (64-bit) Windows 10

jendowns:
## Affected issues

- Partially resolves #563 (for the `Tearsheet`, not `Wizard`)

## Proposed changes

- provide `isLoading` via `renderMain` render prop so folks can access loading state & make their own inputs, links, buttons, etc., un-focussable when `isLoading={true}`
- set buttons rendered directly in Tearsheet (like "Delete" and "Close" button) to `disabled` if `isLoading={true}`
- after #572, props `isOpen`, `loading`, and `loadingMessage` are being passed through via the spread attribute... this causes some noise in the console in local development... so by explicitly defining them before the spread attribute, we can avoid these local warnings.


SimonFinney:
Thanks @jendowns! I'm curious to understand the impact (if any) this may have on user experience if someone is stepping through the `Wizard` and while loading in-between, the main and sidebar content becomes hidden each time - See https://deploy-preview-576--ibm-security.netlify.app/?path=/story/patterns-wizard--dynamic as an example.

@petervachon I'm interested in hearing your thoughts on what the expected interaction should be in these cases?

jendowns:
@SimonFinney I updated the approach in this PR. Instead of hiding inner elements when `loading: true`, I changed so that whenever `loading` is changed, it finds focussable elements inside the `Tearsheet` and sets their `tabindex` to `-1` (if `loading:true`) or to `0` (`loading:false`)

I specifically added this logic to non-user-added buttons generated in the `Tearsheet` (like the "Delete" button, etc) instead of relying just on the query selector, due to speed. When I was testing this with even just 100 extra form elements to see how quickly their tab indexes were set, the "Delete" and "Close" buttons always took 0.5 seconds to become un-tabbable, which was too slow if someone tabbed right when the page loaded.

jendowns:
Yeah while the UX of hiding content was kinda strange, one of the upsides was that you didn't have to rely on the speed of `querySelectorAll` to set tab indexes. Even with some updates I made, `querySelectorAll` is still a bit slow 🤔 If you tab right when the page loads, there's a 0.5-1 second or so period where you could still tab to some focussable elements... 

@SimonFinney do you have an idea to speed this up? Seems difficult since the content is user-provided...

jendowns:
@SimonFinney I have updated this PR to provide guidance for using `isLoading` via `renderMain`. I also updated the desc of this PR because technically the original issue still persists for the `Wizard` (because it is not utilizing `isLoading` yet).

src/components/Tearsheet/Tearsheet.js
```diff
@@ -78,6 +78,9 @@ class Tearsheet extends Component {
         onClick: onDeleteButtonClick,
       },
       labels,
+      loading,
+      loadingMessage,
+      isOpen,
```
SimonFinney, line 83: 👏

src/components/Tearsheet/Tearsheet.js
```diff
@@ -111,15 +114,18 @@ class Tearsheet extends Component {
...
                     {this.state.loadingMessage}
                   </div>
-                </Loading>
+                </LoadingMessage>
               )}
 
               {renderSidebar && (
-                <section className={`${namespace}__sidebar`}>
+                <section
+                  aria-hidden={this.state.loading}
```
SimonFinney, line 126: Interesting! This indicates to screen readers that anything within these sections should not be read, right? If so, perhaps we can revisit programmatically 'skipping' the tab focus in combination with this to reduce the requirement on consumers to manage it themselves.

jendowns: @SimonFinney I'm not sure what you mean here. There's a difference between what `aria-hidden` is doing and what the rest of the code to prevent tab focus is doing. 

In addition, I don't believe we should try to mutate user-provided props... in this case, it would be mutating `renderMain` content.

src/components/Tearsheet/Tearsheet.stories.js
```diff
@@ -45,20 +47,76 @@ const id = 'search';
 const selector = `#${id}`;
 const selectorPrimaryFocus = () => document.querySelector(selector);
 
-const renderMain = () => (
-  <Form className={namespace}>
+const renderMain = ({ isLoading }) => (
```
SimonFinney, line 50: 🔥

src/components/Tearsheet/Tearsheet.js
```diff
@@ -141,6 +147,7 @@ class Tearsheet extends Component {
                         kind="ghost-danger"
                         onClick={onDeleteButtonClick}
                         renderIcon={icon}
+                        tabIndex={this.state.loading ? -1 : 0}
```
SimonFinney, line 150: This feels like an opportunity for us to pull this out as a variable and prevent us from repeating the ternary.

src/components/Tearsheet/Tearsheet.js
```diff
@@ -149,7 +156,10 @@ class Tearsheet extends Component {
                 </section>
               )}
 
-              <section className={`${namespace}__main`}>
+              <section
+                aria-hidden={this.state.loading}
+                className={`${namespace}__main`}
+              >
                 {!closeButton.isDisabled && (
                   <IconButton
```
SimonFinney, line 164: This just occurred to me and is likely scope creep that can be resolved in a follow-up, but should the user be prevented from closing the `Tearsheet` using the close button while it's loading? @petervachon

SimonFinney: Good point!

jendowns: @petervachon did you have any thoughts about @SimonFinney's question above?

petervachon: Yes, the user should be prevented from closing the tearsheet while it's loading.

PR #580, opened 2020-06-01
feat(decorator): export icons as components, add story and tests

### LINKED ISSUE #479: Alert color shapes as their own component
Decouple the alert color shapes from the DataDecorator so they can be used on their own as components. Example use would be in a table column.

Component should include the shape and a label to the right displaying the number value

jendowns:
## Affected issues

- Resolves #479

## Proposed changes

- export icons as separate components (i.e., `<Decorator.Low />`,  `<Decorator.High />`, etc.)
- add tests for new icon exports
- add new story to display new icon components

## Testing instructions

- see new story (`Decorator` > `Icons` story)
- see tests & output

### DIFF
--- a/src/components/DataDecorator/Decorator/Decorator.js
+++ b/src/components/DataDecorator/Decorator/Decorator.js
@@ -187,7 +187,7 @@ function generateIconExports(...iconNames) {
         path={icons[iconName]}
         size={size}
         viewBox="0 0 16 16"
-        description={description}
+        aria-label={description}
         {...other}
       />
     );
--- a/src/components/DataDecorator/Decorator/Decorator.js
+++ b/src/components/DataDecorator/Decorator/Decorator.js
@@ -177,7 +177,7 @@ Decorator.defaultProps = {
 function generateIconExports(...iconNames) {
   const namedExports = {};
   iconNames.forEach(iconName => {
-    const Icon = ({ className, description, size, ...other }) => (
+    const IconComponent = ({ className, description, size, ...other }) => (
       <Icon
         className={classnames(
           `${namespace}__icon--${iconName.toLowerCase()}`,
@@ -191,7 +191,7 @@ function generateIconExports(...iconNames) {
         {...other}
       />
     );
-    Icon.propTypes = {
+    IconComponent.propTypes = {
       /** Optional class name. */
       className: PropTypes.string,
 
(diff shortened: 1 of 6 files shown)


jendowns:
@cameroncalder could you provide an SVG for that?

cameroncalder:
@jendowns yes, here ya go. It'll be 16x16

[high-decorator-icon.zip](https://github.com/carbon-design-system/ibm-security/files/4732981/high-decorator-icon.zip)

jendowns:
Thanks @cameroncalder! This has been updated. Could you take another look when you get a chance?

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -170,4 +170,53 @@ Decorator.defaultProps = {
       : 'No score',
 };
 
+/**
+ * Generate exports for individual severity icon types.
+ * @param {array} iconNames array of icon names
+ */
+function generateIconExports(...iconNames) {
```
SimonFinney, line 177: 👏

SimonFinney:
This is awesome! The code and tests LGTM, but I noticed a couple of things:

Decorator

We lose the `propType` information here presumably because the icons are dynamically generated - I don't foresee this as being a huge issue because the code sample provides enough clarity into usage, but I'd like to hear your thoughts?

The other was a requirement in the originating issue (https://github.com/carbon-design-system/ibm-security/issues/479#issue-596534931):

`Component should include the shape and a label to the right displaying the number value`

Am I overlooking the label somewhere? I'm not exactly sure what's meant by the 'number' value, but I'm assuming it's the score - perhaps @cameroncalder can provide some background.

jendowns:
@SimonFinney These icon components aren't aware of the "number" value for severity -- they just display whatever svg path is associated with their name, and I really think that's how it should be.

I think it's beneficial for us to be flexible here, and avoid being prescriptive about text next to the icon (similar to what is done with the "generic" `Icon` component).

It looks like @cameroncalder was able to look at this PR and approve it last week: https://github.com/carbon-design-system/ibm-security/pull/580#pullrequestreview-425419934

SimonFinney:
@jendowns That's fair, but I ask as it was communicated as a requirement, and as I understand we have folks dependent on the work! If it's not necessary, then we can disregard it.

PR #587, opened 2020-06-04
fix(tearsheet): add responsive for small variants

### LINKED ISSUE #482: Small tearsheet: responsive layout
### Details

- **Application / team -** IBM Security Data Explorer
- **Work item -** n/a
- **Needed by -**n/a

### Proposal

When a user views a tearsheet on an extra-large browser window (see screenshot), the majority of the viewport is taken up by negative space and the tearsheet, which should be the focus of the user's attention, is minimal and hard to interact with. I propose an exploration in responsiveness to address this problem.

image

### Goals / use cases

- [ ] Responsiveness for large screens
- [ ] Responsiveness for small screens

SimonFinney:
## Affected issue

Resolves #482

## Proposed changes

- Add responsive variants for `TearsheetSmall`
- Add custom breakpoint for larger screens
- Update tearsheet base to conform with requirements


SimonFinney:
@petervachon We're currently using the maximum breakpoint here, we could potentially include another one if we feel we have a unique use case? https://github.com/carbon-design-system/carbon/blob/master/packages/layout/scss/_breakpoint.scss#L52

SimonFinney:
I agree, it looks strange - this is likely because we have a fluid width of `50%` but it should be fixed.

PR #588, opened 2020-06-04
chore(storybook): filter stories

### LINKED ISSUE #573: Generate missing Storybook demos
### Proposal

We import and then export a number of components from our upstream library, Carbon.

All of these components have existing stories in Carbon's library. Usually, unless we have made changes to a component and need to reflect those changes in a specific demo, we just copy Carbon's story for that particular component.

It would be ideal if there was a way to iterate through our component exports, and determine if we've created an existing storybook demo. If we don't have a storybook demo for that exported component, then we could generate a basic demo and/or link to Carbon's existing demo.

This would not apply to Security org specific components, since we write those & have a requirement to create unique storybook demos for them. This would apply to components that get added to upstream Carbon (for example: #546)

SimonFinney:
## Affected issue

Resolves #573

## Proposed changes

- Identify list of stories from `carbon-components-react` that are not currently available in `@carbon/ibm-security`
- Verify list can be imported correctly and generate status messages
- Generate proxy stories that can be processed by Storybook environment using a manual script for now


src/components/stories/Textarea.stories.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * @file Textarea stories.
+ * @copyright IBM Security 2020
+ */
+
+export Textarea from 'carbon-components-react/es/components/TextArea/Textarea-story';
```
jendowns, line 6: Existing `TextArea` story: https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/TextArea/TextArea.stories.js

src/components/stories/DataTableSkeleton.stories.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * @file DataTableSkeleton stories.
+ * @copyright IBM Security 2020
+ */
+
+export DataTableSkeleton from 'carbon-components-react/es/components/DataTableSkeleton/DataTableSkeleton-story';
```
jendowns, line 6: Existing `DataTableSkeleton` story: https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/DataTable/stories/with-skeleton.js

SimonFinney:
Thanks @jendowns , that's a good point - Because Carbon doesn't explicitly define categories for their stories, it looks like Storybook will default any top-level stories to 'Other'.

I'll investigate if it's possible to rename this via global configuration, but otherwise, we may need to follow a similar approach https://github.com/IBM/carbon-addons-iot-react/blob/master/.storybook/manager-head.html#L15-L21

jendowns:
@SimonFinney Would it be possible to remove the story source plugin for these specific stories? Or even better: link to the Carbon story source code inside that plugin tab. The reason I bring this up is because that plugin content never loads for me... it just displays the loading message. I guess this is happening because of the way we are importing these stories... so it may be unavoidable. But with a message or something like that, the experience won't be frustrating/confusing for users.

On the topic of categories and "Other" -- I'm totally okay with removing the existing categories we have, if that's what you are asking? I'm super familiar with the list of components at this point, but I still find myself scrolling up and down to find `IconButton` versus `IconButtonBar` etc. 😆  I think it would be a better experience to just have everything in one category (that is, no category) so all components are in alphabetical order. 😅 What do you prefer?

jendowns:
Hey @SimonFinney did you have any thoughts on https://github.com/carbon-design-system/ibm-security/pull/588#issuecomment-651269053?

SimonFinney:
Thanks for your patience - Progress on this needs to be deprioritised for other efforts

PR #604, opened 2020-06-10
fix(data-table): spread props to absorb changes from Carbon

SimonFinney:
## Proposed changes

- Spread `DataTable` props to absorb changes from Carbon
- Add knob to toggle `stickyHeader`
- Update stories to accept relevant props


SimonFinney:
@jendowns I just pushed an update for the stories so they accept props coming in from the knobs - let me know your thoughts!

PR #606, opened 2020-06-11
feat(panelv2): click outside or use esc key to close panel

### LINKED ISSUE #481: Panel: Close panel by clicking "off" of panel
### Details

- **Application / team -** IBM Security Data Explorer
- **Work item -** n/a
- **Needed by -** n/a

### Proposal

It would make the Panel v2 component a bit more "delightful" and usable if a user could click on the overlay (or anywhere outside of the panel) to close the panel, in addition to clicking the close button.

#### To close the panel, a user can

- [ ] click the overlay
- [ ] use the 'Esc' key

#### Other

- [ ] Functionality is on by default and toggled by a prop

jendowns:
## Affected issues

- Resolves #481 

## Proposed changes

- make `PanelV2` a functional component
- instead of tracking `isOpen` in state, use `onClose` to handle closing the panel with the escape key, close button, or outside-of-panel clicks
- add `onOverlayClick` func prop to `Portal` so that `PanelV2` can pass `onClose` to it.

## Testing instructions

In the `PanelV2` story, open a panel and try clicking outside the panel or using the escape key to close the panel.

### DIFF
--- a/src/components/PanelV2/PanelV2.js
+++ b/src/components/PanelV2/PanelV2.js
@@ -7,7 +7,7 @@
 
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
-import React, { Component, createRef, Fragment } from 'react';
+import React, { createRef, Fragment, useState, useEffect } from 'react';
 import Close20 from '@carbon/icons-react/lib/close/20';
 
 import deprecate from 'carbon-components-react/lib/prop-types/deprecate';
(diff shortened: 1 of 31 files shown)


src/components/Portal/Portal.js
```diff
@@ -109,6 +109,7 @@ class Portal extends Component {
         this.overlay.setAttribute('tabIndex', '-1');
         this.overlay.classList.add(`${namespace}__overlay`);
         rootNode.appendChild(this.overlay);
+        this.overlay.addEventListener('mousedown', this.handleOverlayClick);
```
SimonFinney, line 112: I wonder if we should only apply this event listener if `onOverlayClick` has been defined by the consumer, to avoid it always executing an empty method. We may be able to prevent this by setting the `defaultProp` to `null` or `undefined` and wrapping this line in a conditional.

SimonFinney: For clean up, we'll also need to ensure the DOM event is removed when the component unmounts.

src/components/PanelV2/PanelV2.js
```diff
@@ -32,80 +32,73 @@ const getInstanceId = setupGetInstanceId();
...
-      this.setBodyMargin();
-    }
-  }
-
-  componentDidUpdate(prevProps, prevState) {
-    if (this.state.isOpen && !prevState.isOpen && isClient()) {
-      this.setBodyMargin();
-    }
-  }
+function PanelV2({
```
SimonFinney, line 35: 🔥

src/components/PanelV2/PanelV2.js
```diff
@@ -119,47 +112,50 @@ export default class PanelV2 extends Component {
...
             stopPropagationEvents={stopPropagationEvents}
+            onOverlayClick={onClose}
           >
+            {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
             <section
               className={classnames(namespace, className)}
               role="dialog"
               aria-label={ariaLabel}
               aria-modal="true"
+              onKeyDown={handleKeyDown}
```
SimonFinney, line 128: One thing I've noticed here that is potentially problematic is that if a user clicks a non-interactive element of the `Panel` while it's open, for example selecting the text, the `onKeyDown` handler won't trigger:

Panel

jendowns: Good catch! Looking into this, I had to add `tabIndex={-1}`. 👀 

From MDN:

> Keyboard events are only generated by <inputs>, <textarea> and anything with the contentEditable attribute or with tabindex="-1".

https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event

The `Modal` appears to take the same approach with `tabIndex`, etc.

jendowns:
Thanks @SimonFinney! This PR has been updated 👍

SimonFinney:
Awesome! The code LGTM, but when I click inside the panel in Chrome, it adds the default focus state -

Panel

Do we want to remove it or should we leave it as-is?

SimonFinney:
Sorry - it just occurred to me, we might want a design review for this

PR #610, opened 2020-06-11
fix(tearsheet-small): pass through isLoading, provide examples for making content non-interactive when loading

### LINKED ISSUE #601: TearsheetSmall is editable using 'Tab' key with loading
## Bug
TearsheetSmall is editable using 'Tab' key with loading

This bug is the same issue with the Tearsheet #563

jendowns:
## Affected issues

- Resolves #601 

## Proposed changes

Similar approach as #576 -
- pass through `isLoading` with loading state so that it can be used to making interactive elements non-interactive while `TearsheetSmall` is loading
- provide guidance for using `isLoading`
- make sure that all `TearsheetSmall` buttons are set to `disabled={true}` whenever loading state is active, so that they cannot be tabbed to in the background

Cleanup:
- in the `Tearsheet`, replace usage of `tabIndex` with the `disabled` attribute to make elements non-interactive during loading state

### DIFF
--- a/src/components/Tearsheet/TearsheetSmall/TearsheetSmall.stories.js
+++ b/src/components/Tearsheet/TearsheetSmall/TearsheetSmall.stories.js
@@ -177,7 +177,7 @@ storiesOf(patterns('TearsheetSmall'), module)
         isOpen
         loading={loading}
         labels={labels}
-        body={({ ...isLoading }) => (
+        body={isLoading => (
           <>
             <p>
               Whenever the TearsheetSmall is loading, please use{' '}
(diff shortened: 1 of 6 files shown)


src/components/Tearsheet/TearsheetSmall/TearsheetSmall.stories.js
```diff
@@ -172,6 +177,64 @@ storiesOf(patterns('TearsheetSmall'), module)
         isOpen
         loading={loading}
         labels={labels}
+        body={({ ...isLoading }) => (
```
SimonFinney, line 180:
```suggestion
        body={({ isLoading }) => (
```
I believe this should apply the correct `tabIndex` for `CodeSnippet`s, as the current deploy is outputting `-1`:

Tearsheet

jendowns:
Thanks @SimonFinney! This has been updated 👍

PR #619, opened 2020-06-16
docs(icon-button-bar): include example with box-shadow between icons

### LINKED ISSUE #568: Divider between buttons in IconButtonBar
### Details

- **Application / team -** Advisor

### Proposal

The first two buttons in our IconButtonBar is used to switch between views. The third button can be used on either view, so we've added a "divider" before the third button. This divider is actually just a left border on the third button. It would be nice to have an actual divider element between the buttons because on focus the buttons' border covers the "divider".

image

### Goals / use cases

- [ ] Be able to add a divider between buttons

jendowns:
## Affected issues

- Resolves #568 

## Proposed changes

- demonstrate how to add a divide between two icon buttons (not including the overflow menu button)

NOTE: a divider on either side of the overflow menu button would need a style override, since the overflow button is different from the ordinary icon buttons in this component)

### DIFF
--- a/src/components/IconButtonBar/IconButtonBar.stories.js
+++ b/src/components/IconButtonBar/IconButtonBar.stories.js
@@ -8,6 +8,7 @@ import { storiesOf } from '@storybook/react';
 import { select, number, boolean } from '@storybook/addon-knobs';
 
 import React from 'react';
+import { gray } from '@carbon/colors';
 import Add16 from '@carbon/icons-react/lib/add/16';
 import Add20 from '@carbon/icons-react/lib/add/20';
 import Add24 from '@carbon/icons-react/lib/add/24';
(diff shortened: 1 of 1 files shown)


SimonFinney:
Thanks @jendowns ! The main thing I wanted to raise was that in the original ask, the divider should not be covered by the focus state (https://github.com/carbon-design-system/ibm-security/issues/568#issue-625085582), but in @cameroncalder 's example, it looks like it will, so we likely need to decide on the intended behavior here.

/cc @carbon-design-system/ibm-security-design for further clarification and review.

jendowns:
@SimonFinney @cameroncalder If we don't want to recommend a type/style of divider, I'm happy to close this issue. But either way it would be helpful if @carbon-design-system/ibm-security-design could follow up in the originating issue #568 with either a clear indication that there is not a standard divider, or with a divider recommendation that I could then include in this PR for the demo.

cameroncalder:
@jendowns Let's include pixel spacers on each side of a divider line to separate from the icon ghost buttons

jendowns:
Closing for a different approach

PR #630, opened 2020-06-22
feat(icon-button-bar): add divider prop with demo and tests

### LINKED ISSUE #568: Divider between buttons in IconButtonBar

jendowns:
## Affected issues

- Resolves #568 

## Proposed changes

- add `divider` prop that renders a divider element on either side of an icon button, depending on prop value
- add storybook knobs/demo for new prop
- add tests for new prop

## Testing instructions

-

### DIFF
--- a/src/components/IconButtonBar/IconButtonBar.js
+++ b/src/components/IconButtonBar/IconButtonBar.js
@@ -45,7 +45,7 @@ const IconButtonBar = ({
     <Fragment key={action.label || `${namespace}__button--icon--${index}`}>
       {(action.divider === 'left' || action.divider === 'sides') && (
         <span className={`${namespace}__divider`} aria-hidden>
-          <span className={`${namespace}__divider__inner`} aria-hidden />
+          <span className={`${namespace}__divider__inner`} />
         </span>
       )}
       <IconButton
@@ -56,7 +56,7 @@ const IconButtonBar = ({
       />
       {(action.divider === 'right' || action.divider === 'sides') && (
         <span className={`${namespace}__divider`} aria-hidden>
-          <span className={`${namespace}__divider__inner`} aria-hidden />
+          <span className={`${namespace}__divider__inner`} />
         </span>
       )}
     </Fragment>
--- a/src/components/IconButtonBar/IconButtonBar.js
+++ b/src/components/IconButtonBar/IconButtonBar.js
@@ -42,13 +42,25 @@ const IconButtonBar = ({
   );
 
   const renderIconButton = action => (
-    <IconButton
-      {...action}
-      key={action.label || `${namespace}__button--icon`}
-      size={size}
-      tooltip={tooltip}
-      tooltipDirection={iconTooltipDirection}
-    />
+    <>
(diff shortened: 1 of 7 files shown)


src/components/IconButtonBar/IconButtonBar.js
```diff
@@ -41,14 +41,25 @@ const IconButtonBar = ({
...
-      key={action.label || `${namespace}__button--icon`}
-      size={size}
-      tooltip={tooltip}
-      tooltipDirection={iconTooltipDirection}
-    />
+  const renderIconButton = (action, index) => (
+    <Fragment key={action.label || `${namespace}__button--icon--${index}`}>
+      {(action.divider === 'left' || action.divider === 'sides') && (
+        <span className={`${namespace}__divider`} aria-hidden>
+          <span className={`${namespace}__divider__inner`} aria-hidden />
```
SimonFinney, line 48: Awesome! As I understand it, the `aria-hidden` declaration on the parent `divider` element should be sufficient to remove this from the accessibility tree.

... Unless different screen readers handle this in different ways? What's the rationale for adding it to the `inner` element?

src/components/IconButtonBar/IconButtonBar.stories.js
```diff
@@ -35,11 +35,16 @@ import { TooltipDirection } from '../IconButton/IconButton';
 storiesOf(patterns('IconButtonBar'), module).add(
   'default',
   () => {
-    const size = select('size', sizes, sizes[0]);
+    const size = select('size', sizes, 'sm');
     /* eslint-disable no-nested-ternary */
     const actions = [
       {
         className,
+        divider: select(
```
SimonFinney, line 43: Would you mind including a `null` / `undefined` option to provide an example of the `IconButtonBar` without any dividers?

src/components/IconButtonBar/_index.scss
```diff
@@ -49,4 +49,16 @@
       width: $icon-button-bar__sizing__dimensions;
     }
   }
+
+  &__divider {
+    background-color: transparent;
```
SimonFinney, line 54: Perhaps I'm overlooking something, but at first glance, this appears to be redundant - are we able to safely remove this or does it cover something specific?

jendowns: This style rule is for clarity and comprehension.

jendowns:
Thanks @SimonFinney this is ready to review again 👍

src/components/IconButtonBar/_index.scss
SimonFinney: If it's for clarity for subsequent developers, would it be more suitable to indicate this via a comment? That way it wouldn't be included in the build output.

PR #634, opened 2020-06-23
fix(summary-card): prevent `z-index` leak

### LINKED ISSUE #633: SummaryCard - icon ghost button layering
The SummaryCard icon ghost button’s z-index appears to be off compared to the header.
Referencing the Detail page layout example:
https://layout--ibm-security.netlify.app/?path=/story/patterns-unstable-layout--detail

SummaryCard

SimonFinney:
## Affected issue

Resolves #633

## Proposed change

Add `z-index` to `SummaryCardFooter` to prevent `z-index` from leaking out

### DIFF
--- a/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
+++ b/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
@@ -14573,6 +14573,7 @@ input:not(output):not([data-invalid]):-moz-ui-invalid {
   display: flex;
   background-color: var(--ui-01, #262626);
   border-top: 1px solid var(--ui-02, #393939);
+  z-index: 0;
 }
 
 .security--summary-card__select {
--- a/src/components/SummaryCard/_mixins.scss
+++ b/src/components/SummaryCard/_mixins.scss
@@ -1,7 +1,7 @@
 ////
 /// Summary card mixins.
 /// @group summary-card
-/// @copyright IBM Security 2019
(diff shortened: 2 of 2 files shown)


SimonFinney:
I like this approach because it scopes the `z-index` order locally to the summary card's footer - I believe it's originally in place to support how it appears visually when the content animates in, but my gut is that it doesn't need to compete with any other 'global' `z-index` on the screen.

Granted it probably doesn't need a value of `8001` (I wasn't sure about the intention behind this), so let me see if I can clean this up a bit!

jendowns:
Do you mind lowering that 8001 computed value?

SimonFinney:
I think it's necessary because of how this is currently set up:

1. The footer overlay uses Carbon's overlay `z-index` value - https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/SummaryCard/_mixins.scss#L48-L53
2. The action is required to be on top of the overlay to maintain the animation - https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/SummaryCard/_mixins.scss#L127-L128
3. There's also a transparent overlay that's set to 1 less than the overlay `z-index` - https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/SummaryCard/_mixins.scss#L259-L260

It feels like these all need to work together to maintain how it all works together visually - is this correct?

jendowns:
For sure, so the intent behind my question is: does it need to be so high, especially considering the change you made to the z-index of the footer?

PR #644, opened 2020-06-30
Expose (Data)Decorator onContextMenu prop

sbutler2901:
@jendowns the most important aspect of this is the ability to include the `onContextMenu` prop for the `DataDecorator` and `Decorator` components. The overriding of the `DataDecorator`'s `onClick` can be removed if its not desirable.

### DIFF
--- a/src/components/DataDecorator/DataDecorator.js
+++ b/src/components/DataDecorator/DataDecorator.js
@@ -55,7 +55,21 @@ class DataDecorator extends Component {
       title,
       scoreThresholds,
       scoreDescription,
+      onClick: propOnClick,
+      onContextMenu: propOnContextMenu,
     } = this.props;
+    const onClick = !propOnClick
+      ? event => {
+          event.stopPropagation();
+          this.toggleOpen();
+        }
+      : propOnClick;
+    const onContextMenu = propOnContextMenu
+      ? event => {
+          event.preventDefault();
+          propOnContextMenu(event);
+        }
+      : undefined;
     const decoratorProps = {
       className,
       inline,
@@ -66,6 +80,8 @@ class DataDecorator extends Component {
       title,
       scoreThresholds,
       scoreDescription,
+      onClick,
+      onContextMenu,
     };
     const componentLabels = {
       ...defaultLabels.labels,
@@ -80,14 +96,7 @@ class DataDecorator extends Component {
     };
     return (
       <Fragment>
-        <Decorator
-          {...decoratorProps}
-          active={this.state.isOpen}
-          onClick={event => {
-            event.stopPropagation();
-            this.toggleOpen();
-          }}
-        />
+        <Decorator {...decoratorProps} active={this.state.isOpen} />
         <PanelV2
           isOpen={this.state.isOpen}
           stopPropagation={stopPropagation}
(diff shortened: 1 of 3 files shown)


dbrugger:
looks like snapshots need to be updated

src/components/DataDecorator/DataDecorator.js
```diff
@@ -55,7 +55,21 @@ class DataDecorator extends Component {
...
     } = this.props;
+    const onClick = !propOnClick
+      ? event => {
+          event.stopPropagation();
+          this.toggleOpen();
+        }
+      : propOnClick;
+    const onContextMenu = propOnContextMenu
+      ? event => {
+          event.preventDefault();
```
jendowns, line 69: Why is right-clicking being prevented? Is it being replaced with something?

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -86,9 +91,13 @@ class Decorator extends Component {
       );
     }
 
-    if (onClick) {
+    if (onClick || onContextMenu) {
```
jendowns, line 94: To me it should only render as a button if `onClick` is provided, otherwise someone could provide just `onContextMenu` and render a `Decorator` that as an interactive element (a button) but cannot actually be left-clicked or activated with the `space` or `enter` keys. That seems like it would cause problems for folks using assistive tech?

src/components/DataDecorator/DataDecorator.js
```diff
@@ -55,7 +55,21 @@ class DataDecorator extends Component {
...
       scoreDescription,
+      onClick: propOnClick,
+      onContextMenu: propOnContextMenu,
     } = this.props;
+    const onClick = !propOnClick
+      ? event => {
+          event.stopPropagation();
+          this.toggleOpen();
+        }
+      : propOnClick;
```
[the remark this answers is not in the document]
sbutler2901: For the stopPropagation I'm not 100% why it was needed. In my testing everything looks fine if I include a basic logging function for `onClick` with propagation _not_ stopped. I'd have to do a rebuild of the addons with it removed from the default implementation of `onClick` to see if it has something to do with toggling the panel.

So, I originally changed this because I thought there was a use case for a consumer to override the default behavior completely, i.e., opening the panel! But as I mentioned, this is no longer needed so the change could completely be removed.

src/components/DataDecorator/DataDecorator.js
sbutler2901: Yes! A custom secondary click menu is being created so the default browser one doesn't need to be displayed.

src/components/DataDecorator/Decorator/Decorator.js
sbutler2901: I will look at this! My use case only requires the ability to provide a `onContextMenu`  propagated through the `DataDecorator` which always provides an `onClick`.

src/components/DataDecorator/DataDecorator.js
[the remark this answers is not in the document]
sbutler2901: I removed the option to override this default.

src/components/DataDecorator/Decorator/Decorator.js
```diff
@@ -160,6 +171,7 @@ Decorator.defaultProps = {
   href: undefined,
   inline: false,
   onClick: undefined,
+  onContextMenu: () => {},
```
sbutler2901, line 174: @jendowns I went this route to prevent a TypeError when `this.handleContextMenuClick` is called. Alternatively, a check could be add to see if the prop exists if you prefer:
```
onContextMenu={onContextMenu && this.handleContextMenuClick}
```

Issue #651, opened 2020-07-04
Allow addition of tags through the TagWallFilter


dan-cooke:
## Request - TagWallFilter - with additions
It would be cool if we could add new tags via the `TagWallFilter`

A lot of UIs provide this kind of workflow, and I cannot see any way in carbon at the moment.

Examples:

Github
![image](https://user-images.githubusercontent.com/22816887/86513913-db34b580-be05-11ea-8f87-c31f461242b9.png)

`semantic-ui`[ Dropdown with addition](https://react.semantic-ui.com/modules/dropdown/#usage-allow-additions)

![image](https://user-images.githubusercontent.com/22816887/86513929-f3a4d000-be05-11ea-8078-2a33818e812a.png)

### Proposal
![image](https://user-images.githubusercontent.com/22816887/86513969-48e0e180-be06-11ea-8f01-5d05b57ff988.png)

I just added this text via dev tools, so its not very pretty, but I think a simple clickable option to add the undiscovered searchh item would be a nice workflow.

PR #652, opened 2020-07-04
feat(tag-wall-filter): allow additions of new tags

dan-cooke:
ezgif com-video-to-gif (12)

- This PR allows users to add their own tags not specified in the initial props

## Affected issues

- Resolves https://github.com/carbon-design-system/ibm-security/issues/651

## Proposed changes

- New prop `allowAdditions` added to `TagWallFilter` and passed through to `Filter` .
    - When this prop is `true` and the user has entered a string that is not found in the available items, a new option will be available to allow the user to add this tag to the list of tags.

- New prop `newItemPrefix` - configurable prefix for when a new item is added

## Testing instructions

- Check out the new story for `TagWallFilter` - with additions

1. Type in a random string
2. Click the "Add: <string>"
3. Note the actions

### DIFF
--- a/src/components/TagWallFilter/Filter/Filter.js
+++ b/src/components/TagWallFilter/Filter/Filter.js
@@ -109,6 +109,19 @@ class Filter extends React.Component {
 
     const { inputValue } = this.state;
 
+    const renderInnerMenuItem = ({ itemProps, itemText }) => (
+      <div
+        className={`${namespace}__list-item__entry`}
+        aria-labelledby={itemProps.id}
+      >
+        <span className={`${carbonPrefix}text-truncate--end`} title={itemText}>
+          {itemText}
+        </span>
+        <span className={`${namespace}__add`}>
+          <Icon className={`${namespace}__add__icon`} renderIcon={Add20} />
+        </span>
+      </div>
+    );
     const menuItems = sortItems(
       filterItems(items, { itemToString, inputValue }),
       {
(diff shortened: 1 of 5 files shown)


src/components/TagWallFilter/TagWallFilter.stories.js
```diff
@@ -33,58 +31,34 @@ const buttons = {
...
-const filterItems = selectedItemsMap => ([items, item]) => {
-  if (item.id in selectedItemsMap) {
-    delete items[item.id];
-  }
-  return [items, item];
-};
-const selectFirst = ([items]) => items;
+storiesOf(patterns('TagWallFilter'), module)
+  .add(`TagWall with Filter list for multiselect`, () => {
+    const selectedItems = object('selectedItems', [items[0]]);
```
dan-cooke, line 36: I removed a lot of this complicated functional stuff here, as I think as a user of the storybook - its much easier to understand the API of the component if we just use a hard coded array.

Perhaps I regressed something, let me know

dan-cooke:
Also I'm not 100% sure on the API,

I think it might be more user friendly if we expose a `onAddition` callback - as most users will be using a backend to populate the list of available filters, if the user adds a new filter they will likely want to send this new filter to their backend.

To do this currently, they will have to filter the returned `onChange` lists to find any new tags themselves, not very ergonomic.

src/components/TagWallFilter/Filter/Filter.js
```diff
@@ -265,6 +286,11 @@ Filter.propTypes = {
     })
   ),
 
+  /**
+   * @type {string} The prefix string to add when no item has been found, e.g "Add: yourtag", the default prefix is "Add: "
+   */
+  newItemPrefix: PropTypes.string,
+  allowAddition: PropTypes.bool,
```
jendowns, line 293: Could you also add some proptype documentation for `allowAddition`?

src/components/TagWallFilter/TagWallFilter.js
```diff
@@ -276,6 +284,13 @@ TagWallFilter.propTypes = {
     })
   ),
 
+  /**
+   * @type {string} The prefix string to add when no item has been found, e.g "Add: yourtag", the default prefix is "Add: "
+   */
+  newItemPrefix: PropTypes.string,
+
+  allowAddition: PropTypes.bool,
```
jendowns, line 292: Could add some proptype documentation for `allowAddition`?

SimonFinney:
Tagging @petervachon for interaction review

dan-cooke:
@jendowns 
@petervachon 

Thanks for the feedback, those are some good points. I will address these soon hopefully!

PR #662, opened 2020-07-09
Data decorator panel root

sbutler2901:
## Proposed changes

- Adding DataDecorator `rootNode` prop proxy to PanelV2 component
- Adding DataDecorator `focusTrap` prop proxy to PanelV2 component
- Adding PanelV2 `rootNode` prop proxy to the Portal component

## Testing instructions

- Pass a rootNode (HTMLElement) to DataDecorator and ensure the panel is rendered within that node
- Disable the focus trap for the DataDecorator and verify it is no longer active


sbutler2901:
@jendowns For my use case, I found the focus trap to be broken when the Panel is render within a ShadowRoot. Maybe you could provide some debugging techniques for the focus trap so that I can avoid disabling it.

The close button and any intractable content within the panel would not receive events. For example, a Button rendered with the panel would visually respond to the click, but the event would not fire. This applies when clicking with a mouse or tabbing and using space.

In some test cases pressing escape would not close the panel and instead would allow the close button and other element's events to fire / be received.

SimonFinney:
Hi all! So I understand correctly, does this focus issue still persist if the root node is specified as one within the shadow DOM?

@jendowns We may need to investigate further to ensure focus trapping is compatible within these contexts - we can't sacrifice on accessibility for the purposes of enabling this use case, but it's a huge priority for proxy applications, and I believe we'll come across this issue with any other component that leverages the `Portal` and gets emitted by the shell.

sbutler2901:
Hey ya'll! I did some further testing of the various ibm-security components that use the Portal and they all suffer from the issue. I'm using `ibm-security/carbon-ibm-security-v1.27.1-prerelease.3.tgz` and found that the `Tearsheet`, `TearsheetSmall`, and `PanelV2` all require pressing `escape` to clear the focus trap so that interactions would be received. This is with all of them rendered in a shadow root. I did find that the base Modal component coming from Carbon directly did not have this issue so it has to be an issue with the Portal.

For now, can I request that we merge this PR? I am concerned about the accessibility issue as well, but the default value of the prop is `true` and all of the other aforementioned components support the focusTrap prop so I don't see this change being too divergent!

Once the Portal's focus trap is functional within a shadow root I will re-enable it in my usages, but I don't want this to be a blocker for my subsequent stories.

SimonFinney:
That seems like a fair deal to me - @jendowns what are your thoughts?

@sbutler2901 would you mind creating a follow-up issue for tracking so we can complete that work independently?

sbutler2901:
@SimonFinney @jendowns any more info about this?

I have created a simple issue for tracking the issue here: https://github.com/carbon-design-system/ibm-security/issues/700

Issue #664, opened 2020-07-10
Add support for 'size' prop in DataTable


Yohanna:
Add support for `size` prop in `DataTable`. The base Carbon library already supports this [option here](https://github.com/carbon-design-system/carbon/blob/fe23a6f46b001e56d31376cfebe6e523b27e77f9/packages/react/src/components/DataTable/DataTable.js#L111)

jendowns:
Hey @Yohanna, have you had issues using the `size` prop? While it isn't shown/used in the storybook demos, there's a spread attribute that will allow you to pass it in:
https://github.com/carbon-design-system/ibm-security/blob/6172791f986cbb5d6214ede59e49b7fe7b4c6f80/src/components/DataTable/DataTable.js#L49

Yohanna:
I tried adding a `size="compact"` on the `DataTable` component but it didn't have any effect.

Adding `className="bx--data-table--compact` on the `Table` component works though.

Yohanna:
Here's a repro of the problem: https://codesandbox.io/s/codesandbox-jrd6p?file=/src/index.js

It seems it doesn't work with the `render` prop pattern. If you comment out the `render` prop it works fine.

SimonFinney:
Hi @Yohanna - thanks for raising! You'll need to use the `getTableProps` parameter that's outputted from the `render` method - https://codesandbox.io/s/codesandbox-kubvz

Yohanna:
Hey @SimonFinney & @jendowns, `getTableProps` did the trick! Thanks!

Btw it looks like this wasn't working in v`1.25.0`, I had to upgrade to v`1.27.0` to get it to work.

jendowns:
Good to hear @Yohanna!

Also the versioning situation you mention makes sense. If you view the git history of the line I linked to above, it was added a month ago...
https://github.com/carbon-design-system/ibm-security/blame/6172791f986cbb5d6214ede59e49b7fe7b4c6f80/src/components/DataTable/DataTable.js#L49

...as part of a PR that was merged and released in `1.26.0`: https://github.com/carbon-design-system/ibm-security/pull/604#issuecomment-644634107

PR #666, opened 2020-07-12
fix(data-table): proxy `defaultProps` and `propTypes`

SimonFinney:
## Affected issue

Related to #664

## Proposed changes

- Proxy Carbon's `DataTable` `defaultProps` and `propTypes` - any changes upstream were not being reflected and would cause disparities between both implementations and consumer expectation
- Update snapshots to reflect changes
- Add `size` knob to `DataTable` stories

### DIFF
--- a/src/components/DataTable/DataTable.stories.js
+++ b/src/components/DataTable/DataTable.stories.js
@@ -8,10 +8,14 @@ import { storiesOf } from '@storybook/react';
 
 import { components } from '../../../.storybook';
 
-const sizes = [null, 'compact', 'short', 'tall'];
+import { DataTable } from '../..';
+
+const {
+  defaultProps: { size },
+} = DataTable;
 
 const props = () => ({
-  size: select('Row height (size)', sizes, sizes[0]),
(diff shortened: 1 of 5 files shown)


src/components/DataTable/DataTable.stories.js
```diff
@@ -3,14 +3,17 @@
...
+import { boolean, select } from '@storybook/addon-knobs';
 import { storiesOf } from '@storybook/react';
 
 import { components } from '../../../.storybook';
 
+const sizes = [null, 'compact', 'short', 'tall'];
+
 const props = () => ({
-  useZebraStyles: boolean('Zebra row styles (useZebraStyles)', false),
+  size: select('Row height (size)', sizes, sizes[0]),
```
jendowns, line 14: Since sizes is just being used once, can the size values be included where they are referenced in line 14? In fact, I was wondering if this can be rewritten similar to:

https://github.com/carbon-design-system/carbon/blob/2692813cc53972ef4a460a51d87b7e45285ca742/packages/react/src/components/DataTable/DataTable-story.js#L17-L21

SimonFinney:
Revisiting the scope of this as deeper issues uncovered

PR #667, opened 2020-07-13
fix(scroll-gradient): use `ResizeObserver`

### LINKED ISSUE #626: ScrollGradient not shown in DataTable
## Bug - ScrollGradient does not immediately appear in DataTable when DataTable is initially behind a closed Tab

**Expected behavior -**

ScrollGradient should appear, if needed, when the tab is opened that contains the DataTable

**Actual behavior -**

ScrollGradient is not shown when a tab is opened that contains DataTable. It appears only after the scrollbar is interacted with or the window size is adjusted.

#### 

image

SimonFinney:
## Affected issue

Resolves #626

## Proposed changes

- Replace `resize` event listener with `ResizeObserver` to accommodate situations where the `ScrollGradient` has been mounted but is not visible, for example, the `Tabs` component
- Add polyfill to be conditionally used based on browser support
- Update `ScrollGradient` unit tests
- Add `ResizeObserver` mocks to testing suite
- Clean up Jest configuration


src/components/ScrollGradient/ScrollGradient.js
```diff
@@ -96,7 +97,10 @@ class ScrollGradient extends Component {
     }
 
     if (isClient()) {
-      window.addEventListener('resize', this.updateHandler);
+      // https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
+      this.resizeObserver = new ResizeObserver(this.updateHandler);
```
jendowns, line 101: This is very cool 👏 

One concern I have is about browser support though. I'm looking at the support table on CanIUse and the table is not as green as I'd hoped: https://caniuse.com/#feat=resizeobserver
(For example, Firefox as v68 / July 2019... and Safari as of v13 / Sept 2019 -- that seems very close, < 1 year. 😰)

SimonFinney: Good catch! I've added a polyfill here to better support this.

jendowns: I realize that polyfill gets downloaded a lot but did you take a look at the build history / repo? Looks like the build is failing and there hasn't been an update in 2 years. Are you concerned about support for this new dependency?

config/jest/index.js
```diff
@@ -24,3 +26,5 @@ const toHaveNoDAPViolations = require('./matchers/toHaveNoDAPViolations');
 // For more information, check out the docs here:
 // https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array
 expect.extend({ toHaveNoAxeViolations, toHaveNoDAPViolations });
+
+global.ResizeObserver = resizeObserverMock;
```
diego-codes, line 30: Instead of adding this as a global object for all tests, I would just mock it for the gradient tests:


```js
const observeMock = jest.fn();
const disconnectMock = jest.fn();

beforeAll(() => {  
    window.ResizeObserver = jest.fn(function ResizeObserver() {
      this.observe = observeMock;
      this.disconnect = disconnectMock;
  });
});

beforeEach(() => {
  observeMock.restoreMock();
  disconnectMock.restoreMock();
});

afterAll(() => {
  window.ResizeObserver = undefined;
});
```

By doing this, you don't need to dig deep into the structure of a mock function that relies on the order of tests.

src/components/ScrollGradient/__tests__/ScrollGradient.spec.js
```diff
@@ -58,22 +60,20 @@ describe('ScrollGradient', () => {
...
+    it('adds `ResizeObserver` when mounted', () => {
       scrollGradient = mount(
         <ScrollGradient className={className} color="blue">
           {children}
         </ScrollGradient>
       );
-      expect(spy).toHaveBeenCalledWith('resize', expect.any(Function));
-      spy.mockRestore();
+
+      expect(resizeObserverMock.mock.instances[6].observe).toBeCalled();
```
diego-codes, line 70: With my previous comment, you would be able to write your assertion to something like:

```js
      expect(observeMock).toBeCalledTimes(1);
      // or
      expect(observeMock).toBeCalledWith(1);
```

config/jest/index.js
SimonFinney: Thanks! I much prefer this

SimonFinney: I had some challenges integrating this inline (as indicated as a comment in the code), but I've used your suggestion as a foundation for how this is structured - let me know your thoughts!

diego-codes: Yeah, I think that the changes you made improve the readability and maintainability of the tests 👍

src/components/ScrollGradient/ScrollGradient.js
SimonFinney: 🤷  I've been testing with a newer implementation (https://github.com/juggle/resize-observer) and so far this has been throwing errors in our testing suite as it also requires a dependency on `MutationObserver` polyfills (https://github.com/juggle/resize-observer/issues/64#issuecomment-636805189), because I believe JSDOM doesn't mock these either

PR #686, opened 2020-07-21
docs(FilterPanelCheckboxWithOverflowMenu): implement selectorPrimaryFocus instead of primaryFocus

jendowns:
## Affected issues

partially addresses #683 

## Proposed changes

- update `OverflowMenuItem` in `FilterPanelCheckboxWithOverflowMenu` story to implement `selectorPrimaryFocus` prop

### DIFF
--- a/src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.stories.js
+++ b/src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.stories.js
@@ -28,13 +28,13 @@ storiesOf(
       id="checkbox-id"
     >
       <OverflowMenuItem
-        selectorPrimaryFocus
         itemText="Option 1"
         onClick={action('OverflowMenuItem onClick')}
       />
       <OverflowMenuItem
-        itemText="Option 2"
+        itemText="Option 2 -- selected when menu opens"
         onClick={action('OverflowMenuItem onClick')}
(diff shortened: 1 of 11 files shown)


diego-codes:
Does the UI work as expected without the prop?

jendowns:
@diego-codes This PR is for a storybook demo.

diego-codes:
I know, I was just curious because it's unclear to me what the purpose of the prop is. So if the prop is totally optional, then would there be a point in keeping it for the previews which are what we rely on for documentation.

src/components/FilterPanel/FilterPanelCheckboxWithOverflowMenu/FilterPanelCheckboxWithOverflowMenu.stories.js
```diff
@@ -28,13 +28,13 @@ storiesOf(
...
       <OverflowMenuItem
-        primaryFocus
         itemText="Option 1"
         onClick={action('OverflowMenuItem onClick')}
       />
       <OverflowMenuItem
-        itemText="Option 2"
+        itemText="Option 2 -- selected when menu opens"
         onClick={action('OverflowMenuItem onClick')}
+        data-floating-menu-primary-focus
```
diego-codes, line 37: This isn't a prop that is documented, does it make sense to have it?

jendowns: It's the default value for `selectorPrimaryFocus` in `OverflowMenu`

PR #687, opened 2020-07-21
fix(icon-button-bar): use selectorPrimaryFocus in OverflowMenu

jendowns:
## Affected issues

partially addresses #683 

## Proposed changes

- use `selectorPrimaryFocus` in `OverflowMenu` (set it to a known class name for icon buttons)
- add `setFocus` prop to individual actions (assign it a class name if true)

## Testing instructions

- ensure that `IconButtonBar` functions as expected

### DIFF
--- a/src/components/IconButtonBar/IconButtonBar.js
+++ b/src/components/IconButtonBar/IconButtonBar.js
@@ -25,6 +25,7 @@ const IconButtonBar = ({
   overflowMenuDirection,
   size,
   tooltip,
+  selectorPrimaryFocus,
 }) => {
   const isMaxLength = actions.length > length;
   const iconButtonBarClasses = classnames(namespace, className, {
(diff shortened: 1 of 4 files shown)


src/components/IconButtonBar/IconButtonBar.js
```diff
@@ -73,6 +74,7 @@ const IconButtonBar = ({
         key={action.label}
         onClick={action.onClick}
         disabled={action.disabled}
+        {...action.otherProps}
```
SimonFinney, line 77: Is there an opportunity to spread these flat from the `action` as opposed to having them contained within an `otherProps` property? I'm unfamiliar with this pattern if it's in use elsewhere

jendowns: Consider that the `selectorPrimaryFocus` targets a selector in a child `OverflowMenuItem`, which can be anything the user wants.

PR #697, opened 2020-07-24
fix(toolbar): ensure that toolbar panels have ids, update aria attributes

### LINKED ISSUE #696: Toolbar uses invalid aria-describedby attributes
## Bug - the toolbar component's icons have `aria-describedby` attributes that don't point to anything

**Expected behavior -**

`aria-describedby` on one element should point to the ID of another element that describes it. 

**Actual behavior -**

The `aria-describedby` attributes on the sidebar icons are hardcoded, and point to elements that do not exist. 

### Steps for reproducing

This issue was seen in the `Shell` component, which relies on the `Toolbar` component. 

### Screenshots

Screen Shot 2020-07-24 at 11 51 00 AM
Screen Shot 2020-07-24 at 11 51 19 AM

### Fix

The line that needs to be removed is here: https://github.com/carbon-design-system/ibm-security/blob/a8eb63245b0b620a0ca832803836da11fff959bf/src/components/Toolbar/Toolbar.js#L94

The `aria-describedby`'s value (`type`) comes from these function calls: 

https://github.com/carbon-design-system/ibm-security/blob/a8eb63245b0b620a0ca832803836da11fff959bf/src/components/Toolbar/Toolbar.js#L222-L224

jendowns:
## Affected issues

- Resolves #696

## Proposed changes

- ensure that toolbar `aside` has valid `id`
- add `aria-expanded` boolean to toolbar buttons
- conditionally add `aria-control` to toolbar button if corresponding `aside` is opened

### DIFF
--- a/src/components/Toolbar/Toolbar.js
+++ b/src/components/Toolbar/Toolbar.js
@@ -91,7 +91,14 @@ export default class Toolbar extends Component {
     return (
       <IconButton
         className={iconButtonClasses}
-        aria-describedby={type}
+        aria-expanded={isActiveItem}
+        {
+          // Expanded panels are added to the UI when opened,
+          // so this should not reference an ID that doesn't yet exist.
+          ...(isActiveItem
+            ? { [`aria-controls`]: `${namespace}--toolbar--${type}` }
+            : {})
+        }
         aria-label={label}
         iconClassName={`${namespace}__icon`}
         label={label}
@@ -210,10 +217,14 @@ export default class Toolbar extends Component {
     const { menu, settings, support } = labels;
     const classes = classnames(namespace, className);
     const { isActive } = this.state;
-    const isPanelActive = Object.keys(isActive).some(type => {
-      const { [type]: isActiveItem } = isActive;
-      return isActiveItem;
-    });
+    const activeItems = Object.entries(isActive)
+      // eslint-disable-next-line no-unused-vars
+      .filter(([type, isActiveItem]) => isActiveItem);
+    let currentType = '';
+    let isPanelActive = false;
+    if (activeItems.length > 0) {
+      [[currentType, isPanelActive]] = activeItems;
+    }
 
     return (
       <div ref={this.wrapper}>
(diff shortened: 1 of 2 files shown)


jendowns:
Hey @benborgers -- looking at this & #696, I'm actually not sure that `aria-describedby` made sense in the first place. 🤔 

From [MDN's `aria-describedby` article](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute):

> The aria-describedby attribute is used to indicate the IDs of the elements that describe the object. It is used to establish a relationship between widgets or groups and text that described them. This is very similar to aria-labelledby: a label describes the essence of an object, while a description provides more information that the user might need.

In the current code, those toolbar icon buttons are marked as "described by" the opened panel content. 🤔  But the buttons have `aria-label`s already, so the "described by" relationship, especially when it is meant to reference an entire panel of content, doesn't seem meaningful.

Instead here I've added `aria-controls` to a toolbar button when its corresponding panel is open. From [w3's notes on `aria-controls`](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-controls):

> Identifies the element (or elements) whose contents or presence are controlled by the current element.

So the `aria-controls` relationship, to indicate that a toolbar button controls the opening/closing of panel content, seems more meaningful.

Did you have any thoughts about this?

benborgers:
@jendowns That sounds good! Yeah, I don't think the `aria-describedby` makes too much sense. However, I didn't know that the `aria-describedby` pointed to the open panel of content. When I looked, the IDs that `aria-describedby` was pointing to didn't exist at all. Do you have any idea why that might be?

jendowns:
@benborgers Honestly I'm making an assumption that the missing IDs were originally meant to go on each of the individual panels, based on a best guess... but I'm not 100% sure.

If you are cool with `aria-controls` instead (plus also the additional of the missing IDs, for correct aria referencing) I'll bring this PR out of draft. 👍

benborgers:
@jendowns Yep, `aria-controls` seems like a good solution! Thanks for taking a look at this.

PR #707, opened 2020-07-31
feat(focus-trap): update version, enable outside clicks

### LINKED ISSUE #703: Mouse Events not captured by Notification on click when PanelV2 is open.
## Bug - 

Expected behavior:
When both Panel and Notification are opened, both should be responding to mouse/keyboard interactions. 

Actual Behaviour:
When both Panel and Notification are opened, it seems that Panel capture all the mouse events and Notification does not respond to any mouse events.

Version:1.26.0

SimonFinney:
## Affected issues

Resolves #703 

## Proposed changes

- Update `focus-trap-react` dependency
- Enable `allowOutsideClick` option - https://github.com/davidtheclark/focus-trap#focustrap--createfocustrapelement-createoptions
- Add `focusTrapOptions` configuration object to `Portal`


jendowns:
@SimonFinney I noticed my review was requested but this particular PR is still in draft -- were you looking for a review?

SimonFinney:
Thanks @jendowns , that must've been residual before I converted the PR to draft - this is now ready for review

diego-codes:
@SimonFinney, what is the status of this PR?

SimonFinney:
@diego-codes It's awaiting review - If you have the bandwidth, please jump in and we can push it forward

src/components/Portal/Portal.js
```diff
@@ -203,6 +211,9 @@ Portal.propTypes = {
   /** @type {boolean} Focus trap. */
   focusTrap: PropTypes.bool,
 
+  /** Pass any of the options available in https://github.com/focus-trap/focus-trap#focustrap--createfocustrapelement-createoptions */
```
diego-codes, line 214: The link didn't work, however, this one did:
```suggestion
  /** Pass any of the options available in https://github.com/focus-trap/focus-trap#createfocustrapelement-createoptions */
```

src/components/Portal/Portal.js
```diff
@@ -175,15 +175,23 @@ class Portal extends Component {
...
     return (
       isClient() &&
       createPortal(
         <FocusTrap
           active={focusTrap}
           focusTrapOptions={{
+            allowOutsideClick: () => true,
+            fallbackFocus:
+              fallbackFocus ||
+              (isClient() ? rootNode : Children.toArray(children)[0].type),
```
diego-codes, line 192: Given that this is a major change in functionality, would it make sense to allow consumers to provide these options?

src/components/Portal/Portal.js
```diff
@@ -225,6 +236,7 @@ Portal.propTypes = {
 Portal.defaultProps = {
   children: null,
   focusTrap: true,
+  focusTrapOptions: null,
```
diego-codes, line 239: The default should probably be an empty object. I expect that if a consumer does not provide this prop, then this component will throw an error by trying to spread `null` here:

https://github.com/carbon-design-system/ibm-security/blob/6af2669dec0cac7fc90cf29d92c408145b2d01ea/src/components/Portal/Portal.js#L194

JennChao:
Hi @SimonFinney, 
Based on my understanding, this PR could fix the issue mentioned in https://github.com/carbon-design-system/ibm-security/issues/703
What is the status of this PR?

src/components/Portal/Portal.js
SimonFinney: I was torn about this - it seemed sensible to allow the outside click by default with the option to override it with `focusTrapOptions` so there would be no additional configuration to consumers, but perhaps that could have unintended side effects.

For the `fallbackFocus` I'm less sure what the right approach is here - with the package update that includes the configuration to fix this issue, the `focus-trap-react` package now expects a fallback or else it throws errors. What would be your recommendation here?

diego-codes: I would say that `allowOutsideClick` changes would be considered breaking changes.

For the `fallbackFocus`, I think that your approach is good considering that it is a required option.

SimonFinney: Good point! In that case, we can pass through the `focusTrapOptions` prop from `PanelV2` - would that satisfy the requirement for this issue?

SimonFinney:
Hi @JennChao ! Apologies, I was on vacation so thanks for your patience - I'm wrapping up the review changes for this PR, and I'm hoping we can push this through soon.

JennChao:
> Hi @JennChao ! Apologies, I was on vacation so thanks for your patience - I'm wrapping up the review changes for this PR, and I'm hoping we can push this through soon.

Thank you very much! :)

JennChao:
Hi @SimonFinney,
Any update for this PR? Is it ready to be merged?

SimonFinney:
@diego-codes Are you happy with the changes?

diego-codes (approved):
This looks good to me 👍

PR #709, opened 2020-08-04
fix(tag-wall-filter): fix input positioning

### LINKED ISSUE #702: Make Tag wall filter Search bar  fixed on top
## Request 
Fix Search bar on top

- **Short description -**

The search bar currently disappear when user scrolls down the list of options. It would help the users if the search bar was always visible for easy access. 
 

### Details
- **Application / team -** TII
- **Work item -**  -[]()
- **Needed by -** TII 

### Proposal
Make Tag wall filter search bar stay on top

SimonFinney:
## Affected issue

Resolves #702

## Proposed changes

- Fix `TagWallFilter` input positioning
- Refactor imports

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@carbon/ibm-security",
-  "version": "1.29.0-prerelease.7",
+  "version": "1.29.0",
   "description": "Carbon for IBM Security",
   "license": "Apache-2.0",
   "main": "lib/index.js",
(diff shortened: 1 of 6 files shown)


petervachon:
Looking good! Is there a way to introduce the gradient under the search input on scroll? We have the gradient at the bottom of the tearsheet indicating that there is additional content, I'm wondering if we should include the same effect at the top?

SimonFinney:
Good catch!

SimonFinney:
Thanks @petervachon - I've made the changes suggested, let me know your thoughts 🎉

PR #713, opened 2020-08-05
fix(toolbar): update panel semantics

### LINKED ISSUE #711: Incorrect use of <aside> in Toolbar
## Bug - the `aside` element cannot have `role="menu"`

Accessibility testers pointed out that the `aside` HTML element may not have `role="menu"` (the allowed roles are listed [in the first row of this table](https://fae.disability.illinois.edu/rulesets/ROLE_7/)). 

Looking into it further, I think the issue lies with the `aside` element, not the `role="menu"`. [This answer](https://stackoverflow.com/a/40399165) advises against it, saying: 

> `aside` is for "tangentially related" content, but the primary navigation is hardly only "tangentially" related to the document.

---

Here's the part of the `Toolbar` component: 

https://github.com/carbon-design-system/ibm-security/blob/66ab719ee54ed957adb1c791c464894d829ea569/src/components/Toolbar/Toolbar.js#L250-L280

### Optional information

**Version -** `1.28.0`

SimonFinney:
## Affected issues

Resolves #711

## Proposed changes

- Update `Toolbar` panel to remove role and use of `aside`
- Update snapshot

### DIFF
--- a/src/components/Toolbar/Toolbar.js
+++ b/src/components/Toolbar/Toolbar.js
@@ -247,10 +247,9 @@ export default class Toolbar extends Component {
 
         <Transition className={namespace} component="span">
           {isPanelActive && this.state.showContent ? (
-            <aside
-              className={`${namespace}__panel`}
+            <nav
               id={`${namespace}--toolbar--${currentType}`}
-              role="menu"
+              className={`${namespace}__panel`}
             >
               <IconButton
                 onClick={this.toggleContent}
@@ -260,13 +259,12 @@ export default class Toolbar extends Component {
                 className={`${namespace}__content`}
                 dangerouslySetInnerHTML={{ __html: this.state.content }} // eslint-disable-line react/no-danger
               />
-            </aside>
+            </nav>
           ) : (
             isPanelActive && (
-              <aside
-                className={`${namespace}__panel`}
+              <nav
                 id={`${namespace}--toolbar--${currentType}`}
-                role="menu"
+                className={`${namespace}__panel`}
               >
                 {Object.keys(isActive).map(type => (
                   <Transition
@@ -277,7 +275,7 @@ export default class Toolbar extends Component {
                     {this.renderContent(type)}
                   </Transition>
                 ))}
-              </aside>
+              </nav>
             )
           )}
         </Transition>
--- a/src/components/Toolbar/Toolbar.js
+++ b/src/components/Toolbar/Toolbar.js
@@ -247,7 +247,7 @@ export default class Toolbar extends Component {
 
         <Transition className={namespace} component="span">
           {isPanelActive && this.state.showContent ? (
-            <nav
+            <div
               id={`${namespace}--toolbar--${currentType}`}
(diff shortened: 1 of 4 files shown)


src/components/Toolbar/Toolbar.js
```diff
@@ -247,10 +247,9 @@ export default class Toolbar extends Component {
 
         <Transition className={namespace} component="span">
           {isPanelActive && this.state.showContent ? (
-            <aside
-              className={`${namespace}__panel`}
+            <nav
```
jendowns, line 250: Related to [IBM 2.4.1 Bypass Blocks](http://www.ibm.com/able/guidelines/ci162/bypass_blocks.html)

> Multiple elements with "nav" role do not have unique labels

Here's some notes from that checkpoint in the a11y checker plugin:

> What to do
> For each element with a "nav" role, either set the value of the aria-labelledby attribute to the id value of any visible text that appropriately and uniquely describes that navigation region;
> OR, use the aria-label attribute to add an invisible, unique, non-empty label to an element with the "nav" role.

In this case, it may be helpful to call out that this particular `nav` is related to whatever toolbar section ("Settings", etc.) opened it.

Also I think it would be nice to go ahead and add a label to the "primary" `nav` as well, containing those primary icon buttons (again, "Settings" etc).

SimonFinney: Sounds good, and I like the second suggestion! Feel free to take over if it's something you're interested in and have the bandwidth, otherwise, I'll likely be able to revisit later on in the week

src/components/Toolbar/Toolbar.js
```diff
@@ -247,10 +247,9 @@ export default class Toolbar extends Component {
 
         <Transition className={namespace} component="span">
           {isPanelActive && this.state.showContent ? (
-            <aside
-              className={`${namespace}__panel`}
+            <div
```
jendowns, line 250: @SimonFinney I noticed that you changed the `nav` element to a `div` -- while that removes the IBMa / WCAG warning, I question whether that's the best route to take here. Considering that this is a navigation list, it would be helpful to provide that info to assistive tech by using semantic markup.

Deriohig:
is there anything I can do to help this item move along?

SimonFinney:
Hi @Deriohig ! From what I can remember, I have an update for this fix that I can push for your review once I clean it up a bit. Alternatively, if you have a spare minute and would like to take it on, I can reassign you as the owner?

Deriohig:
> Hi @Deriohig ! From what I can remember, I have an update for this fix that I can push for your review once I clean it up a bit. Alternatively, if you have a spare minute and would like to take it on, I can reassign you as the owner?

sure I can work on this now, Thanks!

Deriohig:
@SimonFinney  not sure why the Netlify failed, says it was a queuing problem I dont have access to re-run it. I think I addressed all the relevant issues though.

jendowns (approved):
Thank you @Deriohig! 🎉

PR #714, opened 2020-08-05
feat(toolbar): support supplying IDs to navigation items

### LINKED ISSUE #710: Shell toolbar has duplicate IDs
## Bug - the toolbar of the `Shell` component uses the same ID for every link in the sub-menu, which isn't valid HTML. 

**Expected behavior -**

Every item has a unique ID. I believe an ID is already passed into the component for every link here: 

https://github.com/carbon-design-system/ibm-security/blob/66ab719ee54ed957adb1c791c464894d829ea569/src/components/Toolbar/Toolbar.js#L161

So maybe that could be appended to something to create the ID? 

**Actual behavior -**

Every ID in the list is `security--nav__list__item` (the default for the `NavItem` component), which results in duplicated IDs. 

Screen Shot 2020-08-04 at 12 03 14 PM

### Tracing

Looks like the ID would be passed into `NavItem` around here? 

https://github.com/carbon-design-system/ibm-security/blob/66ab719ee54ed957adb1c791c464894d829ea569/src/components/Toolbar/Toolbar.js#L174-L179

### Optional information

**Version -** `1.28.0`

SimonFinney:
## Affected issues

Resolves #710

## Proposed changes

- Add support for supplying IDs to toolbar navigation items
- Generate `NavItem` instance ID if none is supplied
- Refactor `Toolbar` mocks to be more flexible with props
- Update snapshot tests

### DIFF
--- a/src/components/Nav/NavItem/NavItem.js
+++ b/src/components/Nav/NavItem/NavItem.js
@@ -3,18 +3,20 @@
  * @copyright IBM Security 2019 - 2020
  */
 
-import Launch16 from '@carbon/icons-react/lib/launch/16';
-
+import { Launch16 } from '@carbon/icons-react';
+import setupGetInstanceId from 'carbon-components-react/lib/tools/setupGetInstanceId';
 import classnames from 'classnames';
 import { bool, elementType, func, node, number, string } from 'prop-types';
 import React, { Component } from 'react';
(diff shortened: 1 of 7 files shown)


Deriohig (approved):
Looks good to me.

PR #721, opened 2020-08-07
feat(combo-button): add menuOffset and menuOffsetFlip props for inner OverflowMenu

### LINKED ISSUE #720: Adding a menuOffset property to ComboButton to add custom styling
Adding a menuOffset property to ComboButton to add custom styling

 Details

Just like OverflowMenu , if a menuOffset property exists for ComboButton, we can apply some custom styling to the menu options. 
Without this, since our comboButton is rendered at the bottom most of the screen, the menu options are getting hidden.

jendowns:
## Affected issues

- Resolves #720

## Proposed changes

- add `menuOffset` prop with default, apply to inner `OverflowMenu`
- add `menuOffsetFlip` prop with default, apply to inner `OverflowMenu`

### DIFF
--- a/src/components/ComboButton/ComboButton.js
+++ b/src/components/ComboButton/ComboButton.js
@@ -175,8 +175,8 @@ ComboButton.propTypes = {
    */
   menuOffset: PropTypes.oneOfType([
     PropTypes.shape({
-      top: PropTypes.number,
-      left: PropTypes.number,
+      top: PropTypes.oneOf([PropTypes.number, PropTypes.string]),
+      left: PropTypes.oneOf([PropTypes.number, PropTypes.string]),
     }),
     PropTypes.func,
   ]),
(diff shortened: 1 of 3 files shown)


src/components/ComboButton/ComboButton.js
```diff
@@ -169,6 +170,28 @@ ComboButton.propTypes = {
   /** @type {string} Overflow menu direction. */
   direction: PropTypes.oneOf([TooltipDirection.TOP, TooltipDirection.BOTTOM]),
 
+  /**
+   * The adjustment in position applied to the floating menu.
+   */
+  menuOffset: PropTypes.oneOfType([
+    PropTypes.shape({
+      top: PropTypes.number,
+      left: PropTypes.number,
```
diego-codes, line 179: Could these values also be strings, like `'auto'`?

jendowns: Good point! Will update that 👍

PR #723, opened 2020-08-08
feat(search-bar): apply titleText prop for inner multiselect

### LINKED ISSUE #722: Invalid aria-labelledby on SearchBar scopes dropdown
## Bug - the aria-labelledby element on the scopes dropdown of the SearchBar component doesn't exist. 

The scopes dropdown of the `SearchBar` component has a `<button>` with two `aria-labelledby` attributes: `downshift-n-label` and `downshift-n-toggle-button`. The first of these, `downshift-n-label`, does not exist as an ID. 

image

I think that this is due to the usage of the `MultiSelect` component in `SearchBar`, here:

https://github.com/carbon-design-system/ibm-security/blob/f1c54896459bb0a4beeebdbe7e248d783dde053e/src/components/SearchBar/SearchBar.js#L162-L173

`MultiSelect` takes an optional `titleText` prop, which is used to create a label for the dropdown. However, since that wasn't provided by `SearchBar`, the label is not rendered. That label _would_ have the ID `downshift-n-label`, but it doesn't exist. 

[Link to the `MultiSelect` component in `carbon-components-react`](https://github.com/carbon-design-system/carbon-components-react/blob/c3bf0123a2a98ca84acfc7e86e69840eaf57a89e/src/components/MultiSelect/MultiSelect.js#L245-L249)

When `titleText` is provided, the `aria-labelledby` IDs are both valid: [Storybook example](https://react.carbondesignsystem.com/?path=/story/multiselect--default)

---

Unfortunately, I'm not quite sure how you'd handle this without making a visual change that results in adding a text label above the scope dropdown :( 

### Optional information

**Version -** `1.28.0`

jendowns:
## Affected issues

- Resolves #722 

## Proposed changes

- add `titleText` prop to provide label for inner `MultiSelect` (the generated `<label>` element is then tied to ARIA attributes in generated markup)
- add `hideScopesLabel` boolean prop to allow you to visually hide the generated `MultiSelect` label while still maintaining the ARIA attributes. NOTE: set to `true` by default for compatibility with existing implementations in product.

## Testing instructions

- pull down branch and experiment with turning `hideScopesLabel` on/off, inspect the ARIA attributes, etc.

### DIFF
--- a/src/components/SearchBar/SearchBar.js
+++ b/src/components/SearchBar/SearchBar.js
@@ -96,7 +96,7 @@ export default class SearchBar extends React.Component {
     titleText: PropTypes.string,
 
     /**
-     * Whether or not the scopes MultiSelecto label is visually hidden.
+     * Whether or not the scopes MultiSelect label is visually hidden.
      */
     hideScopesLabel: PropTypes.bool,
   };
--- a/src/components/SearchBar/SearchBar.js
+++ b/src/components/SearchBar/SearchBar.js
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
+import classnames from 'classnames';
 import Button from '../Button';
 import Search from '../Search';
 import MultiSelect from '../MultiSelect';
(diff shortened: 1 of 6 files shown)


Deriohig:
@diego-codes would be great to get a second review on this issue, thanks! Tracking it this sprint in order to fix some AVT bugs  👀

src/components/SearchBar/SearchBar.js
```diff
@@ -89,6 +89,16 @@ export default class SearchBar extends React.Component {
...
+
+    /**
+     * Provide accessible label text for the scopes MultiSelect.
+     */
+    titleText: PropTypes.string,
+
+    /**
+     * Whether or not the scopes MultiSelecto label is visually hidden.
+     */
+    hideScopesLabel: PropTypes.bool,
```
diego-codes, line 101: Should these be knobs in the storybook previews?

jendowns: Thanks! Just added it 👍

PR #733, opened 2020-08-19
feat(return-to-banner): abstract component, replace `Shell` implementation

### LINKED ISSUE #670: Expose return to banner component
## Description

Expose the return to banner component for consumption across Cloud Paks.

SimonFinney:
## Affected issue

Resolves #670

## Proposed changes

- Abstract `ReturnToBanner` into separate component and add story
- Replace `Shell` implementation
- Add unit tests

### DIFF
--- a/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
+++ b/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
@@ -12489,6 +12489,119 @@ input:not(output):not([data-invalid]):-moz-ui-invalid {
   width: 1px;
 }
 
+.security--return-to-banner {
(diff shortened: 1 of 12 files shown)


SimonFinney:
Interesting point - My initial thoughts are that today the design intent is quite different between both the 'Return to' banner and `SkipToContent`, and they live alongside each other as different forms of navigation, but this doesn't mean that can't change.

@cameroncalder is currently evaluating how the 'Return to' banner lives in a cross-Cloud Pak world so this could be one aspect to consider? See https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-skiptocontent

PR #738, opened 2020-08-24
feat(wizard): add prop to provide accessible label to nav

### LINKED ISSUE #737: Wizard Nav has empty aria-label
## Bug - The aria-label of the nav in the Wizard component is `""`

AVT testers pointed out that the Wizard did not have an aria-label. I did a bit of digging, and it looks like the aria-label is explicitly set to an empty string, without an option to be overridden: 

https://github.com/carbon-design-system/ibm-security/blob/d2c9533c6b78a6b35370efc2e39df12d6d5f66d6/src/components/Wizard/Wizard.js#L212

jendowns:
## Affected issues

- Resolves #737

## Proposed changes

- adds `navLabel` prop with a default value "Steps navigation", to provide an accessible label to the inner `Nav` component. "Steps navigation" versus "Wizard navigation" because "Wizard" is more of a UI dev/design term, and this label is meant to be understood by users.

## Testing instructions

- verify the nav markup has a valid `aria-label` attribute in the "Editable Wizard" variation: https://deploy-preview-738--ibm-security.netlify.app/?path=/story/patterns-wizard--with-editable-wizard

### DIFF
--- a/src/components/Wizard/Wizard.js
+++ b/src/components/Wizard/Wizard.js
@@ -247,6 +247,7 @@ class Wizard extends Component {
       title,
       subTitle,
       className,
+      navLabel, // avoid spreading with rest of props.
       ...other
     } = this.props;
     const componentLabels = {
--- a/src/components/Wizard/Wizard.js
+++ b/src/components/Wizard/Wizard.js
@@ -209,7 +209,7 @@ class Wizard extends Component {
     const { currentStep, valid } = this.state;
 
     return !this.sequentialMode ? (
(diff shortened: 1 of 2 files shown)


jendowns:
Thanks @diego-codes! I thought about adding a storybook knob but wasn't sure about the value versus something more interactive in the UI. Would you prefer there be a knob for this new prop?

diego-codes:
I don't have a strong preference either way. Your reasoning makes sense to me.

PR #757, opened 2020-09-10
fix(scroll-gradient): add immediate child combinator to prevent overw…

### LINKED ISSUE #756: Bug - x-axis ScrollGradient within a y-axis ScrollGradient has overwriting styling
## Bug - 
When an x-axis `ScrollGradient` (e.g. used in the Carbon table with horizontal overflow) is placed within a y-axis `ScrollGradient` (e.g. used on some page layouts), the css controlling the positioning and width of the before and after x-axis scroll gradients is being overwritten by the outer y-axis css.

This can be reproduced and inspected on the sandbox below.

**Expected behavior -**
Scroll gradient indicators for the x-axis `ScrollGradient` (data table) should inherit correct styles.

image

**Actual behavior -**
Scroll gradient indicators for the x-axis `ScrollGradient` (data table) have overwritten styles from outer y-axis `ScrollGradient`.

image

### Steps for reproducing
[CodeSandbox](https://codesandbox.io/s/nervous-pare-09he9?file=/src/index.js)

### Potential fix
I have fixed this very easily by adding the immediate child combinator (`>`) so that `.security--scroll-gradient--y` and `security--scroll-gradient--x` only target the direct `.security--scroll-gradient__before` and `.security--scroll-gradient__after` children elements and not those further down the DOM — as is currently the case. You can test this in the browser using the sandbox.

I'll put that PR up with this issue and attach ✌️

ShayMurnin:
## Affected issues

- Resolves #756 

## Proposed changes

- Add the immediate child combinator (`>`) so that `.security--scroll-gradient--y` and `.security--scroll-gradient--x` only target the direct `.security--scroll-gradient__before` and `.security--scroll-gradient__after` children elements and not those further down the DOM 

## Testing instructions

- I've just tested the css output (built using `yarn build`) on my [sandbox](https://codesandbox.io/s/nervous-pare-09he9?file=/src/index.js) and it works but perhaps there's a better way of validating these kinds of changes?

### DIFF
--- a/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
+++ b/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
@@ -7649,33 +7649,33 @@ input:not(output):not([data-invalid]):-moz-ui-invalid {
   z-index: 1;
 }
 
-.security--scroll-gradient--x .security--scroll-gradient__before,
-.security--scroll-gradient--x .security--scroll-gradient__after {
+.security--scroll-gradient--x > .security--scroll-gradient__before,
+.security--scroll-gradient--x > .security--scroll-gradient__after {
   top: 0;
   width: 3rem;
   height: 100%;
 }
 
-.security--scroll-gradient--x .security--scroll-gradient__before {
+.security--scroll-gradient--x > .security--scroll-gradient__before {
   left: 0;
 }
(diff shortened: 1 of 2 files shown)


SimonFinney (approved):
LGTM! 💯

PR #762, opened 2020-09-22
fix(summary-card): fix action overflow

### LINKED ISSUE #755: Bug - SummaryCardAction expanded content outside of card during animation
**Expected behavior -**

Given that the `SummaryCard` is a container of stand-alone content, the content in the component should remain visibly within it. 

When a user clicks on a `SummaryCardAction` button with expanded content, it should slide the content up within the boundaries or the `SummaryCard`:
image

**Actual behavior -**

In the slide transitions, the expanded content of the `SummaryCardAction` button appears outside of the boundaries of the `SummaryCard`:

image

This behavior breaks the visual principles of a card component. Additionally, it forces a temporary scroll bar on the viewport when the card is positioned near the bottom of the said viewport.

### Steps for reproducing

1. Go to the [`SummaryCard` Storybook preview](https://ibm-security.carbondesignsystem.com/?path=/story/patterns-summarycard--with-primary-label)
2. Click on the first action button on the first card to expand its content
3. Notice the bug described above

### Screenshots

Kapture 2020-09-09 at 12 02 53

### Optional information

**Version -** `1.31.0-prerelease.3`

SimonFinney:
## Affected issue

Resolves #755

## Proposed change

Fix `SummaryCardAction` overlay overflow

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@carbon/ibm-security",
-  "version": "1.32.0-prerelease.4",
+  "version": "1.32.0-prerelease.5",
   "description": "Carbon for IBM Security",
   "license": "Apache-2.0",
   "main": "lib/index.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@carbon/ibm-security",
-  "version": "1.32.0-prerelease.3",
+  "version": "1.32.0-prerelease.4",
   "description": "Carbon for IBM Security",
   "license": "Apache-2.0",
   "main": "lib/index.js",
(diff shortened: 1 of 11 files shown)


diego-codes:
I don't know if you noticed, but there is now a sideways scrollbar:
image

I think that by applying the overflow style for one axis it also applies something for the other axis. You may need to restrict it at a subcomponent level.

SimonFinney:
@diego-codes Good catch! I'm wondering if setting it on both axes is sufficient enough without having to worry about any other unintended side effects.

diego-codes:
I don't think that will work either because I believe that the action buttons with tooltips will get cut off like so:

image

PR #771, opened 2020-10-02
build(storybook): upgrade packages and configuration

SimonFinney:
## Proposed changes

- Upgrade to latest Storybook packages
- Migrate configuration as per https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#to-mainjs-configuration
- Use native Storybook controls for theme addon
- Remove `addon-centered` and update layout configuration as it's now handled natively
- Remove unused packages no longer required with migration


Deriohig (approved):
Noticed one thing with the upgrade, but think this is a bug on their part. The overflow menu has become difficult to select as their custom scroll bars overlay it. 
image

PR #774, opened 2020-10-07
feat: set up mid-line truncation option towards DataDecorator

### LINKED ISSUE #773: Set up mid-line truncation option for DataDecorator but still display whole value in opened panel
## Request - 
We are using DataDecorator component from version @carbon/ibm-security@1.25.0. We have one requirement that is to display the value of DataDecorator with mid-line truncation style when total length of value is more than 20 but still show the whole value without any truncation in the opened panel after clicking. Currently, the value in DataDecorator and panel are the same. I would like to create a PR for the UI improvement I just mentioned.

### Details

- **Application / team -** CP4S - Data Explorer
- **Work item -** []()
- **Needed by -** 

### Proposal

圖片

### Goals / use cases

- [ ] 
- [ ] 
- [ ]

JennChao:
## Affected issues

- Resolves #773 

## Proposed changes

- midLineTruncation prop
- update unit tests
- update publicAPI snapshot
- use whole value as title when title prop is undefined

## Testing instructions

-


src/components/DataDecorator/DataDecorator.js
```diff
@@ -263,6 +281,14 @@ DataDecorator.propTypes = {
 
   /** @type {func} Descriptive text for screen readers that details the severity of a score. */
   scoreDescription: PropTypes.func,
+
+  /** @type {object} Mid-line truncation options applied to value of decorator if applicable. */
+  midLineTruncation: PropTypes.shape({
+    enabled: PropTypes.bool,
```
SimonFinney, line 287: Is it possible to remove the need for an `enabled` or is there a use case for it? Can it be implied when `back`, `front`, and `maxLength` exist?

src/components/DataDecorator/DataDecorator.js
```diff
@@ -59,7 +59,25 @@ class DataDecorator extends Component {
       scoreThresholds,
       scoreDescription,
       onContextMenu: propOnContextMenu,
+      midLineTruncation,
     } = this.props;
+    const truncationUtil = (inputText, maxLength, front, back) => {
```
SimonFinney, line 64: 👏

src/components/DataDecorator/DataDecorator.stories.js
```diff
@@ -35,6 +35,28 @@ const storyProps = () => ({
   renderFooter: () => <Button size="large">Custom footer</Button>,
 });
 
-storiesOf(patterns('DataDecorator'), module).add('Default', () => (
-  <DataDecorator {...storyProps()} />
-));
+const midLineStoryPros = () => ({
```
SimonFinney, line 38:
```suggestion
const midLineStoryProps = () => ({
```

src/components/DataDecorator/DataDecorator.stories.js
```diff
@@ -35,6 +35,28 @@ const storyProps = () => ({
...
+    enabled: true,
+    maxLength: 20,
+    front: 12,
+    back: 4,
+  },
+});
+
+storiesOf(patterns('DataDecorator'), module)
+  .add('Default', () => <DataDecorator {...storyProps()} />)
+  .add('mid-line truncation', () => <DataDecorator {...midLineStoryPros()} />);
```
SimonFinney, line 62:
```suggestion
  .add('mid-line truncation', () => <DataDecorator {...midLineStoryProps()} />);
```

src/components/DataDecorator/DataDecorator.js
```diff
@@ -288,6 +314,7 @@ DataDecorator.defaultProps = {
   stopPropagationEvents: undefined,
   scoreDescription: (score, scoreThresholds) =>
     `Score ${score} out of ${scoreThresholds.slice(-1)[0]}`,
+  midLineTruncation: {},
```
SimonFinney, line 317: If going with an `enabled` prop, perhaps we can provide healthy defaults so it requires less configuration from a consumer's end? For example, these defaults could be the props outlined below in `DataDecorator.stories.js`:

https://github.com/carbon-design-system/ibm-security/pull/774/files#diff-e5a99da599b761a0caca127c61f54c52a1eac4f10da8af6c55c24d9770d2101fR54-R56

src/components/DataDecorator/DataDecorator.js
SimonFinney: Would it make sense to have these changes in the `Decorator` also? I understand it's needed for the `Panel`, so perhaps there's a way to share the logic so there's parity between the `DataDecorator` and `Decorator`.

src/components/DataDecorator/__tests__/DataDecorator.spec.js
```diff
@@ -52,6 +52,49 @@ describe('DataDecorator', () => {
     expect(queryAllByText(/10.0.0.0/i).length === 2);
   });
 
+  test('should use the value of the data decorator to the panel title when no mid-line truncation is required due to mid-line truncation is disabled', () => {
```
SimonFinney, line 55: 👏

SimonFinney (commented):
Thanks @JennChao - this looks awesome! 🎉  I've added a couple of comments, questions, and suggestions - let me know your thoughts

src/components/DataDecorator/DataDecorator.stories.js
JennChao: fixed

src/components/DataDecorator/DataDecorator.stories.js
JennChao: fixed

src/components/DataDecorator/DataDecorator.js
JennChao: I will let `Decorator` component handles the truncation work instead. And by doing so, both `DataDecorator` and `Decorator` can share the same API of truncation.

src/components/DataDecorator/DataDecorator.js
JennChao: fixed

SimonFinney (approved):
LGTM! 🎉  Thank you, and congratulations on your first contribution 👏  👏  👏

JennChao:
> LGTM! 🎉 Thank you, and congratulations on your first contribution 👏 👏 👏

Thank you for your quick review as well! @SimonFinney

PR #741, opened 2020-08-26
feat(storybook): add `addon-docs` support

SimonFinney:
## Proposed changes

- Add `addon-docs` to support detailed guidance documentation
- Refactor `addon-theme` to support theming in the 'Docs' tab
- Add `addon-docs` support for stories written in Component Story Format (CSF)
- Add `addon-controls` (replacement for `addon-knobs`) for better control in 'Docs' tab
- Delegate the configuration for addons and panel based on the format stories are written in
- Remove `meta` helper for flexibility in Storybook story definitions

PR #776, opened 2020-10-07
docs(type-layout): refactor to better support `addon-docs`

SimonFinney:
## Affected issue

Related to #741

## Proposed change

- Reorder component declarations in `TypeLayout.js`
- Set `className` `defaultProp` to `null` to prevent empty string from being generated
- Rewrite prop documentation content to better align with Carbon's


src/components/TypeLayout/TypeLayout.js
```diff
@@ -56,67 +47,76 @@ const TypeLayout = ({
...
+const defaultProps = {
   children: null,
-  size: 'md',
+  className: null,
 };
 
-const { children, className } = propTypes;
-
-const borderPropType = bool;
+const size = 'md';
```
jendowns, line 90: Hey I was wondering why this is abstracted. I'm seeing it only used in the propTypes and default propType below, right? To me this abstraction doesn't add anything, and just takes something away (i.e., immediately understanding what the default value is & what one of the `size` prop options are). Would you consider removing this & referencing `'md'` directly?

src/components/TypeLayout/TypeLayout.js
```diff
@@ -19,19 +19,10 @@ import {
...
-  >
-    {children}
-  </StructuredListCell>
-);
-
 // TODO: `2.x` - Remove deprecated prop `bordered`.
 
 /**
- * Type layout component.
+ * Type layouts provide a means to orderly layout terms and definitions.
```
jendowns, line 25: I think this is a typo here? 

Perhaps:

```suggestion
 * Type layouts provide an orderly layout of terms and definitions.
```

I think it's just the "orderly" part that seems like the typo but feel free to go with whatever you prefer! 👍

src/components/TypeLayout/TypeLayout.js
```diff
@@ -56,67 +47,76 @@ const TypeLayout = ({
...
+
+  /** Provide an optional class to be applied to the containing node */
+  className: string,
 };
 
-TypeLayout.defaultProps = {
-  border: false,
-  bordered: null,
-  className: '',
+const defaultProps = {
```
jendowns, line 85: Why is this being abstracted here? I think it's only used in the default props below?

To me this makes things harder to scan and comprehend. Would you consider moving this information down where it is used?

jendowns (commented):
Thank you @SimonFinney -- please let me know if you have any questions about this feedback 👍

jendowns:
Oh also, would you consider renaming this PR to a `docs` type? My impression of this PR is that is a storybook documentation update, primarily, so technically a `refactor` is meant for a bigger release than intended?

src/components/TypeLayout/TypeLayout.js
SimonFinney: Good point! I'll move it down

src/components/TypeLayout/TypeLayout.js
SimonFinney: Done!

jendowns (approved):
Thank you @SimonFinney 👍

PR #777, opened 2020-10-07
refactor(border): fix helper import, use `carbon--rem`

SimonFinney:
## Proposed change

- Add missing import for partial
- Use `carbon--rem` instead of `carbon--mini-units`


j1mie (approved):
LGTM

PR #779, opened 2020-10-08
fix(toolbar): add default values for compatibility

SimonFinney:
## Proposed change

Add default value for `Toolbar` `mainNavigation` and `ariaLabel` as otherwise causes rendering to crash

Toolbar


Deriohig (approved):
LGTM. TY

PR #783, opened 2020-10-12
feat(layout-modules): add page tab module, update decorator style

SimonFinney:
## Proposed change

- Add `PageTabModule` layout module and sub-components and Storybook stories
- Update decorator styles to better accommodate layout modules


Deriohig (approved):
👍

PR #793, opened 2020-10-14
feat(layout-modules): add `ICAModule`, fix `TitleBarModule` spacing

SimonFinney:
## Proposed changes

- Add `ICAModule` component and stories
- Fix `TitleBarModule` spacing as per design review
- Update ordering of exports to be alphabetical


src/components/UNSTABLE__LayoutModules/ICAModule/index.scss
```diff
@@ -0,0 +1,42 @@
...
+@include security--component($name: #{$layout-modules__name}--ica) {
+  /// Spacing.
+  /// @type Length
+  $ica-module__spacing: $carbon--layout-01;
+
+  padding-top: $ica-module__spacing;
+  padding-bottom: $ica-module__spacing;
+
+  &--hover {
+    padding-top: $ica-module__spacing;
```
Deriohig, line 24: this is probably intentional, but just to be sure the heights of your ICAModule will be different depending if it's interactive or not with these.

Deriohig (approved):
looks good, one small comment thats probably nothing

src/components/UNSTABLE__LayoutModules/ICAModule/index.scss
SimonFinney: That's exactly right 🎉

PR #796, opened 2020-10-14
feat(layout-modules): remove `PageTabModule` from public API

SimonFinney:
## Proposed changes

- Remove `PageTabModule` from public API as it's being excluded for now
- Rename `PageTabModule.stories` to prevent it from being included as part of Storybook
- Revert `withResponsive` decorator to reduce the width on stories as per design review


Deriohig (approved):
LGTM

PR #798, opened 2020-10-16
fix(tearsheet): only render titles if props are provided

jendowns:
## Proposed changes

JSX:
- only render title node if non-required prop `mainTitle` is provided
- only render sidebar title node if non-required prop `sidebarTitle` is provided
- only render sidebar subtitle node is non-required prop `sidebarSubtitle` is provided

Styles:
- move bottom margin on `mainTitle` heading node to top margin of content node in case a `mainTitle` is not provided

## Testing instructions

- You can remove any of those props from the Tearsheet story and see that it does not render an empty heading node


SimonFinney (approved):
LGTM! Thanks 💯

PR #801, opened 2020-10-19
feat(desciption-list): add `stacked` variant

### LINKED ISSUE #789: Request - `DescriptionList` / `TypeLayout` - Stacked
## Request - `DescriptionList` / `TypeLayout` - Stacked

Stacked variation of the `DescriptionList` / `TypeLayout`:

DescriptionList - Stacked

SimonFinney:
## Affected issue

Resolves #789

## Proposed change

Add stacked description list variant


cameroncalder (changes requested):
The spacing looks good! When it breaks down to the stacked variant, the "term" should change to `label-01` size

CassidyJensen (approved):
This looks great! Excited to have access to this. Could you notify me when you set up unit tests for this case? Since it is still new to me - seeing an example would be great for following up with unit tests on the ICA updates.

SimonFinney:
> This looks great! Excited to have access to this. Could you notify me when you set up unit tests for this case? Since it is still new to me - seeing an example would be great for following up with unit tests on the ICA updates.

Thanks! And sure - because this is a simple variation, the [unit test is quite basic](https://github.com/carbon-design-system/ibm-security/blob/789/stacked-description-list/src/components/TypeLayout/__tests__/TypeLayout.spec.js#L108-L114), but you might be able to pull inspiration from the size-related unit tests in this file.

PR #802, opened 2020-10-19
fix(nav): correctly apply `onClick` handlers, spread `Toolbar` props

### LINKED ISSUE #784: onClick doesn't seem to be working for navItems
## Bug 
When i add the onClick event listener it doesn't seem to do anything

SimonFinney:
## Affected issue

Resolves #784

## Proposed change

- Call `onClick` handlers for `NavItem`s generated in `Nav` and `NavList`
- Pass-through props for `Toolbar` to support `onClick` handlers
- Update stories to reflect changes

### DIFF
--- a/src/components/Nav/Nav.stories.js
+++ b/src/components/Nav/Nav.stories.js
@@ -16,7 +16,7 @@ storiesOf(components(name), module).add(
   'Default',
   () => (
     <div style={{ width: '300px' }}>
-      <Nav heading="Nav example" label="Nav">
+      <Nav heading="Nav example" label={Nav}>
         <NavList title="Nav list 1">
           <NavItem key="navitem_1-1" element="span" customprop="uniqueValue">
             Nav item 1-1 (with a custom element)
--- a/src/components/Nav/Nav.stories.js
+++ b/src/components/Nav/Nav.stories.js
@@ -1,16 +1,19 @@
 /**
  * @file Navigation stories.
- * @copyright IBM Security 2019
+ * @copyright IBM Security 2019 - 2020
  */
 
-import React from 'react';
+import { action } from '@storybook/addon-actions';
 import { storiesOf } from '@storybook/react';
+import React from 'react';
 
 import { components } from '../../../.storybook';
 import { Nav, NavItem, NavList } from '../..';
 
-storiesOf(components('Nav'), module).add(
-  'default',
+const { name } = Nav;
+
+storiesOf(components(name), module).add(
+  'Default',
   () => (
     <div style={{ width: '300px' }}>
       <Nav heading="Nav example" label="Nav">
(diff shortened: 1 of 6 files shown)


SimonFinney:
@clairek20 Thanks for your patience - Here's a PR that investigates the issues that you've been encountering. 

Would you mind reviewing the changeset and the deploy preview and let me know if it satisfies your requirements? Thank you!

clairek20 (approved):

clairek20:
Hi @SimonFinney 

Thanks for looking at this - i approved there! Looks good to me!

Yohanna:
Hi @SimonFinney, thanks for fixing the issue!

I'm not sure how to test the changes locally, but I can see on the deploy preview version that the Toolbar component props now have an `onClick` option which is exactly what we need.

Thank you! 👍🏼

src/components/Nav/Nav.stories.js
```diff
@@ -1,24 +1,27 @@
...
+import { action } from '@storybook/addon-actions';
 import { storiesOf } from '@storybook/react';
+import React from 'react';
 
 import { components } from '../../../.storybook';
 import { Nav, NavItem, NavList } from '../..';
 
-storiesOf(components('Nav'), module).add(
-  'default',
+const { name } = Nav;
```
jendowns, line 13: @SimonFinney Was this left over from something else, or am I missing context? I'm not sure what it's doing here aside from renaming two (out of several) occurrences of the word "Nav"?

src/components/Toolbar/_mocks_/index.js
```diff
@@ -52,6 +52,7 @@ const applicationsToGenerate = [
     ]),
   },
   {
+    onClick: console.log,
```
jendowns, line 55: Did you mean to use this over an "action" in storybook?

src/components/Nav/NavList/NavList.js
```diff
@@ -50,11 +50,19 @@ export default class NavList extends Component {
   buildNewItemChild({ props }, index) {
     const { onItemClick, activeHref } = this.props;
 
+    const { onClick } = props;
```
jendowns, line 53: Could `onClick` have a propType definition? And also, what becomes of `onItemClick`? (Deprecated?)

jendowns (approved):
@SimonFinney Approving because it appears to solve the bug & I don't want to hold this up, but I still had a couple things I think should be addressed before merging.

src/components/Toolbar/_mocks_/index.js
SimonFinney: Good idea!

src/components/Nav/NavList/NavList.js
SimonFinney: This is the `onClick` from specific `NavItem`s which never ended up being called - As far as I can tell `onItemClick` is a more general handler for all `NavItem`s within a `NavList`

src/components/Nav/Nav.stories.js
SimonFinney: That's right! It was a minor refactor while I was in adding the `onClick` handler

PR #806, opened 2020-10-22
feat(panel): add overlay toggling

SimonFinney:
## Proposed change

Experiment with toggling panel overlay

## Testing instructions

`withOverlay` and `focusTrap` knobs

### DIFF
--- a/src/components/PanelV2/PanelV2.js
+++ b/src/components/PanelV2/PanelV2.js
@@ -117,7 +117,7 @@ function PanelV2({
       <Transition className={namespace}>
         {isOpen && (
           <Portal
-            focusTrap={focusTrap || withOverlay}
+            focusTrap={withOverlay || focusTrap}
             focusTrapOptions={focusTrapOptions}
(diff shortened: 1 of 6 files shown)


src/components/PanelV2/PanelV2.js
```diff
@@ -116,8 +117,9 @@ function PanelV2({
       <Transition className={namespace}>
         {isOpen && (
           <Portal
-            focusTrap={focusTrap}
+            focusTrap={withOverlay || focusTrap}
             focusTrapOptions={focusTrapOptions}
+            hasOverlay={withOverlay}
```
dbrugger, line 122: for the sake of consistency with the `focusTrap` prop might wanna go without the `has` prefix

SimonFinney: I was following https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Loading/Loading.js#L105 as a reference point, but we can align with the individual component props if that makes more sense

PR #810, opened 2020-10-28
fix(filter-panel-checkbox-overflow): migrate state logic to CSS

### LINKED ISSUE #748: Bug - Hover state background remains on filters after the overflow menu has been opened/closed
## Bug - Hover state background remains on filter after the overflow menu has been opened/closed, even though cursor has been moved away

**Expected behavior -**

When cursor is removed, background colour hover state is also removed and the filter background colour is returned to normal state
image

**Actual behavior -**

Hover state background colour remains even though the cursor has been removed
image

### Steps for reproducing

1. On https://ibm-security.carbondesignsystem.com/?path=/story/patterns-filterpanel-filterpanelcheckboxwithoverflowmenu--default component, note the normal background colour of the filter before hovering
2. Hover over the filter and note the background changes colour in the hover state
3. Click on the overrflow menu to open it
4. Click again on the overflow menu to close it
5. Move the cursor away and click elsewhere on the UI

### Affected browsers

Chrome (Version 85.0.4183.83 (Official Build) (64-bit))
Edge (Version 85.0.564.44 (Official build) (64-bit)) 
Firefox (80.0.1 (64-bit))
Safari (Version 13.1.2 (15609.3.5.1.3))

SimonFinney:
@diego-codes Let me know your thoughts here as you're the original contributor and perhaps I'm missing something glaringly obvious - I ran this against `accessibility-checker` and the Equal Access Accessibility Checker and all seems good.

While debugging #748 and other issues related to this, the logic for managing state related to blurs, focusing, and mouse interactions for the `FilterPanelCheckboxWithOverflowMenu` seemed resource-intensive in comparison to offloading the work to CSS, especially when there could easily be 1000+ of these components on a page at any given time in Data Explorer.

## Affected issue

Resolves #748 

## Proposed changes

- Refactor to move `OverflowMenu` visibility logic to CSS
- Use Carbon's `getMenuOffset` method
- Reduce tests

### DIFF
--- a/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
+++ b/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
@@ -8937,17 +8937,12 @@ input:not(output):not([data-invalid]):-moz-ui-invalid {
   background-color: var(--field-01, #262626);
 }
 
-.security--filter-panel-checkbox-with-overflow-menu--open > .security--filter-panel-checkbox-with-overflow-menu__overflow-button, .security--filter-panel-checkbox-with-overflow-menu:hover > .security--filter-panel-checkbox-with-overflow-menu__overflow-button, .security--filter-panel-checkbox-with-overflow-menu:focus-within > .security--filter-panel-checkbox-with-overflow-menu__overflow-button {
-  opacity: 1;
-}
-
 .security--filter-panel-checkbox-with-overflow-menu__overflow-button {
   position: absolute;
   right: 0;
   height: 1.5rem;
   width: 1.5rem;
   background-color: inherit;
-  opacity: 0;
 }
 
 .security--filter-panel-checkbox-with-overflow-menu__overflow-options {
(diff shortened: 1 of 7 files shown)


diego-codes:
Hi @SimonFinney ! The reason the overflow menu button display was handled with JS is that we saw a significant lag in performance when we had a large number of overflow menu buttons rendered in the DOM (even if they weren't visible). It could take up to 1 - 2 seconds for an overflow menu to appear when the user would click on an overflow menu button. It seemed that by limiting the number of overflow menus on the DOM the response time would be immediate.

Based on that, you may want to run a test to see if there is a lag present when you render possibly hundreds of these.

diego-codes (approved):
This is looking pretty good! This seems to resolve the related issue.

I did that the overflow menu button disappears when we close the overflow menu by clicking the button again. It's probably not a huge problem, but I expected the button to remain since the mouse is still hovering over the checkbox area. What do you think?

Kapture 2020-10-30 at 11 14 43

SimonFinney:
@diego-codes Good catch!

diego-codes (approved):
Looks great, @SimonFinney!

Issue #808, opened 2020-10-27
React `v17` upgrade preparation


SimonFinney:
Issue to prepare for upgrading to React `v17` - https://reactjs.org/blog/2020/10/20/react-v17.html

https://www.npmjs.com/package/eslint-plugin-react-hooks will help with identifying React functionality that is no longer compatible.

PR #817, opened 2020-11-02
build(lint): add `eslint-plugin-react-hooks`

SimonFinney:
## Affected issue

Contributes to #808 

## Proposed change

Add `eslint-plugin-react-hooks` package and configuration


.eslintrc
```diff
@@ -6,6 +6,7 @@
     "plugin:import/recommended",
     "plugin:jsx-a11y/recommended",
     "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
```
diego-codes, line 9: Do you need to include it as a plugin to work?

SimonFinney: It should be enough to extend the configuration unless we need to customize the rules further - https://www.npmjs.com/package/eslint-plugin-react-hooks#installation

diego-codes: Gotcha, perfect! I'll approve then.

diego-codes (approved):

PR #828, opened 2020-11-06
feat: add NavItem element support for Toolbar

### LINKED ISSUE #705: Toolbar - Ability to pass in a link component to use for navigation instead of the default anchor tag
## Request -
When using the toolbar with the shell we would like to use custom link components to allow navigation using the `next.js` router. Currently the toolbar is restricted to links formed using anchor components causing our SPA to reload. 

### Details

- **Application / team -** QRadar / Falcon
- **Work item -** Navigation
- **Needed by -** 

### Proposal

I have seen the updates to the NavItem Component to allow a custom tag to be used #155 , since the toolbar is made up of NavItem components it would be useful if the additional `element` and `customprop` props could be exposed to the toolbar so that a routing link component could be passed in such as [next/link](https://nextjs.org/docs/api-reference/next/link)

areddon:
## Affected issues

- Resolves #705 

## Proposed changes

- Pass `element` prop from `Toolbar` props down to the `NavItem` component.

## Testing instructions

- Verified by modifying `Toolbar` mock data to include an `element` prop and inspecting the rendered HTML in Storybook. Did not update mocks but can if required.

### DIFF
--- a/src/components/Toolbar/Toolbar.js
+++ b/src/components/Toolbar/Toolbar.js
@@ -170,6 +170,7 @@ export default class Toolbar extends Component {
                     {children.map(
                       ({
                         href: navigationListItemHref,
+                        element: navigationListItemElement,
                         content,
                         icon,
                         id: navigationListItemId,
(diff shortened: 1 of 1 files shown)


SimonFinney (approved):
LGTM! Thank you 💯

PR #831, opened 2020-11-09
fix(combo-button): key is not a prop warning

dbrugger:
## Affected issues

```
Warning: ComboButtonItem: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://reactjs.org/link/special-props)
    at ComboButton
```
This might cause developers looking for errors on their side when they use the `ComboButton`.

## Proposed changes

- Don't use a loop for the single button
- Don't try to access `key` from the prop

## Testing instructions

- Run storybook in dev mode (using `yarn start`) and verify there is no more warning
- Verify `ComboButton` still works as exepected


SimonFinney (approved):
LGTM - Thanks! 💯

PR #836, opened 2020-11-10
build(analytics): fix analytics property

SimonFinney:
## Proposed change

Update location of analytics configuration as it wasn't getting initialized, as per https://github.com/storybookjs/storybook/issues/9727

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -10,3 +10,7 @@ import theme from './theme';
 addons.setConfig({
   theme,
 });
+
+// Set the Google Analytics tracking ID if the `master` branch is deployed.
+window.STORYBOOK_GA_ID =
+  process.env.STORYBOOK_BRANCH === 'master' && 'UA-80770450-3';
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -91,8 +91,4 @@ const parameters = {
   viewMode: 'story',
 };
 
-// Set the Google Analytics tracking ID if the `master` branch is deployed.
-window.STORYBOOK_GA_ID =
-  process.env.STORYBOOK_BRANCH === 'master' && 'UA-80770450-3';
-
 export { decorators, parameters };


.storybook/manager.js
```diff
@@ -10,3 +10,7 @@ import theme from './theme';
 addons.setConfig({
   theme,
 });
+
+// Set the Google Analytics tracking ID if the `master` branch is deployed.
+window.STORYBOOK_GA_ID =
+  process.env.STORYBOOK_BRANCH === 'master' && 'UA-80770450-3';
```
diego-codes, line 16: Not sure if there is much you can do about this now, but should the GA ID be provided as an environment variable as well to keep it hidden from source control?

.storybook/manager.js
```diff
@@ -10,3 +10,7 @@ import theme from './theme';
 addons.setConfig({
   theme,
 });
+
+// Set the Google Analytics tracking ID if the `master` branch is deployed.
+window.STORYBOOK_GA_ID =
```
diego-codes, line 15: I'm curious, why does the ID setting work here and not in the other file?

SimonFinney: The addon documentation is lacking in usage guidance and the issue above initially led me to the solution, but as I understand it from https://storybook.js.org/docs/react/configure/overview#configure-story-rendering, `preview.js` is for managing the contents of the `iframe` that renders the stories, while `manager.js` is geared towards handling configuration more specific to Storybook's UI and for setting options, so it made sense to me to place it here.

.storybook/manager.js
SimonFinney: 😆  Yeah, it's a bit late - the rationale at the time was that it would still be accessible either by viewing the website's source or by logging `window.STORYBOOK_GA_ID` to the console.

Granted it takes a bit more work than just browsing source code, but I do know there are spam filters we can leverage if the need arises.

diego-codes (approved):

.storybook/manager.js
diego-codes: If there is a way to regenerate that ID and store it in an environment variable, I would do so.

PR #847, opened 2020-11-26
feat(combo-button): add prop spreading to primary action button

### LINKED ISSUE #845: Proposal: ComboButton —— Allow primary action link to open in a new tab
### Description
Currently we can open `ComboButton` menu items links in a new tab as it allows for spreading props on each item. However this is currently not possible for the primary action itself (when it is a link).

See screenshots below. I think I'll put a PR for this today.

#### As-is
combo-button__as-is

#### Proposed to-be
combo-button__to-be

### Acceptance criteria

- [ ] `target="_blank"` can be added as a prop to the primary action

#### Accessibility

- [ ] Semantic HTML is used by default where relevant, with ARIA attributes used as required
- [ ] No violations produced as a result of automated accessibility tests

#### Breakpoints

- [ ] Mobile, tablet, and desktop, if applicable

#### Browser support

- [ ] [browserl.ist supported browsers](https://github.com/carbon-design-system/ibm-security/blob/master/package.json#L174-L178)

#### Documentation

- [ ] `defaultProps` and `propTypes` are included as part of component definitions, and documented in a human-readable format

#### Internationalization

- [ ] No hardcoded text exposed to the user, even with screen readers

#### Sass

- [ ] Follows the directory structure and naming convention, or provides a good justification for changing it

#### Storybook

- [ ] Stories that capture features of the component
- [ ] Provide controls for customising features of the component

#### Tests

- [ ] Tests are behaviour-driven and written similarly to how a user would interact with the component
- [ ] Automated accessibility tests

ShayMurnin:
### Affected issues

- Resolves #845 

### Proposed changes

- Spread `{...buttonProps}` onto the primary action button. This gives consumers the added flexibility to add props for the underlying `Button` element such as `target="_blank"` which for example provides the ability to open primary action links in a new tab.
- Migrated `ComboButton.spec.js` unit tests to using `react-testing-library` and added accessibility tests

### Testing instructions

- I have added the extra prop `target="_blank"` to the primary action item in `ComboButton.stories.js`. This link should now open in a new tab.
- I have also added an extra `ComboButtonItem` component to `ComboButton.stories.js` which exemplifies the use case of opening a link in a new tab. I think this offers greater visibility for consumers.

### DIFF
--- a/src/components/ComboButton/ComboButton.js
+++ b/src/components/ComboButton/ComboButton.js
@@ -39,6 +39,7 @@ const ComboButton = ({
   // may trigger invalid DOM warnings.
   const primaryActionWithProps = (
     <Button
+      {...buttonProps}
       className={classnames(buttonProps.className, `${namespace}--primary`)}
       disabled={buttonProps.disabled}
       href={buttonProps.href}
--- a/src/components/ComboButton/ComboButton.stories.js
+++ b/src/components/ComboButton/ComboButton.stories.js
@@ -36,6 +36,7 @@ storiesOf(patterns('ComboButton'), module).add(
           href="#"
           onClick={action(`onClick (Item 1)`)}
           renderIcon={ArrowRight20}
+          target="_blank"
         >
           Item 1 -- becomes primary button and text will be truncated
         </ComboButtonItem>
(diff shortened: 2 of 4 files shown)


SimonFinney (approved):
LGTM - Thank you for the test clean-up! 🎉

PR #851, opened 2020-12-02
feat(layout-modules): add `withBackground` higher-order component

SimonFinney:
### Proposed changes

- Expose `withBackground` higher-order component per early adopter feedback and request


Deriohig (approved):
I actually came across this an was gonna create an issue for it for it  🎉

PR #852, opened 2020-12-02
feat(motion): proxy motion package via `globals`

SimonFinney:
### Proposed change

Proxy `@carbon/motion` via `globals/motion/index`


areddon:
@SimonFinney 

There are few components such as the `SummaryCard` that should load the global motion file instead of directly from `@carbon/motion`

PR #862, opened 2020-12-04
fix(data-table): correct scroll gradient not displaying

### LINKED ISSUE #860: Bug: DataTable —— ScrollGradient no longer displaying
### Description

I've just noticed the `ScrollGradient` component wrapping the `DataTable` isn't displaying.

After doing a little digging, I've found the issue. It seems as part of [#6566](https://github.com/carbon-design-system/carbon/pull/6566), an new wrapper was created for the `DataTable`. In the PR it mentions:

> - moved the table to it's own wrapper so that scrolling is applied to the container only, this fixes issues with elements above the table when the table itself overflows

The `.bx-data-table-content` [selector](https://github.com/atikenny/carbon/blob/ad0b533e06042fc85c78ac05b21416f0f4d450e0/packages/components/src/components/data-table/_data-table-core.scss#L25) was  added to control overflow for the table. With this, the content inside the scroll gradient then handles its own scrolling as it has `overflow: auto`. For our scroll gradient effect though, we want the scroll gradient to handle the scrolling of course — so `.security--scroll-gradient__content` being the scroll container instead of `.bx-data-table-content`.

I have recreated the issue with two data tables in this [sandbox](https://codesandbox.io/s/dazzling-wescoff-vc9hl?file=/src/index.js) and I've added a little snippet of css to showcase the fix in the second data table. Looks like the fix is just a matter of targeting:

```css
.security--data-table .bx--data-table-content {
    overflow: unset;
}
```

____

I'll create a PR for this tomorrow and also add a story to showcase to consumers the `ScrollGradient` displayed for `DataTable` with scrolling content.

### Acceptance criteria

- [ ] `ScrollGradient` is displayed for `DataTable` components with scrolling content

#### Accessibility

- [ ] Semantic HTML is used by default where relevant, with ARIA attributes used as required
- [ ] No violations produced as a result of automated accessibility tests

#### Breakpoints

- [ ] Mobile, tablet, and desktop, if applicable

#### Browser support

- [ ] [browserl.ist supported browsers](https://github.com/carbon-design-system/ibm-security/blob/master/package.json#L174-L178)

#### Documentation

- [ ] `defaultProps` and `propTypes` are included as part of component definitions, and documented in a human-readable format

#### Internationalization

- [ ] No hardcoded text exposed to the user, even with screen readers

#### Sass

- [ ] Follows the directory structure and naming convention, or provides a good justification for changing it

#### Storybook

- [ ] Stories that capture features of the component
- [ ] Provide controls for customising features of the component

#### Tests

- [ ] Tests are behaviour-driven and written similarly to how a user would interact with the component
- [ ] Automated accessibility tests

ShayMurnin:
### Affected issues

Resolves #860 

### Proposed changes

Correct scroll gradient not displaying

### Testing instructions

New `with scroll gradient` story added to storybook


SimonFinney (approved):
LGTM - Thank you! 🔥

PR #870, opened 2020-12-16
feat(toolbar): conditionally render settings and support

Kieran-CLX:
### Proposed changes

- Update Toolbar component to optionally render settings and support panels. 
- Added menu panel only toolbar to storybook


Kieran-CLX:
Hi @SimonFinney ,
I know some toolbar tests are failing due to the updates to prop types. Before fixing the tests I first wanted your opinion my proposed changes. 
Regards,
Kieran

SimonFinney:
Hey @Kieran-CLX ! Thanks for submitting - this approach makes sense to me! Will the height of the toolbar remain the same as the original example with this use case?

Kieran-CLX:
> Hey @Kieran-CLX ! Thanks for submitting - this approach makes sense to me! Will the height of the toolbar remain the same as the original example with this use case?

Hi @SimonFinney, My latest commit introduces a conditional height if the settings and support menu items are not being used. I used a style attribute to apply the conditional height. I'd like to get your thoughts on the approach. Have you any other suggestions?

src/components/Toolbar/Toolbar.js
```diff
@@ -236,7 +236,9 @@ export default class Toolbar extends Component {
       },
       renderAddons = [],
     } = this.props;
-
+    const renderSupport = this.props.support.length > 0;
+    const renderSettings = this.props.settings.length > 0;
+    const height = renderSupport && renderSettings ? '100%' : '50px';
     const classes = classnames(namespace, className);
```
SimonFinney, line 242: Instead of an inline style, could the style be applied through a conditional class in https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/Toolbar/_mixins.scss via [classnames](https://www.npmjs.com/package/classnames#usage)? 

Instead of explicitly defining a height of `50px`, perhaps the (`bottom`)[https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/Toolbar/_mixins.scss#L135] can be toggled to achieve the same effect.

SimonFinney (approved):
LGTM! 🔥

It looks like the only thing outstanding is to update the snapshots, and then this can be merged - thank you!

PR #877, opened 2020-12-17
fix(delimited-list): remove unnecessary Hooks dependent

SimonFinney:
### Affected issue

- Contributes to #808 
- Contributes to #973 

### Proposed change

Fix `eslint-plugin-react-hooks` `DelimitedList` warning


areddon:
Is this component supposed to be responsive? Such that the `hasOverflow` can change when the component is resized? In its current form, and likely even before this update doesnt update after the initial render, only when the list of items changes. Which itself is a problem as the consumer has to ensure that the reference to their array of items does not change, otherwise the `useEffect` will be called on every render.

SimonFinney:
> Is this component supposed to be responsive? Such that the `hasOverflow` can change when the component is resized? In its current form, and likely even before this update doesnt update after the initial render, only when the list of items changes. Which itself is a problem as the consumer has to ensure that the reference to their array of items does not change, otherwise the `useEffect` will be called on every render.
> 
> EDIT: Since the reference doesn't update until after the component renders, I am assuming that the component never has the overflow, updating the branch to view the generated Storybook.

@areddon Good question! I'm not entirely sure. If I remember correctly, for this change I sought to resolve the linting error and only paid attention to whether the behavior had regressed from https://ibm-security.carbondesignsystem.com/?path=/story/components-security-delimitedlist .

areddon:
I just noticed that it is using `useLayoutEffect` and not `useEffect`, which will run after all DOM manipulations.

areddon (approved):

PR #885, opened 2021-01-14
feat(toolbar): enable nodes in titles

clairek20:
## Pull request -
Changed toolbar title from prop type string to node so we can have a title as well as a component.

### Affected issues

- Resolves  SECCON-9522

### Proposed changes
- Change title string to node
- Updated storybook mocks also


SimonFinney (approved):
LGTM! 🔥 Thanks

PR #888, opened 2021-01-19
fix(portal): Update `Tooltip` `z-index` inside `Portal`

SimonFinney:
### Proposed change

Update `Tooltip` `z-index` when inside `Portal`, similar to https://github.com/carbon-design-system/carbon/pull/6757


diego-codes (approved):
LGTM!

PR #899, opened 2021-01-28
feat(button): align `Button` sizes with Carbon

SimonFinney:
### Proposed changes

- Align `Button` large variations with Carbon's - https://react.carbondesignsystem.com/?path=/docs/button--default#button-size
- Redirect documentation to Carbon's `Button` documentation
- Remove override


src/components/Button/Button.js
```diff
@@ -47,36 +45,38 @@ const Button = ({
...
+  largeText: null,
 };
 
 Button.propTypes = {
   ...propTypes,
 
-  /** @type {boolean} Specify whether or not the button is in a loading state. While loading, icons provided via the `renderIcon` prop will not be displayed. */
-  loading: PropTypes.bool,
+  /** Specify the kind of `Button` */
+  kind: oneOf(ButtonKinds),
```
Deriohig, line 67: image
 is there a way to get this to display the actual supported strings in the doc?

Deriohig (changes requested):
Think maybe it would be better to define the kinds directly so that its easier to know what values are supported through the docs.

src/components/Button/Button.js
SimonFinney: Not without hardcoding them, unfortunately, which leads to them being out of sync with the actual list. 😭

As a short-term workaround, we've been pointing out directly to Carbon's documentation, which explains why the scope of this PR blew up a little bit... 😆 Let me know your thoughts on what the better approach might be here.

Deriohig: I understand. In the future, it would be great if security components were more distinct from core carbon components. Instead of being a proxy for another library, and having to handle their baggage, we were distinct from them. I think this plays a lot into the great work your doing lately with layouts. Its hard to find a sweet spot where we are easy to use, in sync, and not creating a dependency bottleneck. But I personally feel (maybe im wrong) that layouts are in the right direction.

Deriohig (approved):

Issue #908, opened 2021-02-25
`ICA` trending variant


SimonFinney:
> @CassidyJensen @SimonFinney Hey, this is great to hear! These are the sizes we'd need plus the specs (Default, Large, Xlarge)
> 
> ![Artboard](https://user-images.githubusercontent.com/50155706/109040548-a37bc600-7693-11eb-945f-23a856de6533.png)

Awesome - thanks! We'll need to create a separate issue for the 'trending' arrow, but I believe this will cover it for sizes.

_Originally posted by @SimonFinney in https://github.com/carbon-design-system/ibm-security/issues/907#issuecomment-785765365_

CassidyJensen:
@cameroncalder In the case of being used on the ICAModule - how should the trending arrow appear? 

The default appearance is in a column, but it seems like it should be in a row with the value. I've added screenshots of the current and potential layouts below. 

![image](https://user-images.githubusercontent.com/45407808/109330175-9fc57c00-7818-11eb-9e68-b56d985590fd.png)
![image](https://user-images.githubusercontent.com/45407808/109331062-9dafed00-7819-11eb-8b4e-ff873a82b161.png)
![image](https://user-images.githubusercontent.com/45407808/109331221-d354d600-7819-11eb-986c-0a07589a74ac.png)

cameroncalder:
@CassidyJensen The bottom version would be preferred

PR #910, opened 2021-03-01
feat(ica-update): ica sizes and trending variant

### LINKED ISSUE #907: ICA Size Large
### Description

The ICA / ICAModule needs a large size variation. An additional variable can be supplied to ICA in order to change the size from the default settings. The large version of the ICA uses $heading-06 for the value and $heading-03 for the total. Are there any other requirements / changes for the large version?

image

### Goals / Use cases

- [ ] Provide alternative way to display ICA

CassidyJensen:
## Pull request 
Added sizes default, lg, and xlg to the ICA component. Added trending variant.

### Affected issues

- Resolves #907 and #908

### Proposed changes

- Styling for large and xlarge ICA
- Added trending icon 
- Added large and xlarge for trending. 
- Improved styling to handle ICAModule and ICAWall cases with the trending arrow.

### Testing instructions

- Go to the ICA story on storybook
- Use the ICA size and trending knobs to test the different sizes. (Should match the styling below).
image


cameroncalder (changes requested):
Hey Cassidy, great steps forward here! Have a few notes:


**Universal feedback**
When the “percentage knob” is selected, it seems to restrict the ICA number sizes to change

The percentage should be placed on the same optical baseline as the large number (I’ll include spacing specs. Some include combinations of spacer tokens since in some cases there isn’t one single spacer that works perfectly on its own)

“Percentage” font size should be equal to the “Total” font size

The “total” should be displayed to the right of the large number by default and only shown below the number if space is an issue (position



**XLarge version**
The label font should change to body-short-01

cameroncalder:
An updated image that includes spacer placement for the "Percentage" and "Total"

Artboard

cameroncalder (commented):
Approved with one minor comment. Would you mind removing the space between the "/" and 
 the total value

cameroncalder:
This is looking great 👍

CassidyJensen:
> Approved with one minor comment. Would you mind removing the space between the "/" and
> the total value

oops! I will remove that now.

src/components/ICA/ICA.js
```diff
@@ -146,15 +188,28 @@ ICA.propTypes = {
 
   /** Specify whether or not the values should be truncated. */
   truncate: PropTypes.bool,
+
+  /** The size of the ICA
+   * @type string
+   */
+  size: PropTypes.oneOf(['default', 'large', 'xlarge']),
```
SimonFinney, line 195: For consistency, could we align with [Carbon's naming for sizes](https://react.carbondesignsystem.com/?path=/docs/components-button--default#button-size)? For example, `lg` instead of `large`.

src/components/ICA/ICA.js
```diff
@@ -146,15 +188,28 @@ ICA.propTypes = {
 
   /** Specify whether or not the values should be truncated. */
   truncate: PropTypes.bool,
+
+  /** The size of the ICA
+   * @type string
+   */
```
SimonFinney, line 194: Because the type is generated from the propType, we don't need to add the additional `@type` annotation here.

```suggestion
  /** The size of the ICA. */
```

src/components/ICA/ICA.js
```diff
@@ -81,20 +107,36 @@ const ICA = ({
...
 
   return (
-    <div className={`${namespace} ${className}`} {...other}>
-      <h4 className={`${namespace}__label`}>{label} </h4>
-      <span className={`${namespace}__value`}>{truncatedValue}</span>
-      {shouldDisplayof ? (
-        <span className={`${namespace}__total`}> / {truncatedTotal}</span>
-      ) : null}
+    <div className={ICAClasses} {...other}>
+      <h4 className={`${namespace}__label ${ICAClasses}__label`}>{label} </h4>
```
SimonFinney, line 119: As I understand, with the inclusion of `ICAClasses`, this will output `security--ica__label security--ica security--ica--large__label`, which could cause unintended side effects.

If this is to hook into styles for sizes, would it be sufficient to use a selector like `&--lg &__label`? This could potentially be applied anywhere `ICAClasses` is used.

src/components/ICA/ICA.js
```diff
@@ -81,20 +107,36 @@ const ICA = ({
...
+            className={`${namespace}__trend ${ICAClasses}__trend`}
+            renderIcon={renderIcon}
+          />
+        )}
+        <span className={`${namespace}__value ${ICAClasses}__value`}>
+          {truncatedValue}
+        </span>
+        {shouldDisplayof ? (
+          <span
+            className={`${namespace}__total ${ICAClasses}__total ${trending &&
```
SimonFinney, line 132: Similarly to above, [`classnames`](https://www.npmjs.com/package/classnames#usage) can be leveraged to conditionally handle these classes.

src/components/ICA/_index.scss
```diff
@@ -27,13 +27,57 @@
   & &__total {
     @include carbon--type-style($name: body-short-01);
     margin-bottom: 0;
+    padding-top: $spacing-01 + $spacing-04;
```
SimonFinney, line 30: We have hit consumer Sass conflicts in the past for not namespacing dependency spacing variables - here's an example of how this can be applied:

```suggestion
    padding-top: $carbon--spacing-01 + $carbon--spacing-04;
```

SimonFinney (commented):
Thanks for your patience, and your first contribution -the design looks fantastic and spot on! 🎉

I've added a few comments and suggestions for your consideration - additionally, have you considered the kind of unit tests that could be included with these new variants?

cameroncalder (approved):
LGTM!

src/components/ICA/ICA.js
CassidyJensen: I removed the redundancy in the ICAClasses setup. It should now output `security--ica__label  security--ica--large`. Does this seem like a good fix?

CassidyJensen:
> Thanks for your patience, and your first contribution -the design looks fantastic and spot on! 🎉
> 
> I've added a few comments and suggestions for your consideration - additionally, have you considered the kind of unit tests that could be included with these new variants?

@SimonFinney 
Not yet. Writing unit tests is completely new for me. What would you recommend? 

I have a couple ideas based on the [Testing Documentation](https://github.com/carbon-design-system/ibm-security/tree/dev/docs/testing).
There should be a test for the ICAClasses because they are custom classes correct? 
Would there need to be any additional accessibility testing or would the existing tests still cover it?

src/components/ICA/ICA.js
```diff
@@ -146,15 +181,26 @@ ICA.propTypes = {
 
   /** Specify whether or not the values should be truncated. */
   truncate: PropTypes.bool,
+
+  /** The size of the ICA. */
+  size: PropTypes.oneOf(['default', 'lg', 'xlg']),
```
SimonFinney, line 186:
```suggestion
  size: PropTypes.oneOf(['default', 'lg', 'xl']),
```

src/components/ICA/ICA.js
```diff
@@ -146,15 +181,26 @@ ICA.propTypes = {
...
+
+  /**
+   * Display trending icon
+   * @type bool
+   */
```
SimonFinney, line 191:
```suggestion
  /** Display trending icon. */
```

src/components/ICA/ICA.js
SimonFinney: Apologies for any confusion so far - we can keep the `icaClasses` on the return function's root node, and remove it everywhere else as the styles can handle it from the top level. For example:

```js
  return (
    <div className={icaClasses} {...other}>
      <h4 className={`${namespace}__label`}>{label} </h4>
      ...
```

and in the styles:

```scss
@include security--component($name: ica) {

  &--lg & {
    &__label {
      ...
    }

    &__value {
      ...
    }
  }

  &--xl & {
    &__label {
      ...  
    }
  }

  ...
}
```

src/components/ICA/ICA.js
```diff
@@ -65,11 +69,29 @@ const ICA = ({
...
+  size,
+  trending,
   truncate,
   total,
   value,
   ...other
 }) => {
+  const isSize = sizeValue => size === sizeValue;
+  const isLarge = isSize('large') || isSize('lg');
+  const isXLarge = isSize('xlarge') || isSize('xlg');
```
SimonFinney, line 81: For naming consistency with [Carbon's sizes](https://react.carbondesignsystem.com/?path=/docs/components-button--default#button-size):

```suggestion
  const isLarge = isSize('lg');
  const isXLarge = isSize('xl');
```

src/components/ICA/ICA.js
```diff
@@ -65,11 +69,29 @@ const ICA = ({
...
+  let renderIcon = ArrowUp16;
+  if (isLarge) {
+    renderIcon = ArrowUp20;
+  } else if (isXLarge) {
+    renderIcon = ArrowUp24;
+  }
+
+  const ICAClasses = classnames(className, {
+    [`${namespace}--lg`]: isLarge,
+    [`${namespace}--xlg`]: isXLarge,
```
SimonFinney, line 92: For naming consistency with [Carbon's sizes](https://react.carbondesignsystem.com/?path=/docs/components-button--default#button-size):

```suggestion
    [`${namespace}--xl`]: isXLarge,
```

SimonFinney:
Thanks @CassidyJensen ! As mentioned on Slack, if it helps you could submit the unit tests in a follow-up. Let me know if you haven't any questions on the final few pieces of feedback. 🔥

CassidyJensen:
Doing a follow up with the unit tests would be great! Thank you. 
@cameroncalder Once I get the final approval from @SimonFinney this should be ready to wrap up!

SimonFinney (approved):
LGTM! 🔥 Thank you!

PR #922, opened 2021-04-06
feat(list-module): add ListItemModule stories

lily-peng:
## Pull request 
Adding new stories for ListItem Module:
- Default
- Non Interactive
- As Button
- Detailed
- With Profile Image
- With Tag
- With Component
- With Status Icon
- With In Progress Icon
- With Distributed Columns

### Affected issues

- For [listItem module #243](https://github.ibm.com/security/design-core-experience/issues/243)

### Proposed changes

- 10 stories for ListItem module in Storybook

### Testing instructions

1. Go to the UNSTABLE Layout Modules -> ListItemModule (Canary) in storybook
2. Read through the Docs
3. Browse through the 10 stories
4. Consider functionality such as hover, links, nested components, column layouts, etc.


src/components/UNSTABLE__LayoutModules/ListItemModule/ListItemModule.stories.js
```diff
@@ -0,0 +1,239 @@
...
+export const Default = () => (
+  <ListItemModule href="#">
+    {({ Column, getLayoutProps }) => (
+      <>
+        <Column>
+          <h2 {...getLayoutProps({ type: 'title' })}>List title</h2>
+
+          <span {...getLayoutProps({ type: 'label' })}>Label</span>
+        </Column>
+      </>
```
paul-balchin-ibm, line 42: There's no need for wrapping a single child in a Fragment

src/components/UNSTABLE__LayoutModules/ListItemModule/ListItemModule.stories.js
```diff
@@ -0,0 +1,239 @@
...
+export const NonInteractive = () => (
+  <ListItemModule>
+    {({ Column, getLayoutProps }) => (
+      <>
+        <Column>
+          <h2 {...getLayoutProps({ type: 'title' })}>List title</h2>
+
+          <span {...getLayoutProps({ type: 'label' })}>Label</span>
+        </Column>
+      </>
```
paul-balchin-ibm, line 56: There's no need for wrapping a single child in a Fragment

src/components/UNSTABLE__LayoutModules/ListItemModule/ListItemModule.stories.js
```diff
@@ -0,0 +1,239 @@
...
+export const AsButton = () => (
+  <ListItemModule onClick={() => console.log('hi')}>
+    {({ Column, getLayoutProps }) => (
+      <>
+        <Column>
+          <h2 {...getLayoutProps({ type: 'title' })}>List title</h2>
+
+          <span {...getLayoutProps({ type: 'label' })}>Label</span>
+        </Column>
+      </>
```
paul-balchin-ibm, line 70: There's no need for wrapping a single child in a Fragment

paul-balchin-ibm (approved):

src/components/UNSTABLE__LayoutModules/ListItemModule/ListItemModule.stories.js
lily-peng: Thanks for catching these!

cameroncalder (commented):
This first pass is looking great!! Here are a few notes:

1. The listItem module should also include a visible `ui-03` line to separate a single item from others in a list.
2. Default: Use the same example shown in the Docs as the example shown in the Canvas
3. Use `body-short-01` instead of `body-long-01` for both the list title & description
4. Change description copy to "Description here. It can go up to three lines before truncating."
5. Title + label version should include the title bottom spacer and label top spacer equalling a total of 8px between the two. Spacers should only be overwritten if it impedes the list container padding. For context, this applies to the "Non Interactive and As Button" examples
6. With Profile image: profile image size should be size 24 instead of 32. (Perhaps that is something that will come easier once the C&CS profile image component is available)

lily-peng:
> This first pass is looking great!! Here are a few notes:
> 
> 1. The listItem module should also include a visible `ui-03` line to separate a single item from others in a list.
> 2. Default: Use the same example shown in the Docs as the example shown in the Canvas
> 3. Use `body-short-01` instead of `body-long-01` for both the list title & description
> 4. Change description copy to "Description here. It can go up to three lines before truncating."
> 5. Title + label version should include the title bottom spacer and label top spacer equalling a total of 8px between the two. Spacers should only be overwritten if it impedes the list container padding. For context, this applies to the "Non Interactive and As Button" examples
> 6. With Profile image: profile image size should be size 24 instead of 32. (Perhaps that is something that will come easier once the C&CS profile image component is available)

Thanks for the thorough review!

1. The line is there when multiple `ListItemModule`s are being used. There isn't a visible example of that in Storybook, perhaps we can considering adding it as a story or in the docs?
2. I got rid of the "Detailed" story and used that as the "Default"
3. Done ✅ 
4. Done ✅ 
5. Good catch. It looks like those `margin`s are collapsing, so I changed the top ones to `padding`. This also fixes this behavior in Description+Label combos ("With Profile Image") and between the Description+Component with the 4px bottom of Description being lost in the 16px margin top of Component.
6. It looks like the current out-of-the-box size is 28px, and the only other provided variant is larger. I've added this size as custom styling, as it'll just be temporary

src/components/LayoutModules/ListItemModule/ListItemModule.stories.js
```diff
@@ -0,0 +1,231 @@
...
+  },
+  decorators: [withResponsive],
+};
+
+export const Default = () => (
+  <ListItemModule href="#">
+    {({ Column, getLayoutProps }) => (
+      <>
+        <Column>
+          <Bee16 {...getLayoutProps({ type: 'icon' })} />
```
SimonFinney, line 44: I've been thinking a lot about the API for this recently - what are your thoughts on `getLayoutProps` returning an object of props to choose from instead of the consumer passing a `type` parameter? For example, this line becomes `<Bee16 {...getLayoutProps().icon} />` or for multiple styles, it could become:

```
<ListItemModule href="#">
  {({ Column, getLayoutProps }) => {
    const { icon, title } = getLayoutProps();      

    return (
      <>
        <Bee16 {...icon} />
        <h2 {...title}>List title</h2>
          ...

paul-balchin-ibm: I prefer `const { icon, title } = getLayoutProps();` as it's a more common pattern than `{...getLayoutProps({ type: 'icon' })}`, but just looking at `<Bee16 {...icon} />` doesn't tell me how much prop spreading is actually happening. If I see `{...props}` then I expect they are generic props passed down from the parent (width, height, etc), but `{...icon}` is both specific and vague at the same time.

lily-peng: Hey guys, I agree that looks cleaner on the consumer side. It adds a bit of bulk to our `ListItemModule` itself but it's nice to see all the different options in one place. Just pushed a new commit with new preview link, let me know what you all think.

SimonFinney:
@lily-peng Sorry about the delay, I'll have an opportunity to check this out again soon - it looks like there's some noise in the changeset, would you know what's happening there?

lily-peng:
> @lily-peng Sorry about the delay, I'll have an opportunity to check this out again soon - it looks like there's some noise in the changeset, would you know what's happening there?

Branch looks good now!

PR #241, opened 2019-11-28
feat(modular-styles): add remaining imports, update source import

SimonFinney:
## Affected issues

- Resolves partially #44

## Proposed changes

- Update remaining components
- Fix missing imports
- Update source imports for compilation, Storybook and theme addon
- Remove fixed style overrides
- Clean up

PR #937, opened 2021-04-13
feat(css-grid): use css grid for Page Layouts

lily-peng:
## Pull request 
Updating Page Layouts to utilize the new Carbon CSS Grid:
- Detail
- Overview

### Affected issues

- For [Layout modules narrow grid compatability #241](https://github.ibm.com/security/design-core-experience/issues/241)

### Proposed changes

Update Page Layouts to utilize the new Carbon CSS Grid.

### Testing instructions

1. Go to Page Layouts (Canary) -> Detail -> Default in storybook
2. Go to Page Layouts (Canary) -> Overview -> Default in storybook


lily-peng:
Would love another pair of eyes to look at the Data Table alignment and what's up with those borders

src/components/UNSTABLE__PageLayouts/Detail.stories.js
```diff
@@ -47,187 +48,197 @@ export default {
...
-                  value="5020c08bcc061236643293bf0d897321"
-                />
-              ),
-            },
-          ]}
-        />
-      </Column>
-    </Row>
-  </>
+  <div className={`${settings.prefix}--css-grid`}>
```
SimonFinney, line 51: - [ ] In general, we can destructure the Carbon prefix to reduce the repetition, for example:

```js
const { prefix } = settings;

export const Default = () => (
  <div className={`${prefix}--css-grid`}>
    <div className={`${prefix}--col-span-100`}>
      ...
```

src/components/UNSTABLE__LayoutModules/Layout.stories.js
```diff
@@ -0,0 +1,567 @@
+/**
```
SimonFinney, line 1: This file should now be split out into `Detail.stories` and `Overview.stories` and should no longer be necessary

.storybook/index.scss
```diff
@@ -48,3 +50,13 @@
     }
   }
 }
+
+/*** TEMPORARY ***/
+
+.bx--css-grid {
```
SimonFinney, line 56: Question - is this a relic from local experimentation or is it required as an override or workaround?

src/components/UNSTABLE__PageLayouts/Detail.stories.js
```diff
@@ -47,187 +48,197 @@ export default {
...
+                <SummaryCardHeader title="Vulnerability report" />
+              </SummaryCard>
+            </Column>
+          </Row>
+        )}
+      </CardAreaModule>
+    </div>
+
+    <div className={`${settings.prefix}--col-span-100`}>
+      <RowWithContainer narrow>
```
SimonFinney, line 148: We should now avoid using `RowWithContainer` and instead rely on the CSS grid classes.

For context, this was created as a temporary workaround for the flex grid implementation and will likely cause unintended side effects with the CSS grid implementation - my hope is that with this work we can also remove these workarounds from the codebase.

src/components/UNSTABLE__PageLayouts/Detail.stories.js
```diff
@@ -47,187 +48,197 @@ export default {
...
+            size="lg"
+          />
+        </ActionBarModuleItems>
+      </ActionBarModule>
+    </div>
+
+    <div className={`${settings.prefix}--col-span-100`}>
+      <CardAreaModule>
+        {({ getLayoutProps }) => (
+          <Row narrow>
```
SimonFinney, line 121: Question - is mixing the CSS grid and flex implementations intentional?

SimonFinney (commented):
🙏 Thanks for your patience, and good progress so far - I've submitted some general feedback as well as some questions for things I don't fully understand.

Also, for your consideration, see below some screenshots of some things that seemed off to me:

#### Responsive behaviour

Responsive

#### Column gutter quirks

Padding

Thanks, and let me know your thoughts!

.storybook/index.scss
lily-peng: These are styles taken from `bx--grid` that don't seem to be included for `bx--css-grid`. I guess I marked Temporary just cause I haven't looked into other ways to bring them in

src/components/UNSTABLE__PageLayouts/Detail.stories.js
lily-peng: I'm not sure how compatible LayoutModules themselves are for the CSS Grid, but I'll look into tackling that next. I'm thinking for these, we would pass in a `bx--css-grid` rather than a `Row` ?

.storybook/index.scss
SimonFinney: Ah, I see! Thanks for clarifying - could you include `bx--grid` as an additional class on the grid container?

src/components/UNSTABLE__PageLayouts/Detail.stories.js
SimonFinney: I see - when you have a chance, would you mind elaborating on the challenges you uncovered when working with integrating the layout modules with the CSS grid?

.storybook/index.scss
lily-peng: Oh, I suppose I can, but is it not the intention to try to stay away from the old `bx--grid` implementation? Is it that it's okay to keep `bx--grid` around for its styles since it doesn't actually utilize flex?

src/components/UNSTABLE__PageLayouts/Detail.stories.js
lily-peng: Hey Simon, sorry I haven't had too much time to really dig into details here. In general, modules such as CardArea and ICAModule look like they're mainly set up with the old Row/Column configurations from the docs and examples. Switching over to the CSS grid classes unfortunately doesn't produce a perfect 1-to-1. Mainly the alignments and some of the structure needs changing. 

For instance, the old CardArea takes in a Row with Columns, which since the CSS grid doesn't have Rows anymore, will need to change to a Grid with Columns that comes with their own spacing in the form of padding, etc.

I still need to look into it more, because it's possible my alignment issues could be on my end too. Let me know what you think!

.storybook/index.scss
SimonFinney: You're right, that's definitely the intention so we'd like to avoid it where possible - which styles are required, just the `max-width`?

I guess this hangs on Carbon's intentions for grid `max-width`s moving forward. @tay1orjones could you provide some insight here?

src/components/UNSTABLE__PageLayouts/Detail.stories.js
SimonFinney: Thanks for the background, this is all great insight to have! My understanding was that the `CardAreaModule` only provides vertical spacing, so either a CSS grid or a flexbox implementation could both work in theory, but it sounds like the structure of the `CardAreaModule` may be preventing the use of mixed grid modes or a subgrid?

SimonFinney:
> > #### Column gutter quirks
> > Padding
> > Thanks, and let me know your thoughts!
> 
> Will look into the responsive behavior when I get the chance. Guessing the CSS Grid has its own classes for different screen widths.
> 
> As for the gutter, it's how I can get the background to leave that amount of space on the left, otherwise the text would be flush against it. From the [CSS Grid Layouts](https://github.com/carbon-design-system/carbon/blob/main/packages/grid/examples/css-grid/src/pages/layouts.js) page:
> 
> > Text must manually be given an additional .bx--gutter-start class applying padding-left to align text the column definition.
> 
> Their examples are just on `<p>` blocks, which makes me wonder... if we want our modules to support the CSS grid, would it make sense to add that `.bx--gutter-start` class to text-based modules such as TitleBarModule?

That's an interesting observation! I'd want to understand if that would also make sense for condensed and wide grid configurations too.

🤔 I see what you mean about the gutter here - it looks like we also need something applied to the opposite side to prevent it from flowing all the way to the edge.

SimonFinney:
@lily-peng Based on the conversations above, it looks like there are potentially some challenges with a few things - should we reassign so the details can be dug into further?

.storybook/index.scss
tay1orjones: @SimonFinney Sorry for the wildly late reply on this, unfortunately got lost in my notifications 😓 

Elements with `bx--css-grid` shouldn't be given the `bx--grid` class. The implementations are wholly separate from one another. 

`max-width` was missed and so wasn't included in the initial drop but has since been added back in as of https://github.com/carbon-design-system/carbon/pull/8797. There's been quite a few improvements and fixes since then too.

PR #938, opened 2021-04-15
docs(categories): reorganise categories

SimonFinney:
### Proposed change

Further distinguish between Security-specific and proxied Carbon components


SimonFinney:
@cameroncalder what are your thoughts on something like this? I'm hoping to further distinguish between the mental model of Carbon proxied versus Security-specific

cameroncalder:
I like this. The separation might prove to be valuable during the merge or migration so we can be more deliberate about the state of Security-owned components

cameroncalder (approved):
LGTM!

PR #940, opened 2021-04-16
fix(PanelV2): change footer position from fixed to absolute

a110605:
### Proposed changes
Current panelV2 footer position is set to `fixed`. This would create another stacking context within footer.
This might cause footer overflow menu has lower `z-index` than other element in panel context.

See below screenshot :
Screen Shot 2021-04-16 at 12 22 59

### DIFF
--- a/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
+++ b/src/__tests__/scss/__snapshots__/SCSS.spec.js.snap
@@ -5928,7 +5928,7 @@ a.bx--overflow-menu-options__btn::before {
 
 .security--panel--v2__footer {
   display: flex;
-  position: fixed;
+  position: absolute;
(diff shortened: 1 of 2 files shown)


a110605:
Hi @SimonFinney, @liunate, please review this PR as we discussed.

liunate:
I am happy to see the change merged.

My two cents:
Previous footer `position: fixed` is not semantically/behaviorally precise as it means sticking to edge of viewport but footer is not **Footer of viewport**. 

As footer is **Footer of panel**, its containing panel element `.security--panel--v2 { position: relative }` has already provided a stacking context for footer to **stick** with at the bot edge of panel with `.footer { position: absolute }`.

SimonFinney (approved):
💯  LGTM! Thank you

PR #943, opened 2021-04-21
fix(unstable): align experimental components with Carbon format

SimonFinney:
### Proposed changes

- Replace unstable `Pagination` with proxied Carbon variant
- Align experimental components with Carbon format


Deriohig (approved):
Just to confirm, there is a new unstable pagination component being proxied now? Will this change the underlying functionality or props of the deprecated unstable pagination component?

SimonFinney:
@Deriohig Exactly! This will now reference Carbon's implementation per https://github.com/carbon-design-system/carbon/blob/main/packages/react/__tests__/__snapshots__/PublicAPI-test.js.snap#L7434-L7564 where the props match

PR #949, opened 2021-04-27
fix(telemetry): use dependencies from `carbon-components-react`

SimonFinney:
### Proposed change

Use telemetry dependencies from `carbon-components-react` instead of `npx`


Deriohig (approved):
LGTM 🙏

PR #966, opened 2021-05-24
List module design documentation

rebecca-m-l:
## Design documentation

Includes:
- Anatomy
- Variants
- Grid placement


lily-peng (changes requested):
It looks like we're getting an error when trying to preview: "No url found for submodule path 'ibm-security' in .gitmodules." 

I believe the "submodule" being referenced is an empty folder called ibm-security that was created in this commit. If it's not important, can we try removing it and seeing if the preview would work?

rebecca-m-l:
> It looks like we're getting an error when trying to preview: "No url found for submodule path 'ibm-security' in .gitmodules."
> 
> I believe the "submodule" being referenced is an empty folder called ibm-security that was created in [this commit](https://github.com/carbon-design-system/ibm-security/pull/966/commits/244a90ed13c4743ff36ea780550924e8889650a5). If it's not important, can we try removing it and seeing if the preview would work?

Think this is resolved now and preview should work: https://deploy-preview-966--ibm-security.netlify.app/iframe?utm_source=github&utm_campaign=bot_dp

lily-peng (approved):

cameroncalder (approved):
LGTM 🙌

PR #970, opened 2021-05-28
fix: change require to import syntax for `carbon-components` to allow treeshaking again

dbrugger:
### Affected issues

- `carbon-components` JavaScript ends up entirely in the application bundle

### Proposed changes

- change require to import syntax for `carbon-components` to allow treeshaking again

### Testing instructions

- check the snapshot tests to make sure the same prefix is still applied


SimonFinney (approved):
LGTM - thank you! 🎉

PR #989, opened 2021-06-03
feat(ica): Add showValueWarning option

### LINKED ISSUE #965: Add ICA Warning on Total Display
### Description

Design has shown interest in having the value display in a warning red when the ICA value is larger than the total.

image

### Goals / Use cases

- It does not need to be activated. If there is a value and a total it will check if it needs to apply the warning color. At threshold of 100% or above - it will show the warning color.

CassidyJensen:
### Affected issues

- Resolves #965

### Proposed changes

- Added warning class when value >= total. This class turns the color of the value to $support-01
- Added unit test for class sizes
- Added unit test for warning class
- Added unit test for trending arrow

### Testing instructions

- Test by giving value a number equal to and greater than the total. The value should appear in $support-01.


cameroncalder (changes requested):
Looks like the number is automatically defaulting to red even if a total hasn't been designated in the "Total" input field.
image

Is the red text an option to "turn off" or on? There are cases where equalling the total is not a cause for concern.
image

CassidyJensen:
@cameroncalder I will double check with the designer who requested this. 

@SimonFinney Are there actions I can take to reduce all the changes in this PR? That way I can fix that and the changes requested at the same time?

SimonFinney:
@CassidyJensen It seems like new dependencies were included even though none were explicitly added in this PR - running the following in your branch might revert those changes:

1. `rm -rf node_modules`
2. `yarn cache clean`
3. `yarn`

CassidyJensen:
Revisited. Now has a boolean to control whether the warning color is active, though it is a little odd. If the user does not have a total it applies the warning color, is there a better way to deal with this or is it fine to assume someone would only activate it when there is a total? 

I would say I could add a check for whether the total is above 0, but perhaps there is a usecase where the total is 0 and you would want the warning still.

cameroncalder (approved):
LGTM!

dbrugger:
I wanted to resurrect this PR. the reason why this condition triggers when there is no `total` set is that the `total` defaults to `0`. I am not sure this is a good choice. If there is no `total` set it should probably be left undefined.
This aside, the alternative would be to leave the warning condition to the consumer, and accept a prop like `valueCondition: 'default' | 'warning'`.

PR #990, opened 2021-06-04
refactor(components): remove duplicate style import

SimonFinney:
### Proposed change

Remove duplicate Carbon style import

### DIFF
--- a/src/components/_index.scss
+++ b/src/components/_index.scss
@@ -88,8 +88,5 @@
 @import 'Tooltip/index';
 @import 'UIShell/index';
 
-/// Hidden Carbon proxy.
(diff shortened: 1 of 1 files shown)


lily-peng (commented):
Can I ask, what was the original purpose of the line? A remnant from the old way to import the List component?

SimonFinney:
> Just wondering, what was the original purpose of the line? A remnant from the old way to import the List component?

Exactly -- it likely should've been removed when https://github.com/carbon-design-system/ibm-security/blob/dev/src/components/List/_index.scss was introduced.

lily-peng (approved):
LGTM

Issue #973, opened 2021-05-31
Cloud & Cognitive compatibility


SimonFinney:
#### Description

https://ibm.ent.box.com/notes/825385486329

PR #999, opened 2021-06-14
test(combo-button): prevent duplicate label error

SimonFinney:
### Affected issue

Contributes to #973

### Proposed change

Create unique labels for `ComboButton` accessibility tests to prevent duplicate label error


thekevinboyle (commented):
Yeah, looks good to me

PR #1027, opened 2021-06-30
test(styles): fix file reference

SimonFinney:
### Affected issue

- Contributes to #973 
- Resolves https://github.com/carbon-design-system/ibm-cloud-cognitive/pull/918/checks#step:7:1114

### Proposed change

Make styles test entry point relative to the current directory as the `root` variable will vary between repos. In `ibm-security` it's `__dirname`, but in `ibm-cloud-cognitive`, as it's a monorepo, it's `${__dirname}/../..`


lily-peng (approved):
LGTM :)

PR #1035, opened 2021-07-06
487/allow icons next to dropdowns

clairek20:
## Pull request 

Allow icons on nav list items to follow 
image

### Affected issues

- Resolves 487

### Proposed changes

add icon into navlist but make it optional based on icon prop

### Testing instructions

-


src/components/Nav/NavList/NavList.js
```diff
@@ -116,6 +116,15 @@ export default class NavList extends Component {
         role="menuitem"
       >
         <div className={`${navItemNamespace}__link`}>
+          <div>
```
Deriohig, line 119: is this div needed ? if so, should in not be inside your conditional so your not rendering unused elements when the icon isn't present

src/components/Nav/NavList/NavList.js
```diff
@@ -145,17 +154,25 @@ NavList.defaultProps = {
   onListClick: () => {},
   tabIndex: 0,
   title: '',
+  hasIcon: false,
```
Deriohig, line 157: a better strategy is to simply use a render prop. This way you only need a single prop called renderIcon and you can perform your conditional using that. 
```
          <Icon
            className={`${navItemNamespace}__img`}
            renderIcon={this.props.renderIcon}
          />
          ```

File: src/components/Nav/NavList/NavList.js [hunk 3]
```diff
@@ -116,6 +116,15 @@ export default class NavList extends Component {
         role="menuitem"
       >
         <div className={`${navItemNamespace}__link`}>
+          <div>
+            {this.props.hasIcon && (
+              <img
+                alt={'section1'}
+                className={`${navItemNamespace}__img`}
+                src={this.props.icon}
```
Deriohig, line 124: you should deconstruct your props so that your code looks cleaner. You can see this being done on line 99 already

src/components/Nav/NavList/NavList.js
```diff
@@ -116,6 +116,15 @@ export default class NavList extends Component {
         role="menuitem"
       >
         <div className={`${navItemNamespace}__link`}>
+          <div>
+            {this.props.hasIcon && (
+              <img
```
Deriohig, line 121: I believe there is an `<Icon/>` element that is used in the library for this sort of use case.

src/components/Toolbar/Toolbar.js
```diff
@@ -168,13 +168,18 @@ export default class Toolbar extends Component {
                 const hasIcon = icon !== undefined;
 
                 return children ? (
-                  <NavList key={navigationItemId} title={navigationItemTitle}>
+                  <NavList
+                    hasIcon={hasIcon}
+                    icon={icon}
```
Kieran-CLX, line 173: Has icon could be removed and simple check if `icon` is not null

Kieran-CLX (commented):
See inline comment regarding the use of `hasIcon` prop

src/components/Nav/NavList/NavList.js
clairek20: The reason i actually have it in there is so it does render when the icon isn't present. Without it and when the icon isn't present the alignment was all off.

Deriohig: I think you need to address the underlying cause of the alignment being off then, having empty divs isn't good practice.

Deriohig (approved):
👌

PR #1046, opened 2021-07-21
feat(deps): upgrade Carbon to `10.41`

SimonFinney:
### Proposed change

Upgrade Carbon packages to [`10.41`](https://github.com/carbon-design-system/carbon/releases/tag/v10.41.0), updating snapshots.


dbrugger (commented):
thanks @SimonFinney. do these two changes depend on each other? if not we should split them into individual PRs as they seem to serve a different purpose.

thekevinboyle (commented):
Nice one

PR #1052, opened 2021-07-22
fix(header): Added display:inline to header link to add spacing between words

SridharKopalli:
### Affected issues

- Resolves # Header link no space issue

### Proposed changes

- 

### Testing instructions

- Go to Shell component and make sure the header link has proper spacing


dbrugger (commented):
Thanks @SridharKopalli this is looking good! We only need to update the snapshot.
You can do that by running `yarn jest -u` and commit the updated snapshot after review.

paul-balchin-ibm (approved):

dbrugger (approved):

PR #1053, opened 2021-07-23
fix(dependencies): move prettier-config-carbon to dev dependency

dbrugger:
### Affected issues

- prettier-config-carbon is not needed for consumers 

### Proposed changes

- move prettier-config-carbon to dev dependency

### Testing instructions

-


SimonFinney (approved):
Nice catch, thank you! 🥇

PR #1066, opened 2021-08-03
chore(panel): fix eslint issues

dbrugger:
### Affected issues

- Contributes to #973 
- 
### Proposed changes

- Split up margin state into two dedicated ones that can be cleanly set respectively
- Use callback for the ref instead of an effect. see https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

### Testing instructions

-


SimonFinney (approved):
LGTM — thank you!

PR #1074, opened 2021-08-06
fix(header): set aria-expanded and aria-haspopup on the appropriate element

dbrugger:
### Affected issues

- Contributes to #973 
- `aria-expanded` and `aria-haspopup` are not valid on `li` (role `listitem`)

### Proposed changes

- Move the two attributes to the buttons that have this characteristic.
- Clean up the `children` prop.

### Testing instructions

-


SimonFinney (approved):
LGTM — thank you! 💯

PR #1075, opened 2021-08-06
refactor(truncated-list): simplified class component

dbrugger:
### Affected issues

- Fixes incorrect useEffect declarations
- Fixes duplicate first render
- Contributes to #973 

### Proposed changes

- rewrite as a class component
- cleanup

### Testing instructions

-


SimonFinney (approved):
LGTM — nice refactor, thank you! My only nitpick would be to reduce repetition of `this.props` and `this.state` with destructuring wherever it might make sense to do so, but I'll leave this at your discretion.

PR #1095, opened 2021-08-30
chore(deps): update `accessibility-checker`, align configuration

SimonFinney:
### Affected issue

Contributes to #973

### Proposed changes

Update `accessibility-checker` dependencies and align configuration


.achecker.yml
```diff
@@ -5,5 +5,3 @@ failLevels:
 
 policies:
   - WCAG_2_1
-
-ruleArchive: preview
```
areddon, line 9: Just for clarity, would it be better to have `ruleArchive: lastest` in the file even when it is already the default?

areddon (approved):

.achecker.yml
SimonFinney: I like that — I'll be sure to carry that practice into `ibm-cloud-cognitive`, thank you!

PR #1106, opened 2021-09-27
fix(navitem): Remove blue lines in safari

clairek20:
### Proposed changes

- set outline to 0

### Testing instructions

-

### DIFF
--- a/src/components/Nav/_mixins.scss
+++ b/src/components/Nav/_mixins.scss
@@ -95,6 +95,7 @@
     }
 
     &:focus {
+      outline: 0;
       outline-width: 0;
 
       > #{$list}__item__link {


cameroncalder (changes requested):
Thanks for addressing this! The blue lines are gone, but the nav items with a tag now have more height than the ones without. The nav item height should be the same with or without the beta tag

https://user-images.githubusercontent.com/50155706/135084693-726e09d4-cf45-434f-aef7-3ddefb15b493.mov

clairek20:
> Thanks for addressing this! The blue lines are gone, but the nav items with a tag now have more height than the ones without. The nav item height should be the same with or without the beta tag
> 
>  shell.mov

Hey @cameroncalder 

I was confused as to how removing an outline would effect a hover height, so i had a look at this locally without my change and that hover issue is still an issue with the beta tag.  Seems like a different issue that wasn't caused by this change. It's strange though as if i go to https://ibm-security.carbondesignsystem.com/?path=/story/patterns-shell--default it's fine there, it's just an issue locally and in the deploy preview. I even looked at another PR and the deploy preview just to double check and i'm seeing the same https://github.com/carbon-design-system/ibm-security/pull/1108.

Could we maybe get this merged anyway? I can raise an issue for that in the mean time to investigate.

Thanks a million, Claire.

cameroncalder:
@clairek20 appreciate the clarifying comment! Yes, we can merge this one to address the blue outline, and please do open another issue to further investigate the hover size discrepancy

PR #1108, opened 2021-10-01
fix(storybook): Update addon-storysource for renamed Layout Modules

lily-peng:
### Testing instructions

- In Storybook, check that you can see source code under the Story tab in the Canvas for all components, including Layout Module components such as ListItemModule.
- Also for Layout Modules, check the Docs tab to see the "Hide Code" buttons after stories and their corresponding source code.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -16,7 +16,14 @@ module.exports = {
     '@storybook/addon-links',
     '@storybook/addon-postcss',
     '@storybook/addon-viewport',
-    '@storybook/addon-storysource',
+    {
+      name: '@storybook/addon-storysource',
+      options: {
+        rule: {
+          test: [/(-story)|(\.stories)\.jsx?$/],
+        },
+      },
+    },
     '@storybook/addon-knobs',
     '@storybook/addon-controls',
     '@storybook/addon-actions',


cameroncalder (approved):
LGTM! Thank you Lily! 🙏

.storybook/main.js
```diff
@@ -16,7 +16,14 @@ module.exports = {
     '@storybook/addon-links',
     '@storybook/addon-postcss',
     '@storybook/addon-viewport',
-    '@storybook/addon-storysource',
+    {
+      name: '@storybook/addon-storysource',
+      options: {
+        rule: {
+          test: [/(-story)|(\.stories)\.jsx?$/],
```
SimonFinney, line 23: 👏  Awesome — it could make sense to wildcard this for capturing MDX and other potential file formats here too:

```suggestion
          test: /(-story|.stories).*$/,
```

SimonFinney (commented):
Thanks for figuring this out! I had one suggestion.

.storybook/main.js
lily-peng: Hi Simon, thanks for the suggestion. If we include other files like MDX, a strange little glitch appears at the top of the Layout modules library page:
image

I've narrowed it down to the `<Meta>` metadata tag at the top of the page. I'll have to investigate further; let me know if you've got any ideas! Thanks

SimonFinney: > Hi Simon, thanks for the suggestion. If we include other files like MDX, a strange little glitch appears at the top of the Layout modules library page: image
> 
> I've narrowed it down to the `<Meta>` metadata tag at the top of the page. I'll have to investigate further; let me know if you've got any ideas! Thanks

Oh! I see what you mean — perhaps specific is better in this case? For example, something like the following which only captures JavaScript files:

```suggestion
          test: /(-story|.stories).js$/,
```

PR #1135, opened 2021-12-15
fix(ICA): Make edit icon button inline with ICA label and total

clairek20:
### Proposed changes

Before(button was pushing total label down):
image

After:
image


clairek20:
Hey @SimonFinney, would you mind reviewing this, please? Thanks in advance!

